PR #3, opened 2022-11-09
Enable OpenXR integration on Linux Build test on Github

wangra-google:
Enable OpenXR integration on Linux Build test on Github with "OpenXR tests" disabled

### DIFF
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -37,7 +37,8 @@ jobs:
         run: |
           mkdir build
           cd build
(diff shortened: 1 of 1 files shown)


vettoreldaniele (changes requested):
Can you add a comment that `-DBUILD_TESTS` actually only disables OpenXR tests? (It's unfortunate they chose such a generic name for their CMake option...)

wangra-google:
> Can you add a comment that `-DBUILD_TESTS` actually only disables OpenXR tests? (It's unfortunate they chose such a generic name for their CMake option...)

fixed

vettoreldaniele (approved):
Thanks!

PR #9, opened 2022-11-11
Support raw storage buffers in D3D12 and D3D11

### LINKED ISSUE #5: support grfx::DESCRIPTOR_TYPE_STORAGE_BUFFER in DX11 and DX12
These are unimplemented, but would be useful for compute shader experiments.

DX12: https://github.com/google/bigwheels/blob/main/src/ppx/grfx/dx12/dx12_descriptor.cpp#L315 

DX11: https://github.com/google/bigwheels/blob/main/src/ppx/grfx/dx11/dx11_descriptor.cpp#L260

vettoreldaniele:
This allows the use of STORAGE_BUFFER descriptors in D3D12 and D3D11.

To better identify the actual resource being referenced (raw buffers), I renamed `DESCRIPTOR_TYPE_STORAGE_BUFFER` to `DESCRIPTOR_TYPE_RAW_STORAGE_BUFFER`. The `headless_compute` application is fixed to not rely on CPU-to-GPU buffers and instead use upload and readback buffers.

Future improvements:
* Disallow CPU-to-GPU buffers in Vulkan when they are disallowed in D3D12. Applications should not rely on this, and should use proper readback and upload buffers instead.

Fixes #5.


dneto0 (commented):
Looks good to me.  Thanks for updating the headless_compute example.
Saved me from fumbling around!

src/ppx/grfx/dx12/dx12_descriptor.cpp
```diff
@@ -313,6 +313,22 @@ Result DescriptorSet::UpdateDescriptors(uint32_t writeCount, const grfx::WriteDe
             } break;
 
             case grfx::DESCRIPTOR_TYPE_STORAGE_BUFFER: {
```
wangra-google, line 315: it is mentioned in the description of this change, but i still feel that it is confusing to have use the type Descriptor_type_storage_buffer for raw buffer since the comment in grfx_enums.h also doesn't indicate this is a raw buffer. Maybe better with a new type?
```
DESCRIPTOR_TYPE_STORAGE_BUFFER         = 8,  // RW structured buffer object
```

vettoreldaniele: I will rename it to RAW_STORAGE_BUFFER. There's no use apart from this, so we're safe to rename.

vettoreldaniele:
I ended up adding support in D3D11 too. It wasn't too hard.

wangra-google (approved):

PR #16, opened 2022-11-25
[D3D12] Add XR depth swapchain and submit depth images each frame.

vettoreldaniele:
Vulkan and D3D11 coming next.

Some misc fixes too:
- Comments and formatting.
- Put all the XR settings into its own `xr` struct.
- Disabled annoying warning with MSVC for unscoped enums. They're used everywhere also due to dependencies.

### DIFF
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -209,8 +209,25 @@ struct ApplicationSettings
     bool enableDisplay         = true;
     bool enableImGui           = false;
     bool allowThirdPartyAssets = false;
-    bool enableXR              = false;
-    bool enableXRDebugCapture  = false;
+
+    struct
+    {
+        bool enable             = false;
+        bool enableDebugCapture = false;
+
+        float depthNearPlane = 0.001f;
+        float depthFarPlane  = 10000.0f;
+
+        // OpenXR uses a right-handed system.
+        // The `pos` here is the center position in view space.
+        // A detailed description can be found here:
+        // https://registry.khronos.org/OpenXR/specs/1.0/man/html/XrCompositionLayerQuad.html
+        struct
+        {
+            float3 pos;
+            float2 size;
+        } ui;
+    } xr;
 
     struct
     {
@@ -242,18 +259,6 @@ struct ApplicationSettings
             grfx::Format depthFormat = grfx::FORMAT_UNDEFINED;
             uint32_t     imageCount  = 2;
         } swapchain;
-
-#if defined(PPX_BUILD_XR)
-        // OpenXR is having a right handed system.
-        // The pos here is the center position in view space.
-        // Detailed description can be found here:
-        // https://registry.khronos.org/OpenXR/specs/1.0/man/html/XrCompositionLayerQuad.html
-        struct
-        {
-            XrVector3f  pos;
-            XrExtent2Df size;
-        } ui;
-#endif
     } grfx;
 };
 
@@ -366,6 +371,12 @@ public:
 
     const KeyState& GetKeyState(KeyCode code) const;
     float2          GetNormalizedDeviceCoordinates(int32_t x, int32_t y) const;
+
+    bool IsXrEnabled() const
+    {
+        return mSettings.xr.enable;
+    }
+
 #if defined(PPX_BUILD_XR)
     const XrComponent& GetXrComponent() const
     {
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -215,8 +215,9 @@ struct ApplicationSettings
         bool enable             = false;
         bool enableDebugCapture = false;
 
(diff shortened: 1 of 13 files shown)


include/ppx/application.h
```diff
@@ -209,8 +209,25 @@ struct ApplicationSettings
...
     bool allowThirdPartyAssets = false;
-    bool enableXR              = false;
-    bool enableXRDebugCapture  = false;
+
+    struct
+    {
+        bool enable             = false;
+        bool enableDebugCapture = false;
+
+        float depthNearPlane = 0.001f;
```
wangra-google, line 218: this is not really xr specific values (and they change at per-application bases), better to be consistent with non-xr code when we create projection matrix. Either we apply this setting for all regular projection matrices, or maybe we can just use the value from regular projecting matrices at per application bases.

projects/04_cube_xr/main.cpp
```diff
@@ -317,9 +317,8 @@ void ProjApp::Render()
         float4x4 P = glm::perspective(glm::radians(60.0f), GetWindowAspect(), 0.001f, 10000.0f);
         float4x4 V = glm::lookAt(float3(0, 0, 0), float3(0, 0, 1), float3(0, 1, 0));
 
-        // No need to wait for imageAcquiredFence since xrWaitSwapchainImage is called in AcquireNextImage.
-        if (GetSettings()->enableXR) {
-            P = GetXrComponent().GetProjectionMatrixForCurrentView(0.001f, 10000.0f);
+        if (IsXrEnabled()) {
+            P = GetXrComponent().GetProjectionMatrixForCurrentView();
```
wangra-google, line 321: as mentioned in the other comment, i think it is still better to pass the near and far as param to be consistent with the non-xr code (see line 317 in this file)

include/ppx/application.h
```diff
@@ -209,8 +209,25 @@ struct ApplicationSettings
     bool enableDisplay         = true;
     bool enableImGui           = false;
     bool allowThirdPartyAssets = false;
-    bool enableXR              = false;
-    bool enableXRDebugCapture  = false;
+
+    struct
+    {
+        bool enable             = false;
```
wangra-google, line 215: can you also add an option here to disable depth swapchain as well? this is for easier testing the perf impact of with and without the depth swapchain since it is not required by some reprojection techniques.

include/ppx/application.h
vettoreldaniele: The problem with not having them as part of the XR settings is that then there needs to be a way to pass those values to the XR begin- or end-of-frame routine. The depth info layer needs these values at every frame. I don't currently see a way to modify the render loop to have an application pass these values cleanly. 

A quick way would be to add an application-level method that the application must call every frame like `SetXrNearFarPlanes()`, but I'm not a fan. 

What do you think?

projects/04_cube_xr/main.cpp
vettoreldaniele: See comment above about this issue.

include/ppx/application.h
wangra-google: me neither, adding the extra call requires the application to know where to call this function, and calling it after BeginFrame would break the code.

How about keep the GetProjectionMatrixForCurrentView with near/far plane as input (as it is now), but rename it as GetProjectionMatrixForCurrentViewAndSetXrNearFarPlanes. And in XrComponent, we keep near/far as members, and move updating mCompositionLayerProjectionViews & & mCompositionLayerQuad & mCompositionLayerDepthInfos to EndFrame?

Agree it is still a bit ugly, and I am also open to other options. :) the reason why i mentioned it here is because that it is normal to see application have different near/far planes when they create projection matrix, with this setup, we force the application to use [0.001f, 10000.0f], if they don't, there might be some bugs that could be hard to debug.

vettoreldaniele: > How about keep the GetProjectionMatrixForCurrentView with near/far plane as input (as it is now), but rename it as GetProjectionMatrixForCurrentViewAndSetXrNearFarPlanes. And in XrComponent, we keep near/far as members, and move updating mCompositionLayerProjectionViews & & mCompositionLayerQuad & mCompositionLayerDepthInfos to EndFrame?

I think I will go for this one. It still is a bit dodgy if somebody calls `GetProjectionMatrixForCurrentViewAndSetXrNearFarPlanes` multiple times with different values, but I will add an assert to make sure this doesn't happen.

We can modify this once we have a better solution.

Thanks!

wangra-google: sounds good. thanks :)

vettoreldaniele:
Reworked quite a few things and applied suggestions. PTAL

wangra-google (approved):

PR #19, opened 2022-11-29
update doc about cmake version

RenfengLiu:
`cmake_path` is introduced in cmake version 3.20. Using older cmake will result in error.


docs/building_and_running.md
```diff
@@ -1,5 +1,5 @@
 # Building BigWheels
-The recommended build system is CMake. This repo supports both in-tree, and out-of-tree builds.
+The recommended build system is CMake(version 3.20 and above). This repo supports both in-tree, and out-of-tree builds.
```
vettoreldaniele, line 2: nit: space before parenthesis > `The recommended build system is CMake (version 3.20 and above)`.

RenfengLiu: Thanks for the review. This is done now.

PR #22, opened 2023-01-04
Disable XR Debug Capture by default

wangra-google:
Disable XR Debug Capture by default because it causes crashes while using Monado OpenXR Runtime


RenfengLiu:
Tested with Monado, it fixed the crash.

RenfengLiu (approved):

PR #37, opened 2023-02-10
Reduce API calls for CopyBufferToImage

cassiebeckley:
Batch calls for `CopyBufferToImage`. This has the most apparent benefit in Vulkan, where `vkCmdCopyBufferToImage` takes a `regions` argument and can copy all mip levels from a single buffer into the image with a single API call. However, it also reduces the number of calls in DirectX 12, by going from something like

```
StartCommand()
TransitionImageLayout()
CopyTextureRegion()
TransitionImageLayout()
EndCommand()


StartCommand()
TransitionImageLayout()
CopyTextureRegion()
TransitionImageLayout()
EndCommand()


StartCommand()
TransitionImageLayout()
CopyTextureRegion()
TransitionImageLayout()
EndCommand()


StartCommand()
TransitionImageLayout()
CopyTextureRegion()
TransitionImageLayout()
EndCommand()
```

to

```
StartCommand()
TransitionImageLayout()
CopyTextureRegion()
CopyTextureRegion()
CopyTextureRegion()
CopyTextureRegion()
TransitionImageLayout()
EndCommand()
```

which copies all mip levels with a single command buffer.

### DIFF
--- a/src/ppx/grfx/dx11/dx11_command.cpp
+++ b/src/ppx/grfx/dx11/dx11_command.cpp
@@ -470,33 +470,33 @@ void CommandBuffer::CopyBufferToImage(
     grfx::Image*                                    pDstImage)
 {
     for (auto& pCopyInfo : pCopyInfos) {
-        CopyBufferToImage(&pCopyInfo, pSrcBuffer, pDstImage);
+        CopyBufferToImage(pCopyInfo, pSrcBuffer, pDstImage);
     }
 }
 
 void CommandBuffer::CopyBufferToImage(
-    const grfx::BufferToImageCopyInfo* pCopyInfo,
+    const grfx::BufferToImageCopyInfo& pCopyInfo,
     grfx::Buffer*                      pSrcBuffer,
     grfx::Image*                       pDstImage)
 {
     dx11::args::CopyBufferToImage copyArgs = {};
 
-    copyArgs.srcBuffer.imageWidth      = pCopyInfo->srcBuffer.imageWidth;
-    copyArgs.srcBuffer.imageHeight     = pCopyInfo->srcBuffer.imageHeight;
-    copyArgs.srcBuffer.imageRowStride  = pCopyInfo->srcBuffer.imageRowStride;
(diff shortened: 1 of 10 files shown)


src/ppx/grfx/dx11/dx11_command.cpp
```diff
@@ -464,6 +464,16 @@ void CommandBuffer::CopyBufferToBuffer(
...
+    grfx::Buffer*                                   pSrcBuffer,
+    grfx::Image*                                    pDstImage)
+{
+    for (auto& pCopyInfo : pCopyInfos) {
+        CopyBufferToImage(&pCopyInfo, pSrcBuffer, pDstImage);
+    }
+}
+
 void CommandBuffer::CopyBufferToImage(
     const grfx::BufferToImageCopyInfo* pCopyInfo,
```
Keenuts, line 478: shall this be replaced with a const-ref? (now that you provide a const-ref vector api?)

src/ppx/grfx/vk/vk_command.cpp
```diff
@@ -530,36 +530,48 @@ void CommandBuffer::CopyBufferToBuffer(
...
+}
+
+void CommandBuffer::CopyBufferToImage(
+    const grfx::BufferToImageCopyInfo* pCopyInfo,
+    grfx::Buffer*                      pSrcBuffer,
+    grfx::Image*                       pDstImage)
+{
+    PPX_ASSERT_NULL_ARG(pCopyInfo);
+
+    return CopyBufferToImage({pCopyInfo}, pSrcBuffer, pDstImage);
```
Keenuts, line 574: Actually surprised this is legal, since it's doing a brace initialization of a vector with the pointer.
(Warns on clang/gcc btw).

Either add `*` or replace the function with const ref (as suggested above)

Keenuts (commented):
Thanks! left a few comments, but overall good improvement 😊

src/ppx/grfx/vk/vk_command.cpp
cassiebeckley: Done

Keenuts (approved):

PR #41, opened 2023-02-27
Add sample loading a GLTF file (along with altimeter in GLTF format)

Keenuts:
This is a sample showcasing GLTF file loading.
In its current state, this sample can also load intel's sponza (with some patching to replace png textures with DDS).
There are a lot of things to improve, this is just a baseline so we can build on top and improve it.

Main issues:
 - shaders are compiled multiple times, should not.
 - lot of assumption about the GLTF file (requires exported tangent, at least 1 material, only DDS, no GLB, no embedded GLTF.
 - PBR shader is not accurate. It's mostly so we can visualize the GLTF file. Might want to reuse the PBR shader from the mtl sample.
 - Cannot load GLB/embedded GLTF (need to add texture-from-buffer logic in BigWheels).
 - We copy each bufferview to the GPU as a "buffer". In fact we should have one memory copied from the CPU to the GPU and map Buffers on top. Need to find out how to do that with BigWheels.
 - many more


vettoreldaniele:
> Cannot load GLB/embedded GLTF (need to add texture-from-buffer logic in BigWheels).

Does this not work? https://github.com/google/bigwheels/blob/main/include/ppx/grfx/grfx_queue.h#LL84C30-L84C30

chaoticbob:
> This might, also adding the function to create a grfx::Image from a memory area is trivial.

Assuming that the dimensions, stride, and format are known ahead of time - there's already a function that does this:
https://github.com/google/bigwheels/blob/main/include/ppx/graphics_util.h#L86

There's a create method that takes an external storage pointer and doesn't copy:
https://github.com/google/bigwheels/blob/main/include/ppx/bitmap.h#L70

Keenuts:
> > This might, also adding the function to create a grfx::Image from a memory area is trivial.
> 
> Assuming that the dimensions, stride, and format are known ahead of time - there's already a function that does this: https://github.com/google/bigwheels/blob/main/include/ppx/graphics_util.h#L86
> 
> There's a Bitmap create method that takes an external storage pointer and doesn't copy: https://github.com/google/bigwheels/blob/main/include/ppx/bitmap.h#L70

Thanks for the pointer. When I added this code I used the following logic:
```cpp
gli::texture image = gli::load_dds(glb_bye_array, size);
ppxres = CreateImageFromCompressedImage(pQueue, image, ppImage, options);
```
The main reason I dropped the change for now is I cannot create a GLB/embedded GLTF file with the MSFT_texture_dds extension :/ It's a draft, so blender doen't support it, neither cgltf exported, outside of one binary MSFT provides for windows...

projects/28_gltf/main.cpp
```diff
@@ -0,0 +1,774 @@
...
+        mCamera = PerspCamera(60.0f, GetWindowAspect());
+    }
+
+    // Create descriptor pool large enough for this project
+    {
+        grfx::DescriptorPoolCreateInfo poolCreateInfo = {};
+        poolCreateInfo.uniformBuffer                  = 512;
+        poolCreateInfo.sampledImage                   = 512;
+        poolCreateInfo.sampler                        = 512;
+        poolCreateInfo.combinedImageSampler           = 512;
```
Keenuts, line 621: Need to find out how to replace that for DX11/DX12.

chaoticbob: Is that not working on DX12?

Keenuts: The combinedImageSampler is marked as unsupported (https://github.com/google/bigwheels/blob/main/src/ppx/grfx/dx12/dx12_descriptor.cpp#L44)
Actually DX11 and VK are the only 2 APIs running this sample 😅

Keenuts: Removing it fails in another location I'm fixing right now (compressed textures are created with a size of 1 or 2, while DX12 requires 4 min). Don't know if I can run the DX12 sample with sponza without that line yet.

Keenuts: So same issue as VK on DX12: pool runs out of descriptors of this type, but I cannot set a larger size than the default one.
Might need to fix this.

Keenuts: No, just this HLSL shader. I'm not even sure why storageImage + combinedImageSampler is needed. Removing one of the other causes a pool exhaustion. Now that it is a blocker I have a reason to look into it..

Keenuts: So turns out, It might be driver-related:
 - on linux, cannot run without setting both COMBINED_IMAGE_SAMPLER & STORAGE_IMAGE to 512
 - on windows (cloudtop), can run the VK version without those 2 lines.

Keenuts: Found the reason:
 - radv doesn't handle a per-descriptor pool size, but allocates more or less bytes in a single pool:
 https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/amd/vulkan/radv_descriptor_set.c#L908
 So the real issue was that we allocated more than 512 descriptors for samplers/sampled_image. But it was fine because we allocated less than 512 uniform buffers. Until we reached this limit.
 Adding any other type bumps this global limit, so that's why it worked.
 
 Bumped the descriptor limit to 1024 for the 3 types we need.

Keenuts:
Khronos does not want to push vendor extensions, meaning no DDS extension in samples, meaning no support for it for cgltf (Pr pending, but no info yet).
Fixed the sample so it can load an untained intel sponza scene.
PNG loading is way slower (68s vs 0.7s to load textures). KTX2 was a bit slower as it does transcoding, but we can stash a DDS into KTX2 and detect it using the key-value storage in the file. That might be the best option yet to have a fast loading.

projects/28_gltf/main.cpp
```diff
@@ -0,0 +1,837 @@
...
+            float4   lightPosition;              // Light's position
+            float4   eyePosition;
+        };
+
+        Scene scene                      = {};
+        scene.modelMatrix                = object.model;
+        scene.ambient                    = float4(0.3f);
+        scene.cameraViewProjectionMatrix = mCamera.GetViewProjectionMatrix();
+        scene.lightPosition              = float4(mLightPosition, 0);
+        scene.eyePosition                = glm::float4(mCamera.GetEyePosition(), 0.f);
```
vettoreldaniele, line 746: can drop the `glm::` here

projects/28_gltf/main.cpp
```diff
@@ -0,0 +1,837 @@
...
+
+        object.pUniformBuffer->CopyFromSource(sizeof(scene), &scene);
+    }
+
+    {
+        // FIXME: this assumes we have only PBR, and with 3 materials per texture. Needs to be revisited.
+        constexpr size_t                                    TEXTURE_COUNT    = 3;
+        constexpr size_t                                    DESCRIPTOR_COUNT = 1 + TEXTURE_COUNT * 2 /* uniform + 3 * (sampler + texture) */;
+        std::array<grfx::WriteDescriptor, DESCRIPTOR_COUNT> write;
+        std::memset(write.data(), 0, sizeof(write[0]) * write.size());
```
vettoreldaniele, line 756: This is already initialized through the default constructor. All `grfx::` types initialize to something sensible

projects/28_gltf/main.cpp
```diff
@@ -0,0 +1,837 @@
...
+        scene.modelMatrix                = object.model;
+        scene.ambient                    = float4(0.3f);
+        scene.cameraViewProjectionMatrix = mCamera.GetViewProjectionMatrix();
+        scene.lightPosition              = float4(mLightPosition, 0);
+        scene.eyePosition                = glm::float4(mCamera.GetEyePosition(), 0.f);
+
+        object.pUniformBuffer->CopyFromSource(sizeof(scene), &scene);
+    }
+
+    {
```
vettoreldaniele, line 751: Do we need to update the descriptors at every frame? I don't see anything that changes across frames.

projects/28_gltf/main.cpp
```diff
@@ -0,0 +1,837 @@
...
+
+    std::vector<PerFrame>        mPerFrame;
+    grfx::DescriptorPoolPtr      mDescriptorPool;
+    grfx::DescriptorSetLayoutPtr mSetLayout;
+    grfx::ShaderModulePtr        mVertexShader;
+    grfx::ShaderModulePtr        mPixelShader;
+    PerspCamera                  mCamera;
+    float3                       mLightPosition = float3(10, 100, 10);
+
+    RenderList             mRenderList;
```
vettoreldaniele, line 105: This is never used. I suppose you wanted to use this to sort by material and reduce the state switching?

vettoreldaniele (dismissed):
Do you plan on having some of the loading functionality moved to the `grfx::` namespace as a helper?

.gitmodules
```diff
@@ -40,3 +40,7 @@
 [submodule "third_party/OpenXR-SDK-Source"]
 	path = third_party/OpenXR-SDK-Source
 	url = https://github.com/KhronosGroup/OpenXR-SDK-Source.git
+[submodule "third_party/cgltf"]
+	path = third_party/cgltf
+	url = https://github.com/jkuhlmann/cgltf.git
+  branch = dds
```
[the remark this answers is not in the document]
Keenuts: forgot to remove this line, thanks!

assets/basic/shaders/PbrMetallicRoughness.hlsl
```diff
@@ -0,0 +1,148 @@
...
+
+    const float3 V = normalize(Scene.EyePosition.xyz - input.world_position.xyz);
+    const float3 normal = NormalMap.Sample(NormalMapSampler, input.uv).rgb;
+    const float3 N = normalize(mul(TBN, normal * 2.0 - 1.0));
+
+    const float4 albedo = AlbedoTexture.Sample(AlbedoSampler, input.uv).rgba;
+    if (albedo.a < 0.8f) {
+      discard;
+    }
+    const float roughness = MetalRoughness.Sample(MetalRoughnessSampler, input.uv).g;
```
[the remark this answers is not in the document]
Keenuts: No, the green one. The difference between the altimeter texture and the sponza scene is the red channel being either all-white, or all-black. But on both cases, green is roughness, blue metalness.
(looking at the asset is looks like red is used, but it's just because of the default to 1.0)

projects/28_gltf/main.cpp
```diff
@@ -0,0 +1,837 @@
...
+};
+
+void ProjApp::Config(ppx::ApplicationSettings& settings)
+{
+    settings.appName                    = "gltf";
+    settings.enableImGui                = true;
+    settings.window.width               = 1920;
+    settings.window.height              = 1080;
+    settings.grfx.api                   = kApi;
+    settings.grfx.enableDebug           = true;
```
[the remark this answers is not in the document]
Keenuts: Yes, good point.

projects/28_gltf/main.cpp
```diff
@@ -0,0 +1,837 @@
...
+        float3 color = glm::make_vec3(material.pbr_metallic_roughness.base_color_factor);
+        LoadTexture(ColorToBitmap(color), pQueue, &pOutput->textures[0]);
+    }
+    else {
+        const auto& texture_path = material.pbr_metallic_roughness.base_color_texture;
+        LoadTexture(gltfFolder, texture_path, pQueue, pTextureCache, &pOutput->textures[0]);
+    }
+
+    if (material.normal_texture.texture == nullptr) {
+        LoadTexture(ColorToBitmap(float3(0.f, 1.f, 0.f)), pQueue, &pOutput->textures[1]);
```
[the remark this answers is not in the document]
Keenuts: Was supposed to be `(0, 0, 1)` fixed! The goal was to have a neutral normal (tangent space) in case we don't have more info than the triangle's normal.
Setting it to `0, 0, 0` would not look.
3 screenshots: with normal map, without (0, 0, 1), and without (0, 0, 0):
normal
no-normal
all-zero

projects/28_gltf/main.cpp
```diff
@@ -0,0 +1,837 @@
...
+        scene.ambient                    = float4(0.3f);
+        scene.cameraViewProjectionMatrix = mCamera.GetViewProjectionMatrix();
+        scene.lightPosition              = float4(mLightPosition, 0);
+        scene.eyePosition                = glm::float4(mCamera.GetEyePosition(), 0.f);
+
+        object.pUniformBuffer->CopyFromSource(sizeof(scene), &scene);
+    }
+
+    {
+        // FIXME: this assumes we have only PBR, and with 3 materials per texture. Needs to be revisited.
```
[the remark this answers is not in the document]
Keenuts: Yes, thanks

projects/28_gltf/main.cpp
Keenuts: We don't for now (no camera movement), but we should add an orbital camera right after this is implemented.

projects/28_gltf/main.cpp
Keenuts: Ah thanks, good to know.

projects/28_gltf/main.cpp
Keenuts: Oh yes, thanks, yes iterated a bit around this, forgot to clean this up.

Keenuts:
> Do you plan on having some of the loading functionality moved to the `grfx::` namespace as a helper?

Yes, but not proactively. The idea is to see if there is some usage, how it's used, and from that derive an API/framework for the scene description.

assets/basic/shaders/PbrMetallicRoughness.hlsl
```diff
@@ -0,0 +1,148 @@
...
+    const float3 F = FresnelSchlick(F0, saturate(dot(Lh, Lo)));
+    const float  D = DistributionGGX(cosLh, roughness);
+    const float  G = GASmithSchlickGGX(cosLi, cosLo, roughness);
+
+    const float3 kD = lerp(float3(1, 1, 1) - F, float3(0, 0, 0), metalness);
+    const float3 diffuseBRDF = kD * albedo.rgb;
+    const float3 specularBRDF = (F * D * G) / max(0.00001, 4.0 * cosLi * cosLo);
+    const float3 Co = (diffuseBRDF + specularBRDF) * Lrad * cosLi + 0.3f * albedo.rgb;
+
+    return float4(saturate(Co), 1);
```
wangra-google, line 147: Any reason you want to clamp the range to [0,1]?

projects/28_gltf/main.cpp
```diff
@@ -0,0 +1,837 @@
...
+        else if (type == cgltf_attribute_type_texcoord && *ppUv == nullptr) {
+            // For UV we only load the first TEXCOORDs (FIXME: support multiple tex coordinates).
+            *ppUv = data;
+        }
+    }
+
+    PPX_ASSERT_MSG(*ppPosition != nullptr && *ppUv != nullptr && *ppNormal != nullptr && *ppTangent != nullptr, "For now, only supports model with position, normal, tangent and UV attributes");
+}
+
+void ProjApp::LoadPrimitive(const cgltf_primitive& primitive, grfx::BufferPtr pStagingBuffer, grfx::Queue* pQueue, Primitive* pOutput) const
```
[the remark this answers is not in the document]
Keenuts: added a comment on the prototype definition.

projects/28_gltf/main.cpp
```diff
@@ -0,0 +1,837 @@
...
+        Primitive*           pPrimitive;
+        grfx::DescriptorSet* pDescriptorSet;
+
+        Renderable(Material* m, Primitive* p, grfx::DescriptorSet* set)
+            : pMaterial(m), pPrimitive(p), pDescriptorSet(set) {}
+    };
+
+    struct Object
+    {
+        float4x4                model;
```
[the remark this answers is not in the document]
Keenuts: `modelMatrix` to match the other Scene struct.

projects/28_gltf/main.cpp
```diff
@@ -0,0 +1,837 @@
...
+        }
+
+        Object item;
+        item.model = compute_object_matrix(&node);
+
+        for (size_t j = 0; j < node.mesh->primitives_count; j++) {
+            const size_t primitive_index = primitiveToIndex.at(&node.mesh->primitives[j]);
+            // FIXME: support meshes with no material.
+            // For now, assign the first available material.
+            PPX_ASSERT_MSG(pMaterials->size() != 0, "Doesn't support GLTF files with no materials.");
```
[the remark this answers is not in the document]
Keenuts: Not sure that would be the best location, as it's not an issue with LoadMaterial, but with the mesh creation step, where we should lazily create some material if none is defined no?

assets/basic/shaders/PbrMetallicRoughness.hlsl
```diff
@@ -0,0 +1,148 @@
...
+    const float  cosLh = saturate(dot(N, Lh));
+
+    const float3 F = FresnelSchlick(F0, saturate(dot(Lh, Lo)));
+    const float  D = DistributionGGX(cosLh, roughness);
+    const float  G = GASmithSchlickGGX(cosLi, cosLo, roughness);
+
+    const float3 kD = lerp(float3(1, 1, 1) - F, float3(0, 0, 0), metalness);
+    const float3 diffuseBRDF = kD * albedo.rgb;
+    const float3 specularBRDF = (F * D * G) / max(0.00001, 4.0 * cosLi * cosLo);
+    const float3 Co = (diffuseBRDF + specularBRDF) * Lrad * cosLi + 0.3f * albedo.rgb;
```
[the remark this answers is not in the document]
Keenuts: Replaced 0.3 with Scene.Ambient, nice catch forgot this one

projects/28_gltf/main.cpp
```diff
@@ -0,0 +1,837 @@
...
+    PPX_CHECKED_CALL(frame.imageAcquiredFence->WaitAndReset());
+    // Wait for and reset render complete fence
+    PPX_CHECKED_CALL(frame.renderCompleteFence->WaitAndReset());
+
+    // Update camera(s)
+    mCamera.LookAt(float3(19, 1, -3), float3(0, 0, 0));
+
+    // Update uniform buffers
+    for (auto& object : mObjects) {
+        // Draw uniform buffers
```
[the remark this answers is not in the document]
Keenuts: of course, also duplicate of the comment above, removed!

assets/basic/shaders/PbrMetallicRoughness.hlsl
wangra-google: it would be wrong to normalize the output this way for pbr, should use tonemapping to convert from hdr to sdr.

assets/basic/shaders/PbrMetallicRoughness.hlsl
```diff
@@ -0,0 +1,149 @@
...
+    const float3 F = FresnelSchlick(F0, saturate(dot(Lh, Lo)));
+    const float  D = DistributionGGX(cosLh, roughness);
+    const float  G = GASmithSchlickGGX(cosLi, cosLo, roughness);
+
+    const float3 kD = lerp(float3(1, 1, 1) - F, float3(0, 0, 0), metalness);
+    const float3 diffuseBRDF = kD * albedo.rgb;
+    const float3 specularBRDF = (F * D * G) / max(0.00001, 4.0 * cosLi * cosLo);
+    const float3 Co = (diffuseBRDF + specularBRDF) * Lrad * cosLi + Scene.Ambient * albedo.rgb;
+
+    return float4(Co, 1);
```
wangra-google, line 148: you will need tonemapping to convert from hdr to sdr. rainhard should be fine.

Keenuts: one detail: this "PBR" shader is really just so we can visualize the pbr material. It's not correct.

Keenuts:
Sorry for the force-push, messed-up while rebasinb. Should be fine now.

assets/basic/shaders/PbrMetallicRoughness.hlsl
wangra-google: just one more line for rainhard: Co = Co/(Co+float3(1,1,1));

vettoreldaniele (approved):

assets/basic/shaders/PbrMetallicRoughness.hlsl
Keenuts: done, thanks!
I should invest in an HDR screen one day :smile:

wangra-google (approved):

PR #42, opened 2023-02-28
[WIP] Enable pre-record command buffer for benchmark

RenfengLiu:
Add an option `enablePreRecordCmd` to pre-record command buffer. When enabled, it will pre-record command buffer and use it for submission. 

This requires creating an array of command buffers(the same as the number of swapchain images) instead of reuse the same one for every frame.

### DIFF
--- a/src/ppx/application.cpp
+++ b/src/ppx/application.cpp
@@ -1009,6 +1009,10 @@ void Application::DispatchConfig()
         mSettings.grfx.numFramesInFlight = mSettings.grfx.swapchain.imageCount;
     }
 
+    if (mSettings.grfx.numFramesInFlight > 1) {
+        PPX_LOG_WARN("Image count in swapchain should match numFramesInFlight.");
+        mSettings.grfx.swapchain.imageCount = mSettings.grfx.numFramesInFlight;
+    }
     // Decorate DX's API name with shader bytecode mode
     std::stringstream ss;
     ss << ToString(mSettings.grfx.api);
--- a/projects/04_cube/main.cpp
+++ b/projects/04_cube/main.cpp
@@ -54,9 +54,9 @@ private:
     grfx::Viewport                    mViewport;
     grfx::Rect                        mScissorRect;
     grfx::VertexBinding               mVertexBinding;
-    void                              BuildCommandBufferForFrame(uint32_t imageIndex);
+    void                              BuildCommandBuffer(uint32_t imageIndex);
     uint32_t                          AcquireFrame(PerFrame& frame);
-    ppx::grfx::CommandBufferPtr&      GetCommandBuffer(uint32_t imageIndex);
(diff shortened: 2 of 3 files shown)


src/ppx/application.cpp
```diff
@@ -998,6 +998,21 @@ void Application::DispatchConfig()
         mSettings.window.title = mSettings.appName;
     }
 
+    // ImGui doesn't support prebuilt command buffer.
+    if (mSettings.grfx.enablePreRecordCmd) {
+        PPX_LOG_WARN("Disable ImGui when enable prebuilt command buffer.");
```
vettoreldaniele, line 1003: Can you reword the warning message to be similar to others, e.g. "ImGui does not support prebuilt command buffers, will be disabled"

src/ppx/application.cpp
```diff
@@ -998,6 +998,21 @@ void Application::DispatchConfig()
...
+        mSettings.enableImGui = false;
+    }
+
+    if (mSettings.grfx.numFramesInFlight > mSettings.grfx.swapchain.imageCount) {
+        PPX_LOG_WARN("numFramesInFlight should not exceed image count in swapchain.");
```
vettoreldaniele, line 1008: Also here, please say what action you're taking, i.e. "numFramesInFlight will be set to the swapchain image count"

src/ppx/application.cpp
```diff
@@ -998,6 +998,21 @@ void Application::DispatchConfig()
...
+        mSettings.grfx.numFramesInFlight = mSettings.grfx.swapchain.imageCount;
+    }
+
+    if (mSettings.grfx.numFramesInFlight > 1) {
```
vettoreldaniele, line 1012: I don't think setting the image count to the number of frames in flight is always desired, and it's certainly confusing for the user setting the two options. You can have X frames in flight and Y swapchain images where X != Y and still work (because you block if X < Y, and you have extra images if Y > X).

vettoreldaniele (commented):
I will wait to fully review until the PR is marked for review, but I just left some initial comments on the options/settings that it's maybe easier to deal with earlier than later.

src/ppx/application.cpp
vettoreldaniele: Yes, it would introduce additional complexity. But what I'm saying is, we're exposing these two options to the user and it's weird that we're then forcing them to match in some way. Maybe a better way is to rework the settings: an example is to only have swapchainCount and then a boolean `multipleFramesInFlight` that will automatically assume we use `num frames in flight = swapchain count`, instead of two integer options.

So you can have
```
 mSettings.grfx.swapchain.imageCount = 3;
 mSettings.grfx.multipleFramesInFlight = true;
```
which will be equivalent to the current `mSettings.grfx.numFramesInFlight  = 3`
or 
```
 mSettings.grfx.swapchain.imageCount = 3;
 mSettings.grfx.multipleFramesInFlight = false;
```

which will be equivalent to the current `mSettings.grfx.numFramesInFlight = 1`.

PR #43, opened 2023-02-28
grfx: fix issue with compressed textures

Keenuts:
DXT* documentation states an image cannot have a size which is not a multiple of the block-size.
However, it is possible to generate mipmap levels which size is less then than minimum alignment, causing a validation error.

This commit tried to fix the loading by computing the correct stride, and prevent the loading of a bogus texture.


src/ppx/graphics_util.cpp
```diff
@@ -639,26 +640,44 @@ Result CreateImageFromCompressedImage(
...
 
-        // bytesPerTexel is a bit misleading - for block compressed formats it's actually the
-        // number of bytes per tile. An image has to be made up of an integral number of tiles, so
-        // the actual width and height in memory are multiples of this value
-        ls.bufferWidth  = RoundUp<uint32_t>(ls.width, grfx::GetFormatDescription(format)->bytesPerTexel);
-        ls.bufferHeight = RoundUp<uint32_t>(ls.height, grfx::GetFormatDescription(format)->bytesPerTexel);
+        // If the DDS file contains textures which size is not a multiple of 4, something is wrong.
+        // Since imagemagick can create invalid mipmap levels, I'd assume it can also create invalid
+        // textures with non-multiple-of-4 sizes. Asserting to catch those.
+        PPX_ASSERT_MSG(ls.width % blockWidth == 0, "Width must me multiple of 4! Invalid DDS?");
```
chaoticbob, line 661: Should this just return an error when one of the dimensions isn't a multiple of 4? The assert will most likely be compiled out in release mode and it seems like a hard fail is a more obvious behavior than a possible ambiguous crash.

Keenuts: Yes, good point. Changed into an error + message

src/ppx/grfx/grfx_format.cpp
```diff
@@ -18,208 +18,168 @@
...
-    // FORMAT_BC7_SRGB,
-    {FORMAT_DATA_TYPE_SRGB, FORMAT_ASPECT_COLOR, 4, -1, FORMAT_LAYOUT_COMPRESSED, FORMAT_COMPONENT_RED_GREEN_BLUE_ALPHA, {{-1, -1, -1, -1}}},
+    // We don't support non-square blocks for compressed textures.
+    //                +-----------------------------------------------------------+
+    //                |                             ,-> bytes per block           |
+    //                |                             |  ,-> block width            |
+    //                | grfx format       | type    |  |   Components             |
+    COMPRESSED_FORMAT(BC1_RGBA_SRGB,        SRGB,   8, 4,  RED_GREEN_BLUE_ALPHA),
+    COMPRESSED_FORMAT(BC1_RGBA_UNORM,       UNORM,  8, 4,  RED_GREEN_BLUE_ALPHA),
+    COMPRESSED_FORMAT(FORMAT_BC1_RGB_SRGB,  SRGB,   8, 4,  RED_GREEN_BLUE),
```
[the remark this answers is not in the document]
Keenuts: Oh so that was a mistake.
Since it's an array, the index is important, and before there was this comment to say: this line is this grfx::FORMAT_X.
The first argument of the define here is not used. It's purely an hint for reading this table.
So made the mistake of not removing the redundant FORMAT_ prefix for this hint block.

Looking back, it was a bad idea to make it look like an argument. Changed that to a comment.

chaoticbob (approved):
LGTM

PR #47, opened 2023-03-09
Add a pack_glb utility to create glb files from GLTF

cassiebeckley:
I have another script that uses this as a library as part of the process of compressing all images and then packing them into a GLB, but this works standalone as well. I split up the commits for PR readability and because we haven't decided on exactly what extension and container format to use for compressed images.

### DIFF
--- a/tools/pack_glb.py
+++ b/tools/pack_glb.py
@@ -1,3 +1,17 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 import argparse
 import json
 import logging
--- a/tools/pack_glb.py
+++ b/tools/pack_glb.py
@@ -18,8 +18,7 @@ import logging
 import os
 from pathlib import Path
 import sys
-from typing import Any, BinaryIO
-from typing_extensions import NotRequired, TypedDict
+from typing import BinaryIO, TypedDict
 from urllib.parse import urlparse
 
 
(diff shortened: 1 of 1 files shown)


tools/pack_glb.py
```diff
@@ -0,0 +1,202 @@
...
+        # BIN chunk
+        # chunkLength
+        file.write(aligned_bin_length.to_bytes(4, byteorder="little"))
+        # chunkType 0x004E4942
+        file.write(b"BIN\x00")
+        # chunkData
+        for source_file_name in self.source_files:
+            with open(self.source_dir / source_file_name, "rb") as source_file:
+                while True:
+                    read = source_file.read(4096)
```
Keenuts, line 169: why read by 4K? Why not `file.write(source_file.read())` ?
Is it in case the buffer is larger than the ram? Or something I'm missing?
If that's the case, maybe a larger limit would be nice? less syscalls.

tools/pack_glb.py
```diff
@@ -0,0 +1,202 @@
...
+            if "uri" in buffer:
+                # this doesn't handle data uris; we can add support if necessary
+                uri = urlparse(buffer["uri"])
+                if uri.scheme != "" or uri.netloc != "":
+                    raise GLTFError(f'support for uri not implemented: "{uri}"')
+                # record the path of the locally stored external buffer to be packed
+                self.source_files.append(Path(uri.path))
+                buffer_offsets[index] = self.bin_length
+            else:
+                if index != 0:
```
Keenuts, line 93: Maybe add a comment to https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#glb-stored-buffer so we understand why the first buffer omits the property?

Keenuts (commented):
Thanks for this! A few details, but good overall.

tools/pack_glb.py
cassiebeckley: Yeah, it was intended to avoid using too much RAM. I can just remove the limit for now and add it back if we ever have issues with it

tools/pack_glb.py
```diff
@@ -0,0 +1,202 @@
...
+# limitations under the License.
+
+import argparse
+import json
+import logging
+import os
+from pathlib import Path
+import sys
+from typing import Any, BinaryIO
+from typing_extensions import NotRequired, TypedDict
```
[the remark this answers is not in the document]
cassiebeckley: Sure, I'll just use `# type: ignore` for any errors that come from `uri` being a required field.

tools/pack_glb.py
cassiebeckley: Done

Keenuts (approved):

PR #48, opened 2023-03-23
Fixed a typo in the Texture class interface

mdagois:
**GetImageFormat** was mispelled as **GeImageFormat** (missing **t**).


Keenuts:
Oh nice catch, thanks for fixing!

PR #49, opened 2023-03-23
Add an option to run an application in headless mode.

vettoreldaniele:
To run an application in headless mode, pass the `--headless` flag. All APIs are supported. Software rendering also supports this mode.

In headless mode, no window or graphics surface is created. The graphics swapchain is created and managed manually, i.e. the render targets (color images) and depth images are created explicitly. The `Present` and `AcquireNextImage` calls are also handled manually by submitting empty command buffers that signal or wait for semaphores as needed.

Additional changes:
* Explicitly set severity of D3D12 debug message, and ignore the non-matching clear value warnings that cannot be avoided in headless mode. They are noisy, performance-only warnings.
* Add a CMake option to the Linux build to avoid linking to XCB if only headless execution is desired (`-DPPX_LINUX_HEADLESS`).

### DIFF
--- a/benchmarks/headless_compute/main.cpp
+++ b/benchmarks/headless_compute/main.cpp
@@ -88,7 +88,7 @@ private:
 void ProjApp::Config(ppx::ApplicationSettings& settings)
 {
     settings.appName                        = "compute_operations";
-    settings.enableDisplay                  = false;
+    settings.headless                       = true;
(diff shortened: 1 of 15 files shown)


chaoticbob (approved):
LGTM. 

Looked this over a couple of times and nothing jumped out at me.

PR #53, opened 2023-03-27
Rework Bitmap::Fill to avoid undefined behavior

vettoreldaniele:
Currently, `Bitmap::Fill` takes a float array and converts to integer when needed. This is undefined behavior as the desired integer value may not be representable in floating point.

This PR changes `Bitmap::Fill` to take a templated type instead. The helper `CreateTexture1x1` is also changed to allow the caller to specify the data type (which will be used to determine the type of the underlying texture too).

### DIFF
--- a/include/ppx/bitmap.h
+++ b/include/ppx/bitmap.h
@@ -86,7 +86,8 @@ public:
     Result Resize(uint32_t width, uint32_t height);
     Result ScaleTo(Bitmap* pTargetBitmap) const;
 
-    void Fill(float r, float g, float b, float a);
+    template <typename PixelDataType>
+    void Fill(PixelDataType r, PixelDataType g, PixelDataType b, PixelDataType a);
 
     // Returns byte address of pixel at (x,y)
     char*       GetPixelAddress(uint32_t x, uint32_t y);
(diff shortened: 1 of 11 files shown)


include/ppx/bitmap.h
```diff
@@ -195,6 +196,28 @@ class Bitmap
...
+
+    PixelDataType rgba[4] = {r, g, b, a};
+
+    const uint32_t channelCount = Bitmap::ChannelCount(mFormat);
+
+    char* pData = mData;
+    for (uint32_t y = 0; y < mHeight; ++y) {
+        for (uint32_t x = 0; x < mWidth; ++x) {
+            PixelDataType* pPixelData = reinterpret_cast<PixelDataType*>(pData);
+            for (uint32_t c = 0; c < channelCount; ++c) {
```
Keenuts, line 213: memcpy?

Keenuts (approved):
the for loop is also readable, so maybe memcpy is not better

include/ppx/bitmap.h
vettoreldaniele: I'll leave the for just in this case because I think it's a little more readable and I bet the compiler optimizes to the same thing.

PR #55, opened 2023-03-28
Use filesystem::path for building screenshot path

vettoreldaniele:


dnovillo (approved):
I think I've seen similar patterns in the `projects/` directory.  But that does not need to be done here.

PR #58, opened 2023-03-30
Added the first version of the order-independent transparency sample

mdagois:
Supported algorithms:

* Unsorted over
* Weighted sum
* Weighted average
  * Fragment count
  * Exact coverage


projects/oit_demo/WeightedAverageWithCoverage.cpp
vettoreldaniele, line 1: Given the similarities between weighted average and weighted average with coverage, is it possible to share the common code and have the code take different paths only when necessary?

vettoreldaniele (commented):
Thanks for the sample!

assets/oit_demo/shaders/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+
+generate_rules_for_shader(
+    "composite"
+    SOURCE "${PPX_DIR}/assets/oit_demo/shaders/Composite.hlsl"
+    INCLUDES ${INCLUDE_FILES}
+    STAGES "ps" "vs")
+
+generate_group_rule_for_shader(
+    "shader_oit_demo"
+    CHILDREN
```
Keenuts, line 68: you are missing the other shaders here. Won't build on a clean build.

assets/oit_demo/shaders/Composite.hlsl
```diff
@@ -0,0 +1,34 @@
...
+#include "FullscreenVS.hlsli"
+
+Texture2D    OpaqueTexture       : register(OPAQUE_TEXTURE_REGISTER);
+Texture2D    TransparencyTexture : register(TRANSPARENCY_TEXTURE_REGISTER);
+
+float4 psmain(VSOutput input) : SV_TARGET
+{
+    const int3 texelCoord = int3(input.position.xy, 0);
+
+    const float3 opaqueColor = OpaqueTexture.Load(texelCoord).rgb;
```
Keenuts, line 26: This causes validation errors on Vulkan:
 - this load requires StorageImageReadWithoutFormat,
 - the shader pipeline is build without this capability enabled.
 Not sure how to do that for BigWheels. Not sure of the implications to always enable this capability?

assets/oit_demo/shaders/FullscreenVS.hlsli
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+
+struct VSOutput
+{
+    float4 position : SV_POSITION;
+    float2 uv       : TEXCOORD;
+};
+
+VSOutput vsmain(uint VertexID : SV_VertexID)
+{
```
Keenuts, line 22: Could `basic/shaders/FullScreenTriangle.hlsl` be reused? or is it different?

Keenuts (commented):
Thanks for this, cool sample!

projects/oit_demo/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Background
+
+Transparency rendering is a technique used to render objects that partially let light go through them.
+It is heavily used in games to represent glass, fluids (e.g. smoke) and elements that have no physical equivalent in reality (e.g. magic).
+It is also often used in computer-aided design to let creators see through multiple layers of architectural and mechanical designs.
+
+Transparency is typically achieved by alpha blending fragments from back to front from the camera's perspective.
+Technically, this means sorting the geometry on the CPU before sending it to the GPU for rendering.
+However, in addition to having a non-negligeable CPU cost, that solution cannot produce accurate results in several situations, e.g. intersecting geometry or meshes with concave hulls.
```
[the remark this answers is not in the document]
mdagois: Nice catch. Fixed.

projects/oit_demo/OITDemoApplication.cpp
```diff
@@ -0,0 +1,478 @@
...
+{
+    AddSupportedAlgorithm("Unsorted over", ALGORITHM_UNSORTED_OVER);
+    AddSupportedAlgorithm("Weighted sum", ALGORITHM_WEIGHTED_SUM);
+    if (GetDevice()->IndependentBlendingSupported()) {
+        AddSupportedAlgorithm("Weighted average", ALGORITHM_WEIGHTED_AVERAGE);
+        AddSupportedAlgorithm("Weighted average with coverage", ALGORITHM_WEIGHTED_AVERAGE_WITH_COVERAGE);
+    }
+}
+
+void OITDemoApp::Setup()
```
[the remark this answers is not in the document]
mdagois: I added a command line option, **--algorithm**, that takes the algorithm ID. I updated the README to mention it.

projects/oit_demo/WeightedAverageWithCoverage.cpp
mdagois: You are right. They are variations of the same algorithm. There is even one more variation I want to add later.

I merged the shader code and C++ code of the two algorithms into a single algorithm with two variations: **fragment count** and **exact coverage**.
In the GUI, it is possible to switch between variations via a combo box under the **weighted average** algorithm.

Note that both variations yields very similar results for now, but I will had mesh/situations that highlights the differences.

assets/oit_demo/shaders/CMakeLists.txt
mdagois: Ah I forgot to add them. Fixed. Thanks!

assets/oit_demo/shaders/Composite.hlsl
mdagois: I don't get that error in Vulkan (I mostly develop in Vulkan with debug at true).
I use 2 different AMD GPUs: WX3200 and RX6700XT.
Maybe my driver is more permissive than yours?

It makes sense that it happens though.
I changed that **Load** into a regular **Sample**.
I believe it should fix the issue on your driver.
Please let me know if you encounter more of those.

assets/oit_demo/shaders/FullscreenVS.hlsli
mdagois: The vertex shader is equivalent, but I use several different pixel shaders.
Also, I might actually have to add more interpolated values for other algorithms I plan to add.
Those interpolation would not make sense in the generic **FullScreenTriangle.hlsl**.
So, I'd rather keep my code for fullscreen passes separated for now, if that's ok.

assets/oit_demo/shaders/Composite.hlsl
Keenuts: Ah yes, that's better, no need to add the capability. Btw, all Loads will generate this capability, so the others also need patching.

I can reproduce on my workstation (WX 3200), not if I use the AMD gpu, but if I use LLVMpipe.
(forced the gpu index to 1, which is llvmpipe for me).

vettoreldaniele (approved):

assets/oit_demo/shaders/Composite.hlsl
mdagois: Ah I see!
I thought it was related to the image format (uint while others are float), but it is more of a software driver limitation.
I have another PR coming soon.
I will remove all loads from it, as sampling with a nearest sampler just works fine.

Keenuts (approved):
Thanks, all validation errors are gone 😊

PR #60, opened 2023-03-30
Remove GLSL fluid simulation shaders

vettoreldaniele:


dnovillo (approved):
Oops.  Thanks for catching this!

PR #62, opened 2023-03-31
Remove GGP support

vettoreldaniele:
This also allows us to go back to the mainline `glfw` repository which did not have GGP support.


vettoreldaniele:
> Thank you!
> 
> This will require changes to the internal importer, right?

Why? I don't think so unless I'm missing something?

dnovillo:
> > Thank you!
> > This will require changes to the internal importer, right?
> 
> Why? I don't think so unless I'm missing something?

The location of glfw.  Or is the internal mirror already using upstream?

vettoreldaniele:
> 



> > > Thank you!
> > > This will require changes to the internal importer, right?
> > 
> > 
> > Why? I don't think so unless I'm missing something?
> 
> The location of glfw. Or is the internal mirror already using upstream?

Already using upstream!

dnovillo:
> > 
> 
> > > > Thank you!
> > > > This will require changes to the internal importer, right?
> > > 
> > > 
> > > Why? I don't think so unless I'm missing something?
> > 
> > 
> > The location of glfw. Or is the internal mirror already using upstream?
> 
> Already using upstream!

Excellent!  Thanks :)

PR #65, opened 2023-03-31
Upload screenshots as artifacts as part of CI runs

vettoreldaniele:


.github/workflows/linux-build.yml
```diff
@@ -47,7 +47,20 @@ jobs:
       - name: Run runtime tests
         run: |
           cd build/bin
-          xvfb-run -a ./vk_03_square_textured --frame-count 2
-          xvfb-run -a ./vk_09_obj_geometry --frame-count 2
-          xvfb-run -a ./vk_13_normal_map --frame-count 2
-          xvfb-run -a ./vk_fishtornado --frame-count 2
+
+          xvfb-run -a ./vk_03_square_textured --frame-count 2 --screenshot-frame-number 1
```
Keenuts, line 51: What about storing frame 0?
When we test, we don't really see the 1st frame since the window get's initialized.
So we might have broken things during the first step (maybe some uniform are not initialized), and this will allow to catch those no?

Keenuts (approved):
Nice addition! Will ease the review process!

PR #68, opened 2023-04-04
cli: add --screenshot-filepath option

Keenuts:
BigWheels allow setting the output directory, but generates the filename.
This limits our options as we cannot write to arbitrary files (memfd?) to integrate bigwheels into other applications.

### DIFF
--- a/src/ppx/application.cpp
+++ b/src/ppx/application.cpp
@@ -1128,8 +1128,14 @@ void Application::TakeScreenshot()
     unsigned char* texels = nullptr;
     screenshotBuf->MapMemory(0, (void**)&texels);
 
-    std::string filename = "screenshot_frame" + std::to_string(mFrameCount) + ".ppm";
-    std::string filepath = (std::filesystem::path(mStandardOptions.screenshot_out_dir) / filename).string();
+    std::string filepath;
+    if (mStandardOptions.screenshot_filepath.empty()) {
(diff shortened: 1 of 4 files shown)


Keenuts:
Yeah, agree. I guess it was to allow multiple screenshots per run?

src/ppx/application.cpp
```diff
@@ -1128,14 +1128,9 @@ void Application::TakeScreenshot()
...
-    if (mStandardOptions.screenshot_filepath.empty()) {
-        std::string filename = "screenshot_frame" + std::to_string(mFrameCount) + ".ppm";
-        filepath             = (std::filesystem::path(mStandardOptions.screenshot_out_dir) / filename).string();
-    }
-    else {
-        filepath = std::filesystem::path(mStandardOptions.screenshot_filepath).string();
-    }
+    std::string filepath = mStandardOptions.screenshot_path.empty()
+                               ? "screenshot_frame" + std::to_string(mFrameCount) + ".ppm"
+                               : mStandardOptions.screenshot_path;
```
vettoreldaniele, line 1133: Should we add ".ppm" if it's missing?

Keenuts: In favor of a no.
-> it adds complexity ("complexity" yes 😊)
-> if the user wants it, they can specify it.
-> this way we don't restrict the filename used, maybe the user wraps a script calling `mktemp()` or another function and just wants us to write the content to this file without extensions?

vettoreldaniele (approved):

PR #70, opened 2023-04-04
Copy the glft project as the start point for benchmark.

RenfengLiu:
This PR copies the `28_gltf` sample and rename it as benchmark, so that people can start to modify the sample for the benchmark implementation.


wangra-google (approved):

Keenuts:
Shouldn't this sample be moved to the `benchmarks` directory instead?
`benchmarks/graphic_pipeline` by ex?

PR #71, opened 2023-04-05
Fixed shader array support in Vulkan

mdagois:
In Vulkan, arrays are flattened at descriptor layout creation, i.e. they are turned into as many descriptors of size 1 as the size of the original array. However, flattened arrays were not fully supported in the rest of code. This PR addresses the issues with two following fixes.

* Added a compiler option to flatten arrays
* Took "arrayIndex" into account when updating the descriptor sets


Keenuts (approved):
Thanks for fixing.

PR #72, opened 2023-04-05
benchmark: add ability to change shader on the fly

Keenuts:
This is a first requirement for this benchmark: allow us to have multiple shaders running and swap them at runtime, or not.

The PBR code is taken from the 28_gltf sample, but common parts are moved to a shared file.
The idea is to share the same set of inputs across all shaders for this benchmark, simplifying the pipeline layout handling.

### DIFF
--- a/projects/benchmark/CMakeLists.txt
+++ b/projects/benchmark/CMakeLists.txt
@@ -18,4 +18,6 @@ project(benchmark)
 add_samples_for_all_apis(
     NAME ${PROJECT_NAME}
     SOURCES "main.cpp"
-    SHADER_DEPENDENCIES "shader_pbr_metallic_roughness")
+    SHADER_DEPENDENCIES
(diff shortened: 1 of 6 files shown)


projects/benchmark/CMakeLists.txt
vettoreldaniele, line 1: We have a `benchmarks` folder separate from `projects` for benchmarks. And in fact the assets are under `assets/benchmarks` which is normally used by applications under `benchmarks`. It's ok to have here for now but we should think about how to organize the two folders best moving forward

vettoreldaniele (approved):

projects/benchmark/CMakeLists.txt
Keenuts: Yes, I actually think this sample should be moved to `benchmarks/` as it is a benchmark, not a sample to show bigwheels features

PR #73, opened 2023-04-05
move benchmark sample to benchmarks folder

Keenuts:
This is not a sample to showcase bigwheels/explain, but an actual benchmark tool. Should be moved to the proper location.


benchmarks/graphic_pipeline/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
 
-project(benchmark)
+project(graphic_pipeline)
```
vettoreldaniele, line 16: nit: graphic -> graphics

Keenuts: fixed, moved the sample to 'graphics_pipeline'

vettoreldaniele (approved):

PR #85, opened 2023-04-11
build: add deterministic build flavor

Keenuts:
Samples relies on time to compute animations. This means the actual output might vary slightly between runs, adding noise to any measure. This commits adds a build flavor to only allow deterministic behaviors.

Right now, only the timer seem to be a source of non-determinism. Replaced with `frame_count / 60.f` (60fps target) on such builds.

### DIFF
--- a/src/ppx/application.cpp
+++ b/src/ppx/application.cpp
@@ -1682,7 +1682,11 @@ Result Application::CreateShader(const std::filesystem::path& baseDir, const std
 
 float Application::GetElapsedSeconds() const
 {
+#if defined(DETERMINICTIC)
+    return static_cast<float>(mFrameCount * (1.f / 60.f));
+#else
     return static_cast<float>(mTimer.SecondsSinceStart());
+#endif
 }
 
 const KeyState& Application::GetKeyState(KeyCode code) const
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 # Options.
 # ------------------------------------------------------------------------------
 option(PPX_BUILD_PROJECTS "Build sample projets" ON)
+option(PPX_DETERMINISTIC "Makes BigWheels deterministic" OFF)
 
 # ------------------------------------------------------------------------------
 # Detect DXC presence. This is REQUIRED to compile DXIL and SPIR-V shaders.
@@ -155,6 +156,11 @@ else()
         -Wno-nullability-completeness \
         -Wno-deprecated-anon-enum-enum-conversion")
 endif()
+
+if (PPX_DETERMINISTIC)
+    message("Building deterministic flavor of ${PROJECT_NAME}")
+    add_definitions("-DDETERMINISTIC")
+endif ()
 
 set(CMAKE_CXX_STANDARD 20)
 


src/ppx/application.cpp
```diff
@@ -1682,7 +1682,11 @@ Result Application::CreateShader(const std::filesystem::path& baseDir, const std
 
 float Application::GetElapsedSeconds() const
 {
+#if defined(DETERMINICTIC)
```
vettoreldaniele, line 1685: misspelled define name

vettoreldaniele (changes requested):
What about a runtime command line flag (on top of a build flag, potentially)?

src/ppx/application.cpp
vettoreldaniele: We also usually bring the PPX prefix to defines in the source code too.

Keenuts:
> What about a runtime command line flag (on top of a build flag, potentially)?

I thought of that first, but then realized it might not be the best option for simplicity.
Compile-time switch means we don't need to worry about initialization order (we should parse all flags before calling any method impacted non-deterministic side effects by ex?)
Or we could disable pthread linking, to make sure we have no multi-threading enabled (if this becomes a need)

vettoreldaniele:
> > What about a runtime command line flag (on top of a build flag, potentially)?
> 
> I thought of that first, but then realized it might not be the best option for simplicity. Compile-time switch means we don't need to worry about initialization order (we should parse all flags before calling any method impacted non-deterministic side effects by ex?) Or we could disable pthread linking, to make sure we have no multi-threading enabled (if this becomes a need)

I think just having the parsing in the Application initialization code is enough to guarantee that the majority of sane use cases will be covered. The Application class is normally a singleton that is initialized on a single thread. Whatever the application will do after, even if multi threaded, will use the deterministic clock.

Keenuts:
> > > What about a runtime command line flag (on top of a build flag, potentially)?
> > 
> > 
> > I thought of that first, but then realized it might not be the best option for simplicity. Compile-time switch means we don't need to worry about initialization order (we should parse all flags before calling any method impacted non-deterministic side effects by ex?) Or we could disable pthread linking, to make sure we have no multi-threading enabled (if this becomes a need)
> 
> I think just having the parsing in the Application initialization code is enough to guarantee that the majority of sane use cases will be covered. The Application class is normally a singleton that is initialized on a single thread. Whatever the application will do after, even if multi threaded, will use the deterministic clock.

For the threads, I meant we could completely remove the thread support at build time, as multi-threading brings non-determinism.

vettoreldaniele (approved):
Thanks for the patch!

I don't think we should worry too much about threading for now. Especially for graphics, multithreading does not necessarily (and should not, for a correct application) imply non-determinism. Future automated testing should help in that regard.

src/ppx/application.cpp
```diff
@@ -1682,6 +1682,9 @@ Result Application::CreateShader(const std::filesystem::path& baseDir, const std
 
 float Application::GetElapsedSeconds() const
 {
+    if (mStandardOptions.deterministic) {
+        return static_cast<float>(mFrameCount * (1.f / 60.f));
```
vettoreldaniele, line 1686: I think this may end up non-deterministic across machines due to different rounding modes or other FP flags - but I think a better solution can wait.

Keenuts: Do you mean if we recompile, or the same binary run on 2 machines?
I would be surprised if running the same binary on 2 machines yields different rounding operations. Isn't float rounding spec'd?

Keenuts: (goal here is for multiple runs of the same binary to be deterministic, not across builds)

vettoreldaniele: Rounding mode can be set both at runtime (https://en.cppreference.com/w/c/numeric/fenv/feround) or at compile time (with compiler flags). I don't know if there are more knobs at the system level that can play a role. I don't know how much of a problem this is but it's worth keeping in mind as a potential source of non-determinism.

vettoreldaniele: If the goal is same binary on the same system, we are good. Same binary on different systems may be a problem yet uncommon I bet (e.g. a library calling the rounding mode setter or a system-level knob, if it exists).

s-perron: > Isn't float rounding spec'd?

That depends on the compile time options that are used. If you are using any form of fast math, the compiler could change the order of operations, to give different results. Even on the same machine, clang and gcc could return different results.

However, I would not worry about that for now.

Keenuts: > That depends on the compile time options that are used.

That for sure, but I meant if x86's divss & cvttss2si producing the same result on AMD or M1 through rosetta by ex. But yes, agree it's not that important for now.

> Rounding mode can be set both at runtime (https://en.cppreference.com/w/c/numeric/fenv/feround) 

Oh didn't knew that! thanks!

PR #86, opened 2023-04-11
materials: add flags to select default aspect

Keenuts:
The sample can show some models using different shaders/materials. This commit adds command line options to define the default selection of shader/material/mesh to render.
This is useful to control the behavior of the app in headless mode.

### DIFF
--- a/projects/15_basic_material/main.cpp
+++ b/projects/15_basic_material/main.cpp
@@ -481,11 +481,11 @@ void ProjApp::Setup()
     }
 
     const auto& cl_options = GetExtraOptions();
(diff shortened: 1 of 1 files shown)


projects/15_basic_material/main.cpp
```diff
@@ -480,6 +480,14 @@ void ProjApp::Setup()
         }
     }
 
+    const auto& cl_options = GetExtraOptions();
+    mMaterialIndex         = cl_options.GetExtraOptionValueOrDefault<int32_t>("material-index", 0);
```
vettoreldaniele, line 484: You can use `uint32_t` since the index variables are unsigned

Keenuts: Ah yes, thanks!

PR #87, opened 2023-04-11
Update setup-windows10-sdk action to latest version

vettoreldaniele:
Fixes a Github Actions warning.


Keenuts:
What was the warning btw? for future reference if we find this again?

PR #91, opened 2023-04-13
Added a new algorithm to the OIT sample

mdagois:
* Added the first buffer-based algorithm (buckets)
* Added a way to find out if the hardware supports fragment shader stores and atomics
* Fixed an issue that was preventing the creation of a draw pass without depth/stencil
* Performed some code clean up based on the previous PR comments

### DIFF
--- a/projects/oit_demo/Buffer.cpp
+++ b/projects/oit_demo/Buffer.cpp
@@ -16,6 +16,8 @@
 
 void OITDemoApp::SetupBuffer()
 {
+    mBuffer.countTextureNeedClear = true;
+
     // Count texture
     {
         grfx::TextureCreateInfo createInfo         = {};
@@ -67,12 +69,11 @@ void OITDemoApp::SetupBuffer()
 
     // Gather pass
     {
-        grfx::DrawPassCreateInfo2 createInfo  = {};
-        createInfo.width                      = mBuffer.countTexture->GetWidth();
-        createInfo.height                     = mBuffer.countTexture->GetHeight();
-        createInfo.renderTargetCount          = 0;
-        createInfo.pDepthStencilImage         = nullptr;
-        createInfo.renderTargetClearValues[0] = {0, 0, 0, 0};
+        grfx::DrawPassCreateInfo2 createInfo = {};
+        createInfo.width                     = mBuffer.countTexture->GetWidth();
+        createInfo.height                    = mBuffer.countTexture->GetHeight();
+        createInfo.renderTargetCount         = 0;
+        createInfo.pDepthStencilImage        = nullptr;
         PPX_CHECKED_CALL(GetDevice()->CreateDrawPass(&createInfo, &mBuffer.gatherPass));
     }
 
@@ -91,7 +92,7 @@ void OITDemoApp::SetupBuffer()
 
         PPX_CHECKED_CALL(GetDevice()->AllocateDescriptorSet(mDescriptorPool, mBuffer.gatherDescriptorSetLayout, &mBuffer.gatherDescriptorSet));
 
-        grfx::WriteDescriptor writes[4] = {};
+        std::array<grfx::WriteDescriptor, 4> writes = {};
 
         writes[0].binding      = SHADER_GLOBALS_REGISTER;
         writes[0].type         = grfx::DESCRIPTOR_TYPE_UNIFORM_BUFFER;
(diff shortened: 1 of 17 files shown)


projects/oit_demo/Buffer.cpp
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    // Gather pass
+    {
+        grfx::DrawPassCreateInfo2 createInfo  = {};
+        createInfo.width                      = mBuffer.countTexture->GetWidth();
+        createInfo.height                     = mBuffer.countTexture->GetHeight();
+        createInfo.renderTargetCount          = 0;
+        createInfo.pDepthStencilImage         = nullptr;
+        createInfo.renderTargetClearValues[0] = {0, 0, 0, 0};
```
vettoreldaniele, line 75: Since we have no render targets, you can omit this

projects/oit_demo/Buffer.cpp
```diff
@@ -0,0 +1,294 @@
...
+        grfx::DescriptorSetLayoutCreateInfo layoutCreateInfo = {};
+        layoutCreateInfo.bindings.push_back(grfx::DescriptorBinding{SHADER_GLOBALS_REGISTER, grfx::DESCRIPTOR_TYPE_UNIFORM_BUFFER, 1, grfx::SHADER_STAGE_ALL_GRAPHICS});
+        layoutCreateInfo.bindings.push_back(grfx::DescriptorBinding{CUSTOM_TEXTURE_0_REGISTER, grfx::DESCRIPTOR_TYPE_SAMPLED_IMAGE, 1, grfx::SHADER_STAGE_ALL_GRAPHICS});
+        layoutCreateInfo.bindings.push_back(grfx::DescriptorBinding{CUSTOM_UAV_0_REGISTER, grfx::DESCRIPTOR_TYPE_STORAGE_IMAGE, 1, grfx::SHADER_STAGE_ALL_GRAPHICS});
+        layoutCreateInfo.bindings.push_back(grfx::DescriptorBinding{CUSTOM_UAV_1_REGISTER, grfx::DESCRIPTOR_TYPE_STORAGE_IMAGE, 1, grfx::SHADER_STAGE_ALL_GRAPHICS});
+        PPX_CHECKED_CALL(GetDevice()->CreateDescriptorSetLayout(&layoutCreateInfo, &mBuffer.gatherDescriptorSetLayout));
+
+        PPX_CHECKED_CALL(GetDevice()->AllocateDescriptorSet(mDescriptorPool, mBuffer.gatherDescriptorSetLayout, &mBuffer.gatherDescriptorSet));
+
+        grfx::WriteDescriptor writes[4] = {};
```
vettoreldaniele, line 94: Use `std::array<WriteDescriptor, 4>` so you can do `.size()` later.  Same for similar uses below

projects/oit_demo/Buffer.cpp
```diff
@@ -0,0 +1,294 @@
...
+        PPX_CHECKED_CALL(GetDevice()->CreateGraphicsPipeline(&gpCreateInfo, &mBuffer.combinePipeline));
+
+        GetDevice()->DestroyShaderModule(VS);
+        GetDevice()->DestroyShaderModule(PS);
+    }
+}
+
+void OITDemoApp::RecordBuffer()
+{
+    static bool sTextureNeedClear = true;
```
vettoreldaniele, line 223: Can you move this inside of `mBuffer` to avoid using static?

projects/oit_demo/Buffer.cpp
```diff
@@ -0,0 +1,294 @@
...
+            grfx::RESOURCE_STATE_RENDER_TARGET,
+            grfx::RESOURCE_STATE_SHADER_RESOURCE,
+            grfx::RESOURCE_STATE_SHADER_RESOURCE,
+            grfx::RESOURCE_STATE_SHADER_RESOURCE);
+
+        sTextureNeedClear = false;
+    }
+
+    {
+        mCommandBuffer->TransitionImageLayout(mBuffer.countTexture, 0, 1, 0, 1, grfx::RESOURCE_STATE_SHADER_RESOURCE, grfx::RESOURCE_STATE_GENERAL);
```
vettoreldaniele, line 248: We generally use `PPX_ALL_SUBRESOURCES`, e.g.:
```
mCommandBuffer->TransitionImageLayout(mBuffer.countTexture, PPX_ALL_SUBRESOURCES, grfx::RESOURCE_STATE_SHADER_RESOURCE, grfx::RESOURCE_STATE_GENERAL);
```

vettoreldaniele (changes requested):

projects/oit_demo/Buffer.cpp
mdagois: I moved the variable into mBuffer. Fixed.

projects/oit_demo/Buffer.cpp
mdagois: I switched to using std::array for WriteDescriptors in all algorithms.

vettoreldaniele:
Seems there is a conflicting file (probably needs rebase?)

mdagois:
> Seems there is a conflicting file (probably needs rebase?)

Someone had fixed a warning I had also fixed. I rebased to the latest main. That should do the trick.

vettoreldaniele:
I think the README update is missing, otherwise looks fine to me

mdagois:
> I think the README update is missing, otherwise looks fine to me

Oh my! You are right. I forgpt the reafme. I will update it tomorrow and push again. :)

mdagois:
I updated the README. Please have a look.

Issue #80, opened 2023-04-07
android: allow reusing Android files for multiple samples


Keenuts:
Right now, the chosen sample is hard-coded in the manifest. We should find a way to reuse those. Probably with a manifest template, generating the real manifest at build time.
Or maybe we should pack all samples in one APK and have a way to select the one we start (each sample being an activity entrypoint?)

at-ninja:
This should be doable with [build variants](https://developer.android.com/build/build-variants) to allow choosing a project from the command line or Android Studio. I'm happy to take this issue.

Keenuts:
Just so you know, I'm working on a PR to cleanup the android build, which will change it quite a lot. If everything goes well, the manifest will have en entry to determine the build.
So should be fine with build variant, just don't invest too much time right now?

Keenuts:
@at-ninja while cleaning up the android build, I found a way to not hardcode the library name in the MainActivity and manifest, thanks to your suggestion.
So seems like what will be needed it create multiple build variants.
And maybe tweak the build so we don't embedded all the .so in the APK, but only the one we load.

Keenuts:
> Just a high level comment. For Android, each sample application should target its own APK and be separately debuggable from the IDE.

Yes, that's what I intend to have. What I meant by not hardcoding is not to have 1 manifest by app in the repo. But of course, once compiled, we get 1 sample per APK.

apazylbe:
@Keenuts is this fixed with https://github.com/google/bigwheels/pull/101 ?

Keenuts:
Yes, thanks, forgot to close!

PR #94, opened 2023-04-13
android: cleanup android build

### LINKED ISSUE #75: Cleanup root folder from optional Android files
https://github.com/google/bigwheels/pull/46 brought Android support, but cluttered the project. We should be able to remove many optional things (custom theme, proguard, etc)

Keenuts:
This commit removes most unused files for an Android build, and tried
to simplify the build.
It gives a starting point for #80 as is provides build variants and
removes sample name duplication in the manifest and activity.

Fixes #75


Keenuts:
This CL shouldn't be merged until https://github.com/google/bigwheels/pull/93 is merged (as it's based on this commit)

Keenuts:
What we should end-up with is probably a mix, as there are some good ideas on how to handle the multi-projects on the other PR.
Main incompatibility is around the library: this project still keeps all projects as independent .so. Meaning the activity could load any of these.

Keenuts:
Created a new PR to "rebase" #90 on top of this (#101).

local.properties
vettoreldaniele, line 1: This should be in the git ignore?

vettoreldaniele (commented):
Seems reasonable to me but I don't have experience with Android builds - maybe ask Shan to double check these?

local.properties
Keenuts: Yes, missed that, thanks!

vettoreldaniele (approved):
It's probably time to add a Gradle build to the CI (after this PR series) :)

https://docs.gradle.org/current/userguide/github-actions.html

build.gradle
apazylbe, line 1: I am just curious, is there a reason to move these android related files (build.gradle, gradle.properties, gradlew, settings.gradle) out of android directory?

shanminchao:
> Seems reasonable to me but I don't have experience with Android builds - maybe ask Shan to double check these?

I think Nathan understands the Android build system stuff way better than I do by now. :)

His flattening of the build structure fixes a lot of nastiness with my initial commits, and imo it's much cleaner now.
I also verified this PR works as expected on my Pixel 6.

PR #96, opened 2023-04-13
Remove FXC/DXBC support in favor of DXIL-only for DX12

vettoreldaniele:
Further simplifies the framework, and asset handling.


vettoreldaniele:
I've uncovered a bug in our CMake rules that cause duplicate targets which rarely created problems in the past but easier to hit now that there are less targets. I'll work on fixing that and unblock the CI.

vettoreldaniele:
> I've uncovered a bug in our CMake rules that cause duplicate targets which rarely created problems in the past but easier to hit now that there are less targets. I'll work on fixing that and unblock the CI.

I've made some progress in identifying the problem but it's very annoying. We use `MAIN_DEPENDENCY` to create shader compile commands for Visual Studio, which forces VS to use the custom command instead of trying to "compile" the shader using the built-in VS function (which uses FXC). The problem is that CMake says you can only ever use one file as a `MAIN_DEPENDENCY`. Since we compile the same HLSL shader multiple times (one for each stage) this creates duplicate targets. I'm thinking of a solution.

chaoticbob:
> > I've uncovered a bug in our CMake rules that cause duplicate targets which rarely created problems in the past but easier to hit now that there are less targets. I'll work on fixing that and unblock the CI.
> 
> I've made some progress in identifying the problem but it's very annoying. We use `MAIN_DEPENDENCY` to create shader compile commands for Visual Studio, which forces VS to use the custom command instead of trying to "compile" the shader using the built-in VS function (which uses FXC). The problem is that CMake says you can only ever use one file as a `MAIN_DEPENDENCY`. Since we compile the same HLSL shader multiple times (one for each stage) this creates duplicate targets. I'm thinking of a solution.

The old/original version of the CMake rules got around this by assigning specific tags to each *target*. For instance if the high level shader target was `my_vs_shader` it would become `my_vs_shader_dxil` or `my_vs_shader_spv`. This of course meant you couldn't have two high level targets named the samething - but that's probably not too different from what's happening now.

vettoreldaniele:
> > > I've uncovered a bug in our CMake rules that cause duplicate targets which rarely created problems in the past but easier to hit now that there are less targets. I'll work on fixing that and unblock the CI.
> > 
> > 
> > I've made some progress in identifying the problem but it's very annoying. We use `MAIN_DEPENDENCY` to create shader compile commands for Visual Studio, which forces VS to use the custom command instead of trying to "compile" the shader using the built-in VS function (which uses FXC). The problem is that CMake says you can only ever use one file as a `MAIN_DEPENDENCY`. Since we compile the same HLSL shader multiple times (one for each stage) this creates duplicate targets. I'm thinking of a solution.
> 
> The old/original version of the CMake rules got around this by assigning specific tags to each _target_. For instance if the high level shader target was `my_vs_shader` it would become `my_vs_shader_dxil` or `my_vs_shader_spv`. This of course meant you couldn't have two high level targets named the samething - but that's probably not too different from what's happening now.

It's unfortunately worse than that (in fact, the fact we're generating multiple targets is the problem for VS). I opened https://github.com/google/bigwheels/issues/97.

vettoreldaniele:
Ok, I've added a CI workaround for issue #97 until it's fixed, and also upgraded the CI workflow to use a newer version of Windows and to download and use the latest DX software rasterizer which has shader model 6.5 support.

dnovillo:
So, does this mean that we are now forced to build with `-j1` in Windows until #97 is fixed?

vettoreldaniele:
It was always the case. This PR just makes the problem more visible in the CI. Most users won't see this problem building one project at a time, it only occurs when building the entire solution at once (which is not a common case if you're working on one sample). As a temporary workaround, you can always trigger the build multiple times until it works (doing it twice almost always solves the problem for me).

dnovillo:
Oh, OK.  In that case, it's not a big problem.  We can put this in and fix
#97 separately.

dnovillo:
While #97 is outstanding, would you mind adding a small note in the README file about the issue?  I agree that it's not a huge deal, but it may trip folks building the whole thing.  Thanks.

LGTM with that addition.

dnovillo (approved):

vettoreldaniele:
> While #97 is outstanding, would you mind adding a small note in the README file about the issue? I agree that it's not a huge deal, but it may trip folks building the whole thing. Thanks.
> 
> LGTM with that addition.

Done.

PR #101, opened 2023-04-17
build: add support for standalone XR devices

Keenuts:
This commit is a rewrite of #90 to make it work with the Android build refactoring (#94).

Co-authored-by: Drew Thomas <async@google.com>


build.gradle
```diff
@@ -0,0 +1,171 @@
...
+            externalNativeBuild {
+                ndk {
+                    abiFilters 'arm64-v8a', 'armeabi-v7a'
+                }
+                cmake {
+                  arguments '-DPPX_BUILD_XR=TRUE',
+                            '-DPPX_XR_QUEST=TRUE',
+                            '-DBUILD_LOADER=FALSE',
+                            '-DBUILD_ALL_EXTENSIONS=FALSE',
+                            "-DOCULUS_OPENXR_SDK=${oculusOpenxrSdkLocation}"
```
at-ninja, line 111: The Quest OpenXR loader is currently omitted from this PR. Can you link the imported loader?

My change for this is [here](https://github.com/google/bigwheels/pull/90/files#diff-3bd332c63c934c2fa3d25499843e6da0cbc9fd176e2ac9a798a5a6d4c650ae20).

at-ninja (commented):
The Quest variant currently doesn't build. Once the OpenXR loader is re-linked, I'll try running it again.

local.properties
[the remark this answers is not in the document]
Keenuts: yes, was done in the parent commit

Keenuts:
All right, I missed some bits, but I think they are there now. Also rebased on main to this PR only have the new changes :)

src/ppx/xr_component.cpp
```diff
@@ -106,18 +137,32 @@ void XrComponent::InitializeBeforeGrfxDeviceInit(const XrComponentCreateInfo& cr
     instanceCreateInfo.enabledApiLayerCount       = (uint32_t)xrInstanceLayers.size();
     instanceCreateInfo.enabledApiLayerNames       = xrInstanceLayers.data();
     instanceCreateInfo.applicationInfo.apiVersion = XR_CURRENT_API_VERSION;
+
+#if defined(PPX_ANDROID)
```
apazylbe, line 141: Should this also check defined(PPX_OPENXR) if it's only relevant for OpenXR and the extension is only enabled if PPX_OPENXR is true above ?

src/ppx/xr_component.cpp
```diff
@@ -53,6 +73,17 @@ void XrComponent::InitializeBeforeGrfxDeviceInit(const XrComponentCreateInfo& cr
 
     std::vector<const char*> xrInstanceExtensions;
     xrInstanceExtensions.push_back(graphicsAPIExtension);
+#if defined(PPX_ANDROID) and defined(PPX_OPENXR)
```
at-ninja, line 76: This is needed for Quest builds. I think just checking for PPX_ANDROID should be enough.

src/ppx/xr_component.cpp
at-ninja: This is needed for all mobile OpenXR implementations, including Quest. PPX_ANDROID should be sufficient.

cmake/GraphicsSample.cmake
```diff
@@ -31,6 +31,17 @@ function(_add_sample_internal)
         add_executable("${TARGET_NAME}" ${ARG_SOURCES})
     endif()
 
+    if(PPX_BUILD_XR AND PPX_XR_QUEST)
+      # Quest devices do not use the standard OpenXR loader.
+      add_library(openxr_loader SHARED IMPORTED)
+      set_property(
+￼          TARGET
```
at-ninja, line 38: When I tried to build the Quest variants, the spacing here caused issues for me. Re-formatting this locally fixed the issue for me. Is there a rogue whitespace character?

at-ninja (commented):
The change to `Swapchain::IsHeadless` in `grfx_swapchain.h` is also needed, as [the code](https://github.com/google/bigwheels/blob/main/src/ppx/application.cpp#L791) currently assumes it is running in headless mode during initialization and crashes.

cmake/GraphicsSample.cmake
Keenuts: Yes it is.. File is back to ascii-only! Thanks

CMakeLists.txt
```diff
@@ -122,8 +122,14 @@ else()
         -Wno-deprecated-copy \
         -Wno-deprecated-enum-enum-conversion \
         -Wno-pointer-bool-conversion \
-        -Werror \
     ")
+
+    if (NOT PPX_BUILD_XR)
+        # OpenXR has way too many warnings to enabled -Werror with it.
```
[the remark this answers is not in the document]
Keenuts: thanks, fixed

src/ppx/xr_component.cpp
```diff
@@ -53,6 +73,17 @@ void XrComponent::InitializeBeforeGrfxDeviceInit(const XrComponentCreateInfo& cr
...
+    XrLoaderInitInfoAndroidKHR loaderInitInfoAndroid = {XR_TYPE_LOADER_INIT_INFO_ANDROID_KHR};
+    loaderInitInfoAndroid.applicationVM              = createInfo.androidContext->activity->vm;
+    loaderInitInfoAndroid.applicationContext         = createInfo.androidContext->activity->javaGameActivity;
+    PFN_xrInitializeLoaderKHR pfnInitializeLoaderKHR = nullptr;
+    CHECK_XR_CALL(xrGetInstanceProcAddr(XR_NULL_HANDLE, "xrInitializeLoaderKHR", (PFN_xrVoidFunction*)(&pfnInitializeLoaderKHR)));
+    PPX_ASSERT_MSG(pfnInitializeLoaderKHR != nullptr, "Cannot get xrInitializeLoaderKHR function pointer!");
+    CHECK_XR_CALL(pfnInitializeLoaderKHR((const XrLoaderInitInfoBaseHeaderKHR*)&loaderInitInfoAndroid));
+    xrInstanceExtensions.push_back(XR_KHR_ANDROID_CREATE_INSTANCE_EXTENSION_NAME);
+#endif // defined(PPX_ANDROID)
```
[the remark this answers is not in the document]
Keenuts: fixed thanks

src/ppx/xr_component.cpp
Keenuts: But makes sense not to enable it on Android, as android can be non-openxr. Fixed.

src/ppx/xr_component.cpp
Keenuts: Same as the other, might be best not to enable it on normal android devices no?

src/ppx/xr_component.cpp
apazylbe: Upon closer look, would this code even run in a non-openxr case? It looks like it should only run if XR is enabled?

at-ninja (approved):

apazylbe (approved):

PR #103, opened 2023-04-17
Resizable window, offscreen framebuffer, and resizable viewport

hysw:
* Enable window resizing, resize/recreate depend on platform
* Enable offscreen framebuffer, and copy resulting image to swapchain
* Enable resize of viewport (the actual resize knob will be in a separate CL)

Note, this is still WIP, I'd like to have a review before I modify all of the samples/benchmark

### DIFF
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -22,6 +22,7 @@
 #include "ppx/timer.h"
 #include "ppx/xr_component.h"
 #include "ppx/fs.h"
+#include "ppx/render_target.h"
 
 #include <deque>
 #include <filesystem>
@@ -238,6 +239,7 @@ struct ApplicationSettings
     bool        headless              = false;
     bool        enableImGui           = false;
     bool        allowThirdPartyAssets = false;
+    bool        useRenderTarget       = false;
 
     struct
     {
@@ -295,6 +297,16 @@ struct ApplicationSettings
             grfx::Format depthFormat = grfx::FORMAT_UNDEFINED;
             uint32_t     imageCount  = 2;
         } swapchain;
+
+        // Requires useRenderTarget
+        struct
+        {
+            bool         offscreen   = false;
+            grfx::Format colorFormat = grfx::FORMAT_UNDEFINED;
+            grfx::Format depthFormat = grfx::FORMAT_UNDEFINED;
+            uint32_t     width       = 4096;
+            uint32_t     height      = 4096;
+        } framebuffer;
     } grfx;
 };
 
@@ -343,6 +355,8 @@ protected:
     void DrawImGui(grfx::CommandBuffer* pCommandBuffer);
     void DrawDebugInfo(std::function<void(void)> drawAdditionalFn = []() {});
     void DrawProfilerGrfxApiFunctions();
+
+    virtual void DrawDebugUI();
 
 public:
     int  Run(int argc, char** argv);
@@ -394,6 +408,18 @@ public:
         return mSwapchain[index];
     }
 
+    RenderTarget* GetRenderTarget() const
+    {
+        if (mOffscreenRenderTarget) {
+            return mOffscreenRenderTarget.get();
+        }
+        if (mRenderTargetImGuiHook) {
+            return mRenderTargetImGuiHook.get();
+        }
+        return mSwapchainRenderTarget.get();
+    }
+    RenderTarget* GetSwapchainRenderTarget() const { return mSwapchainRenderTarget.get(); }
+
     float    GetElapsedSeconds() const;
     float    GetPrevFrameTime() const { return mPreviousFrameTime; }
     uint64_t GetFrameCount() const { return mFrameCount; }
@@ -434,6 +460,8 @@ private:
     Result InitializeGrfxDevice();
     Result InitializeGrfxSurface();
     Result InitializeGrfxSwapchain();
+    Result InitializeRenderTarget();
+    Result UpdateRenderTarget();
     Result InitializeImGui();
     void   ShutdownImGui();
     void   StopGrfx();
@@ -486,6 +514,12 @@ private:
     float             mAverageFrameTime  = 0;
     double            mFirstFrameTime    = 0;
     std::deque<float> mFrameTimesMs;
+
+    // Requires offscreenRenderTarget
+    // Application => [OffscreenBuffer] => [ImGuiHook] => Swapchain
+    std::unique_ptr<IndirectRenderTarget>    mOffscreenRenderTarget;
+    std::unique_ptr<RenderTargetPresentHook> mRenderTargetImGuiHook;
+    std::unique_ptr<SwapchainRenderTarget>   mSwapchainRenderTarget;
 
 #if defined(PPX_BUILD_XR)
     XrComponent mXrComponent;
--- a/include/ppx/config.h
+++ b/include/ppx/config.h
@@ -21,7 +21,7 @@
 #       define VC_EXTRALEAN
 #   endif
 #   if !defined(WIN32_LEAN_AND_MEAN)
-#       define WIN32_LEAN_AND_MEAN 
+#       define WIN32_LEAN_AND_MEAN
 #   endif
 #   if !defined(NOMINMAX)
 #       define NOMINMAX
@@ -112,6 +112,8 @@ enum Result
     ERROR_NO_GPUS_FOUND                = -16,
     ERROR_REQUIRED_FEATURE_UNAVAILABLE = -17,
     ERROR_BAD_DATA_SOURCE              = -18,
+    ERROR_OUT_OF_DATE                  = -19,
+    ERROR_SUBOPTIMAL                   = -20,
 
     ERROR_GLFW_INIT_FAILED          = -200,
     ERROR_GLFW_CREATE_WINDOW_FAILED = -201,
@@ -191,6 +193,8 @@ inline const char* ToString(ppx::Result value)
         case Result::ERROR_NO_GPUS_FOUND                              : return "ERROR_NO_GPUS_FOUND";
         case Result::ERROR_REQUIRED_FEATURE_UNAVAILABLE               : return "ERROR_REQUIRED_FEATURE_UNAVAILABLE";
         case Result::ERROR_BAD_DATA_SOURCE                            : return "ERROR_BAD_DATA_SOURCE";
+        case Result::ERROR_OUT_OF_DATE                                : return "ERROR_OUT_OF_DATE";
+        case Result::ERROR_SUBOPTIMAL                                 : return "ERROR_SUBOPTIMAL";
 
         case Result::ERROR_GLFW_INIT_FAILED                           : return "ERROR_GLFW_INIT_FAILED";
         case Result::ERROR_GLFW_CREATE_WINDOW_FAILED                  : return "ERROR_GLFW_CREATE_WINDOW_FAILED";
--- a/include/ppx/grfx/dx12/dx12_swapchain.h
+++ b/include/ppx/grfx/dx12/dx12_swapchain.h
@@ -70,10 +70,13 @@ protected:
         grfx::Fence*     pFence,
         uint32_t*        pImageIndex) override;
 
+    virtual Result ResizeInternal(uint32_t width, uint32_t height) override;
+
 private:
     DXGISwapChainPtr     mSwapchain;
     HANDLE               mFrameLatencyWaitableObject = nullptr;
     D3D12CommandQueuePtr mQueue;
+    UINT                 mSwapchainFlags;
 
     //
     // Store sync internval so we can control its behavior based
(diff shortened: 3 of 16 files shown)


chaoticbob (changes requested):
A few high level things:

- The current `RenderTarget` class should be called something else. `RenderTarget` has a very specific meaning in both D3D12 and Vulkan. It's synonymouse with *color attachment*. 
- Is the `RenderTarget` class really needed? Can't the existing `RenderPass` class be used to store the same information?
- I don't think applications should opt into using it - rather applications should opt out of presenting. Application should always render to the final/output render pass. 
- For ImGui, there are cases where the application would rather want to render to the actual swapchain image rather than a previous render pass. For example if there's any posts or color correction applied. 
- I think it would be more natural for applications to call something like `GetCurrentOutputPass()` for the offscreen buffer. This would replace the 3 calls to the swapchain acquire and fence reset. It should also block until an output pass becomes available. The call to the swapchain's present could just get replaced with a generalized `Present()` that actually presents or just NOOP on headless.

vettoreldaniele:
It's probably worth documenting the interaction with the headless option (which fakes the swapchain)

chaoticbob:
> > The current `RenderTarget` class should be called something else.
> 
> Naming is hard, and this is the reason I'm not modifying all sample yet. I might call it VirtualSwapchain or something...

One of the hardest problems in computer science I've heard. 

> > Can't the existing RenderPass class be used to store the same information?
> 
> It's keep tracking of interaction of RenderPass and swapchain, so RenderPass can probably store most of the information related to actually rendering. It also re-create swapchain and renderpass when things changes, so that part is not going to be in RenderPass.

Got it. FWIW the recreate will be triggered at the app level, for instance on a window resize event or on a device rotation, suspend/resume, sleep/wake, etc. So it should be modular enough to handle these cases. I didn't look at the create code too closely - does it recreate the surface as well?

> > For ImGui, there are cases where the application would rather want to render to the actual swapchain image rather than a previous render pass.
> 
> It's actually rendering directly to Swapchain image as of now.

Excellent.

> > I think it would be more natural for applications to call something like GetCurrentOutputPass() for the offscreen buffer.
> 
> That's probably the next step. I'm trying to keep it a drop in replacement for swapchain, so it won't take a week trying to change all the samples.

Got it. Just to make sure we're on the same page, you've accounted for required fencing between the different submits and access calls?

hysw:
> So it should be modular enough to handle these cases. I didn't look at the create code too closely - does it recreate the surface as well?

It's different for each platform :( 
* DX12 need to resize the swapchain
* Vulkan need to recreate the surface and swapchain and behavior depends on platform OUT_OF_DATE(linux) or SUBOPTIMAL(windows)

> Just to make sure we're on the same page, you've accounted for required fencing between the different submits and access calls?

Image layout transition is calling fence already, although I do need to verify the fences emitted by image layout transition is enough.

chaoticbob:
> > Just to make sure we're on the same page, you've accounted for required fencing between the different submits and access calls?
> 
> Image layout transition is calling fence already, although I do need to verify the fences emitted by image layout transition is enough.

I was meaning that work associated with the render targets needs to be completed before the copy to the swapchain. There should be a fence associated with each frame in flight to prevent copy hazards.

hysw:
> I was meaning that work associated with the render targets needs to be completed before the copy to the swapchain. There should be a fence associated with each frame in flight to prevent copy hazards.

I'm a bit confused, I don't think there need to be a (vulkan) fence between rendering and copy?
* AcquireImage/Present are synchronized by the actual driver swapchain
* The sequencing between submits are done with semaphore < Do you mean this as fence?
* Image/memory barrier are inserted by layout transition code

chaoticbob:
> > I was meaning that work associated with the render targets needs to be completed before the copy to the swapchain. There should be a fence associated with each frame in flight to prevent copy hazards.
> 
> I'm a bit confused, I don't think there need to be a (vulkan) fence between rendering and copy?
> 
> * AcquireImage/Present are synchronized by the actual driver swapchain
> * The sequencing between submits are done with semaphore < Do you mean this as fence?
> * Image/memory barrier are inserted by layout transition code

`AcquireImage` is partially synchronized. `vkAcquireNextImage` will block until an image becomes available. D3D12 we manually wait to make it behave similar to Vulkan: https://github.com/google/bigwheels/blob/main/src/ppx/grfx/dx12/dx12_swapchain.cpp#L312-L329

`Present` is not synchornized on either platform. On Vulkan it *can* take a semaphore to indicate that the image is ready for present. D3D12 also immitates the Vulkan behavior here.

The image/memory layout transitions are separate from work synchronization. The layout transition synchronize at the command level. Work synchronization does this at the submission level.

I think I'm losing track of what's trying to be accomplished with this PR? Is the intent here to just decouple the color attachments from the swapchain so they can be a different resolution?

hysw:
> I think I'm losing track of what's trying to be accomplished with this PR? Is the intent here to just decouple the color attachments from the swapchain so they can be a different resolution and make the swapchain recreatable?

Yes.

chaoticbob:
> > I think I'm losing track of what's trying to be accomplished with this PR? Is the intent here to just decouple the color attachments from the swapchain so they can be a different resolution and make the swapchain recreatable?
> 
> Yes.

Thanks. I'll try to keep my head straight about it.

I think this should should be two separate PRs - one for the decoupling the color attachments and one for swapchain recreation. Just to keep things a bit easier to focus on.

Here's my suggestions for decoupling the color attachments:

Add render passes, sync objects, and functions to the Application class. The logic for this needs to be similar what Swapchain was doing to minimize the changes to samples.

Remove the support for depth from swapchain. 

Instead of introducing the `renderTarget` into the ApplicationSettings introduce a `render`.  Calling anything render target is will be super confusing to talk about if it isn't an actual render target.  This would be my suggestion:
```C++
struct ApplicationSettings
{
    std::string appName               = "";
    bool        headless              = false;
    bool        enableImGui           = false;
    bool        allowThirdPartyAssets = false;

    struct
    {
        bool enable             = false;
        bool enableDebugCapture = false;

        // Whether to create depth swapchains in addition to color swapchains,
        // and submit the depth info to the runtime as an additional layer.
        bool enableDepthSwapchain = false;

        // OpenXR uses a right-handed system.
        // The `pos` here is the center position in view space.
        // A detailed description can be found here:
        // https://registry.khronos.org/OpenXR/specs/1.0/man/html/XrCompositionLayerQuad.html
        struct
        {
            float3 pos;
            float2 size;
        } ui;
    } xr;

    struct
    {
        uint32_t    width     = 0;
        uint32_t    height    = 0;
        std::string title     = "";
        bool        resizable = false;
    } window;
 
    // NEW
    struct
    {
        uint32_t width  = 0; // Use window.width if 0
        uint32_t height = 0; // Use window.height if 0
        grfx::Format colorFormat = grfx::FORMAT_R8G8B8A8_UNORM;     
        grfx::Format depthFormat = grfx::FORMAT_UNDEFINED; 
    } render;

    struct
    {
        grfx::Api api               = grfx::API_UNDEFINED;
        bool      enableDebug       = false;
        uint32_t  numFramesInFlight = 1;
        uint32_t  pacedFrameRate    = 60;

        struct
        {
            uint32_t gpuIndex           = 0;
            uint32_t graphicsQueueCount = 1;
            uint32_t computeQueueCount  = 0;
            uint32_t transferQueueCount = 0;
        } device;

        struct
        {
            // NVIDIA only supports B8G8R8A8, ANDROID only supports R8G8B8A8, and
            // AMD supports both. So the default has to special-case either NVIDIA
            // or ANDROID :(
#if defined(PPX_ANDROID)
            grfx::Format colorFormat = grfx::FORMAT_R8G8B8A8_UNORM;
#else
            grfx::Format colorFormat = grfx::FORMAT_B8G8R8A8_UNORM;
#endif
            // REMOVE: grfx::Format depthFormat = grfx::FORMAT_UNDEFINED;
            uint32_t     imageCount  = 2;
        } swapchain;
    } grfx;
};
```

hysw:
> I think this should should be two separate PRs - one for the decoupling the color attachments and one for swapchain recreation. Just to keep things a bit easier to focus on.

Originally I set it as Draft PR, and it seems to have problem with adding reviewer or something. I'm going to split this up probably to more than 2 PRs.

> Add render passes, sync objects, and functions to the Application class. The logic for this needs to be similar what Swapchain was doing to minimize the changes to samples.

I'm making new C++ classes since it's just easier and cleaner than using a giant cluster of if statement.

> Remove the support for depth from swapchain.

There's OpenXR swapchain that actually need depth.

> Instead of introducing the renderTarget into the ApplicationSettings introduce a render. Calling anything render target is will be super confusing to talk about if it isn't an actual render target. This would be my suggestion:

I'm definitely renaming it.

Given I'm essentially faking a swapchain, I'm just going to call it Swapchain (`ppx::Swapchain`). Also in the application setting, it's currently called `settings.grfx.framebuffer`, I guess `render` should also work.

hysw:
> Reuse the RenderPass object if you can.

The global change required is mostly switching out the swapchain implementation, which is
https://github.com/hysw/bigwheels/commit/ac466d9696b459ee9e6a4fec9de92e88f0c20961

hysw:
> This should simplify the swapchain not add more complexity to it.

It is adding a extra step, and also handling more cases. The device swapchain does get a bit simpler (I've yet pull the depth image creation off it).

Here's a rough split of the PR
* Change swapchain to a interface instead of device object (this is so he implementation can differ from device swapchain) https://github.com/hysw/bigwheels/commit/a0dd3383e3dc520df94a0cb4a9a6cb21b7550bda
* Let application (through swapchain wrapper) manage renderpass instead of device swapchain https://github.com/hysw/bigwheels/commit/e9e7b865444a4dce61bc56d82743ddfe7688a08d
* The parts required for window resizing https://github.com/hysw/bigwheels/commit/bd2a023fc50aecd4c1ad561c7c344b3d304a6e74
* The render to a off-screen frame buffer https://github.com/hysw/bigwheels/commit/03dd8fee5dd64d7adf223d3f2bad8060846232f7

hysw:
Spent some more time cleaning up the change, let me know if there's anything more I should change before creating a PR for each of them

* Swapchain stub, so stuff can be added between application and the real swapchain  https://github.com/hysw/bigwheels/commit/a2214e22125bd4a2d8826b4925818a86ff2ae7bf
* Move ImGui render out of application rendering logic (so when off-screen buffer is a thing, it will still render to swapchain directly) https://github.com/hysw/bigwheels/commit/c6d7bc018bbed09f2437e350ccbffdf414af0044
* Move renderpass management logic out of 
grfx::Swapchain, so we can keep track of separate renderpass for swapchain image and for off screen buffer https://github.com/hysw/bigwheels/commit/c235d1045e5746494d9c8f837c53049a43ff8327
* Window resize https://github.com/hysw/bigwheels/commit/526223747a7aeceb389038c26395957fc4e469aa
* An offscreen frame buffer https://github.com/hysw/bigwheels/commit/2ea2c32e4a36a5b28445b142242f7ee77d8799e6

When `grfx.framebuffer.offscreen` is set to `true`, the application render to an off-screen buffer that gets copied to the swapchain
```
Application > [offscreen buffer > copy image] > [ImGui] > wrapper (swapchain errors) > device swapchain
```

There's a few more things that it not in this PR
* Now that ImGui is out of application rendering block, we should be able to time just the rendering part without ImGui
* Remove depth buffer logic from `grfx::swapchain`
* Move the headless logic out of `grfx::Swapchain` (it's off-screen buffer + sync)
* Update some application to actually read up to date swapchain size

chaoticbob:
These changes look fairly intrusive and breaks some of the existing design paradigms. The ObjectPtr pattern is used to force the initailization to null if something isn't assigned.

Overall, these changes feel really awkward to use. There seems to be an overconformance of BigWheels to XR. This should not be the case. The conformance should be the other way around.

A few of the comments regarding the swapchain don't make sense. Width/height are an integral part of the swapchain. I don't quite understand why they would be ambiguous.

hysw:
> These changes look fairly intrusive and breaks some of the existing design paradigms. The ObjectPtr pattern is used to force the initailization to null if something isn't assigned.

Those should be reference instead of pointers, since they can't be null anyway.

> Overall, these changes feel really awkward to use. There seems to be an overconformance of BigWheels to XR. This should not be the case. The conformance should be the other way around.

I agree it's awkward. It has nothing to do with XR. The issue here is each application has direct control over the swapchain. I would like to make sample to use a renderpass instead of going through all the swapchain steps, but that requires a large rewrite for each app. Also at least one of them (fishtornado) need to manage their own command buffer dependencies, so it also needs all the semaphores, which basically means it's using a swapchain.

> A few of the comments regarding the swapchain don't make sense. Width/height are an integral part of the swapchain. I don't quite understand why they would be ambiguous.

The swapchain width/height un-ambiguously refers to the swapchain's image width/height, however in the samples, it is used as synonym of `viewport.{width,height}` and `scissor.{width,height}`, and that only makes sense if our rendering resolution is the same as swapchain resolution.

Depend on the sample viewport/scissor in the samples are obtained with
- `swapchain->Get{Width,Height}`
- `renderpass->Get{Scissor,Viewport}`
-  `app->GetWindow{Width,Height}`
-  `app->Get{Viewport,Scissor}`
and either at the startup or during render call


If I have two month of time fixing this, it would be a lot more clean than stubbing out swapchain implementation.

hysw:
(also mark as draft since this is not intended to be submitted)

hysw:
Part of this is in other PRs, and the rest does not need to be here.

PR #107, opened 2023-04-18
Fixes and update for 15_basic_material

chaoticbob:
- Fixed lighting for Blinn and Phong shaders
- Replaced PBR shader with a homegrown version that is very well behaved
- Added new models and textures to test various material responses using the new PBR shader
- Added a Translate() to TriMesh options
- Updated sample to use all the new things
- Changed default normal maps for materials to true


projects/15_basic_material/main.cpp
```diff
@@ -106,7 +106,10 @@ class ProjApp
     grfx::MeshPtr              mSphere;
     grfx::MeshPtr              mCube;
     grfx::MeshPtr              mMonkey;
-    grfx::MeshPtr              mAltimeterModel;
+    grfx::MeshPtr              mMeasuringTape;
```
Keenuts, line 109: Are those needed, as the pointer is stored in the vector anyway? They are only read during loading, but then stored into the vector.

projects/15_basic_material/main.cpp
```diff
@@ -132,7 +135,11 @@ class ProjApp
     grfx::SamplerPtr                  mSampler;
     MaterialResources                 mWoodMaterial;
     MaterialResources                 mTilesMaterial;
-    MaterialResources                 mAltimeterMaterial;
+    MaterialResources                 mStoneWallMaterial;
+    MaterialResources                 mMeasuringTapeMaterial;
```
Keenuts, line 139: Same for the materials.

projects/15_basic_material/main.cpp
```diff
@@ -202,7 +209,10 @@ class ProjApp
         "Sphere",
         "Cube",
         "Monkey",
-        "Altimeter",
+        "Measuring Tape",
```
Keenuts, line 212: Goodbye altimeter 😢

projects/15_basic_material/main.cpp
```diff
@@ -377,37 +393,92 @@ void ProjApp::SetupMaterials()
...
     PPX_CHECKED_CALL(GetDevice()->CreateDescriptorSetLayout(&createInfo, &mMaterialResourcesLayout));
 
+    // Green metal rust
+    {
+        SetupMaterialResources(
+            "poly_haven/textures/green_metal_rust/diffuse.png",
+            "poly_haven/textures/green_metal_rust/roughness.png",
+            "poly_haven/textures/green_metal_rust/metalness.png",
+            "poly_haven/textures/green_metal_rust/normal.png",
+            mWoodMaterial);
```
Keenuts, line 403: mWoodMaterial is used below also. Fine since those attributes seem useless, but still.

projects/15_basic_material/main.cpp
```diff
@@ -377,37 +393,92 @@ void ProjApp::SetupMaterials()
...
     };
 
     // Tiles
     {
         SetupMaterialResources(
-            "materials/textures/tiles/albedo.png",
-            "materials/textures/tiles/roughness.png",
-            "materials/textures/tiles/metalness.png",
-            "materials/textures/tiles/normal.png",
+            "poly_haven/textures/square_floor/diffuse.png",
```
Keenuts, line 421: square_tiles ? or square_floor_tiles?

Keenuts (approved):
some details, otherwise good. Looks better than the last version!

projects/15_basic_material/main.cpp
chaoticbob: These aren't needed.  I've removed in the latest push and added some logic to pull the vertex binding information.

projects/15_basic_material/main.cpp
chaoticbob: These aren't strictly needed but MaterialResources hangs on to some TexturePtr objects which would get auto destroy when device exists anyway...but if we don't keep them, it looks like we're purposely dangling pointers. I kept them for now but maybe a future PR can help fix that.

projects/15_basic_material/main.cpp
chaoticbob: Changed it to square_floor_tiles to make it very obvious what it is.

chaoticbob:
> some details, otherwise good. Looks better than the last version!

Thanks. Addressed the details.

assets/materials/shaders/PBR.hlsl
```diff
@@ -32,42 +28,39 @@ Texture2D    IBLTex         : register(IBL_MAP_TEXTURE_REGISTER,    MATERIAL_RES
...
 {
-    float alpha   = roughness * roughness;
-    float alphaSq = alpha * alpha;
-    float denom   = (cosLh * cosLh) * (alphaSq - 1.0) + 1.0;
-    return alphaSq / (PI * denom * denom);
+    float NoH    = saturate(dot(N, H));
+    float NoH2   = NoH * NoH;
+    float alpha2 = max(roughness * roughness, EPSILON);
+    float A      = NoH2 * (alpha2 - 1) + 1;
+	return alpha2 / (PI * A * A);
```
apazylbe, line 37: Is this supposed to be indented with extra space? Here and in the next function below (Geometry_SchlickBeckman)

assets/materials/shaders/PBR.hlsl
```diff
@@ -116,108 +112,62 @@ float4 psmain(VSOutput input) : SV_TARGET
...
-        // clang-format on
-        float3 diffuseBRDF = kD * albedo;
-
-        // Calculate specular BRDF
-        float3 specularBRDF = (F * D * G) / max(0.00001, 4.0 * cosLi * cosLo);
-    
-        // Accumulate light's total contribution
-        directLighting += (diffuseBRDF + specularBRDF) * Lrad * cosLi;
+    float3 directLighting = 0;
+	for (uint i = 0; i < Scene.lightCount; ++i) {
```
apazylbe, line 130: Do we need to fix indentation here?

assets/materials/shaders/PBR.hlsl
chaoticbob: The HLSL code doesn't get fed through an automatic formatter. clang-format doesn't play nice with HLSL yet. So it's just subjected to whatever formatting it came from. This code was copied over from some of my other projects.

chaoticbob: Also, we don't currently have an HLSL style guide :(

assets/materials/shaders/PBR.hlsl
apazylbe: Line 130 here, if it should have one fewer \t, same as line 129

assets/materials/shaders/PBR.hlsl
vettoreldaniele: Filed https://github.com/google/bigwheels/issues/108

assets/materials/shaders/PBR.hlsl
chaoticbob: Thanks. Fixed.

PR #109, opened 2023-04-18
Distinguish read-only and read-write structured buffers

vettoreldaniele:
DX12 needs the distinction between SRV and UAV when lowering the descriptor type.


chaoticbob (approved):
LGTM

PR #110, opened 2023-04-18
Added a new buffer-based OIT algorithm

mdagois:
Added a buffer-based OIT algorithm that builds per-pixel linked list of fragments
* It is available under the "Buffer" dropdown
* It has two options

Code refactoring
* Renamed the previous buffer-based algorithm to "buckets" as it uses per-pixel fixed-size buckets to store fragments
* Moved functions and renamed variables in the shader code

### DIFF
--- a/projects/oit_demo/Buffer.cpp
+++ b/projects/oit_demo/Buffer.cpp
@@ -222,6 +222,8 @@ void OITDemoApp::SetupBufferBuckets()
 
 void OITDemoApp::SetupBufferLinkedLists()
 {
+    mBuffer.lists.linkedListHeadTextureNeedClear = true;
+
     // Linked list head texture
     {
         grfx::TextureCreateInfo createInfo         = {};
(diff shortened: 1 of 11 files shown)


projects/oit_demo/Buffer.cpp
```diff
@@ -52,29 +52,30 @@ void OITDemoApp::SetupBuffer()
...
-        createInfo.height                    = mBuffer.countTexture->GetHeight();
-        createInfo.renderTargetCount         = 0;
-        createInfo.pDepthStencilImage        = nullptr;
-        PPX_CHECKED_CALL(GetDevice()->CreateDrawPass(&createInfo, &mBuffer.gatherPass));
+        grfx::DrawPassCreateInfo2 createInfo  = {};
+        createInfo.width                      = mBuffer.buckets.countTexture->GetWidth();
+        createInfo.height                     = mBuffer.buckets.countTexture->GetHeight();
+        createInfo.renderTargetCount          = 0;
+        createInfo.pDepthStencilImage         = nullptr;
+        createInfo.renderTargetClearValues[0] = {0, 0, 0, 0};
```
vettoreldaniele, line 77: No clear values needed

vettoreldaniele (approved):
LGTM

projects/oit_demo/Buffer.cpp
mdagois: Fixed!

PR #111, opened 2023-04-18
Adds custom row stride version of Bitmap::Create

chaoticbob:
This change adds a variant of the Bitmap::Create functions for external storage to take a custom row stride for cases where the row stride is not width * pixel stride.

### DIFF
--- a/include/ppx/bitmap.h
+++ b/include/ppx/bitmap.h
@@ -66,11 +66,17 @@ public:
 
     Bitmap& operator=(const Bitmap& rhs);
 
+    //! Creates a bitmap with internal storage.
     static Result Create(uint32_t width, uint32_t height, Bitmap::Format format, Bitmap* pBitmap);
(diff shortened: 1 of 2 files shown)


include/ppx/bitmap.h
```diff
@@ -182,7 +184,7 @@ class Bitmap
 
 private:
     void   InternalCtor();
-    Result InternalInitialize(uint32_t width, uint32_t height, Bitmap::Format format, char* pExternalStorage);
+    Result InternalInitialize(uint32_t width, uint32_t height, Bitmap::Format format, uint32_t rowStride, char* pExternalStorage);
```
vettoreldaniele, line 187: Can you add a comment that if `rowStride` is zero, then the format's default row stride is used?

chaoticbob: Done. Added comments  for that block of functions to keep things uniform.

vettoreldaniele (approved):

PR #114, opened 2023-04-19
Add command line args to Android builds.

at-ninja:
This uses [Intent extras](https://developer.android.com/reference/android/content/Intent#getExtras()) to allow the Android application to be started over ADB with additional arguments.

For example:
```sh
adb shell am start -n com.google.bigwheels/.MainActivity -e arguments "--help"
```

### DIFF
--- a/src/android/MainActivity.java
+++ b/src/android/MainActivity.java
@@ -1,8 +1,27 @@
 package com.google.bigwheels;
 
+import android.content.Intent;
+import android.os.Bundle;
 import android.view.View;
 import com.google.androidgamesdk.GameActivity;
 
 public class MainActivity extends GameActivity {
   static { System.loadLibrary(BuildConfig.sample_library_name); }
+
+  private String[] getArgsFromIntent(Intent intent) {
+    String argString = intent.getStringExtra("arguments");
+    if (argString != null) {
+      return argString.split("\\s+");
+    }
+    return new String[0];
+  }
+
+  public String[] constructCmdLineArgs() {
+    String[] intentArgs = getArgsFromIntent(getIntent());
+    String[] args = new String[intentArgs.length + 1];
+    // The first argument is the application name, and is ignored by the parser.
+    args[0] = getTitle().toString();
+    System.arraycopy(intentArgs, 0, args, 1, intentArgs.length);
+    return args;
+  }
 }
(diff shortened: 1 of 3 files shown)


src/android/main.cpp
```diff
@@ -54,8 +58,31 @@ extern "C"
             }
 
             if (IsVulkanReady()) {
-                VulkanDrawFrame();
+                std::vector<char*> args;
+                args.reserve(cmdArgs.size());
+                for (std::string& arg : cmdArgs) {
+                    args.push_back(const_cast<char*>(arg.c_str()));
```
Keenuts, line 64: This relies on a few assumptions:
 - nativeSetCommandLineArgs is called once (otherwise pointers passed to main are now invalid)
 - this gets called after nativeSetCommandLineArgs.
 
 Are we sure that's the case? What happens when the activity gets destroyed and recreated, or just suspended (screen lock, rotation, etc).
 
 Is there a way to do it the other way around, have the native code calls an activity method returning the arguments?
 This way you'd get a copy of the arguments to pass down, and avoid this global vector which is modified and read in different locations.

src/android/MainActivity.java
```diff
@@ -1,8 +1,31 @@
...
 public class MainActivity extends GameActivity {
   static { System.loadLibrary(BuildConfig.sample_library_name); }
+
+  @Override
+  protected void onCreate(Bundle savedState) {
+    super.onCreate(savedState);
+
+    String[] intentArgs = getArgsFromIntent(getIntent());
+    String[] args = new String[intentArgs.length + 1];
+    args[0] = ""; // The first argument is ignored by the argument parser.
```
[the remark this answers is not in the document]
at-ninja: Done. Thanks for the suggestion.

src/android/main.cpp
at-ninja: Done. This adds a lot of JNI boilerplate, but the flow makes more sense.

apazylbe:
I am new to the project, but [Contributing page](https://github.com/google/bigwheels/blob/main/CONTRIBUTING.md#for-developers-contributing-a-patch) recommends adding new commits, instead of amending, so that it's easier to see what has changed.

apazylbe (approved):

Keenuts:
Thanks for fixing! Sorry you had to go through JNI calls, always horrible to witness 😅

PR #117, opened 2023-04-19
Indirect lighting for PBR shader in basic_material

chaoticbob:
This change adds indirect lighting for the PBR shader in the basic_material sample. For performance on mobile, the BRDF integration is done using an approximation algorithm instead a LUT.

This change also contains irradiance and environment maps for the indirect lighting. These maps were baked using a tool. The baking process acounts for hemispherical sampling for the irradiance map and roughness on the environment map. This can't be done at run time, it would take forever to generate. The tool will be checked in in a different PR.

- Added indirect lighting calculation to PBR shader.
- Changed Mipmap::LOadFile to load vertically tailed mip maps.
- Added method to creat texture from Mipmap.
- Added EnvDraw.hlsl for environment drawing with color correction and gamma adjustment.


assets/materials/shaders/CMakeLists.txt
```diff
@@ -37,3 +37,11 @@ generate_rules_for_shader("shader_mtl_pbr"
     SOURCE "${PPX_DIR}/assets/materials/shaders/PBR.hlsl"
     INCLUDES ${INCLUDE_FILES}
     STAGES "ps")
+
+generate_rules_for_shader("shader_mtl_env_draw_vertex_shader"
```
vettoreldaniele, line 41: This can be a single rule with `STAGES "vs" "ps"`.

projects/15_basic_material/main.cpp
```diff
@@ -1038,6 +1250,52 @@ void ProjApp::Render()
         GetGraphicsQueue()->CopyBufferToBuffer(&copyInfo, mCpuModelConstants, mGpuModelConstants, grfx::RESOURCE_STATE_CONSTANT_BUFFER, grfx::RESOURCE_STATE_CONSTANT_BUFFER);
     }
 
+    // Update env draw constants
```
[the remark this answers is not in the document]
chaoticbob: In the case of this simple example, it doesn't really matter. Elsewhere on desktop it depends on the API and which memory type and if BAR is exposed or not. If the upload buffer ends up in system memory that's uncached, it will get read across the PCIe bus which would end up detrimental. Uploading to the the GPU guarantees that it will always be read from device memory.

include/ppx/graphics_util.h
```diff
@@ -216,6 +233,13 @@ Result CreateTexture1x1(
     return CreateTextureFromBitmap(pQueue, &bitmap, ppTexture, options);
 }
 
+// Creates an irradiance and an environment texture from an *.ibl file
+Result CreateIBLTexturesFromFile(
+    grfx::Queue*                 pQueue,
+    const std::filesystem::path& path,
+    grfx::Texture**              ppIrradianceTexture,
+    grfx::Texture**              ppEnvironemntTexture);
```
apazylbe, line 241: typo in Environment

src/ppx/graphics_util.cpp
```diff
@@ -966,6 +966,85 @@ Result CreateTextureFromBitmap(
     return ppx::SUCCESS;
 }
 
+Result CreateTextureFromMipmap(
+    grfx::Queue*          pQueue,
+    const Mipmap*         pMipmap,
+    grfx::Texture**       ppTexture,
+    const TextureOptions& options)
+{
+    PPX_ASSERT_NULL_ARG(pQueue);
```
apazylbe, line 975: Not for this CL, but in general, I wonder if this should be PPX_ASSERT_NOT_NULL_ARG

projects/15_basic_material/main.cpp
```diff
@@ -642,6 +765,51 @@ void ProjApp::Setup()
...
+        PPX_CHECKED_CALL(GetDevice()->CreateBuffer(&bufferCreateInfo, &mGpuEnvDrawConstants));
+
+        grfx::WriteDescriptor write = {};
+        write.binding               = 0;
+        write.arrayIndex            = 0;
+        write.type                  = grfx::DESCRIPTOR_TYPE_UNIFORM_BUFFER;
+        write.bufferOffset          = 0;
+        write.bufferRange           = PPX_WHOLE_SIZE;
+        write.pBuffer               = mGpuEnvDrawConstants;
+        PPX_CHECKED_CALL(mEnvDrawSet->UpdateDescriptors(1, &write));
```
apazylbe, line 797: I am curious, is there a reason we can't/shouldn't use one UpdateDescriptors call for all 3 writes?

assets/materials/shaders/CMakeLists.txt
chaoticbob: Fixed. Thanks didn't know this was a thing!

src/ppx/mipmap.cpp
```diff
@@ -209,15 +216,18 @@ Result Mipmap::LoadFile(const std::filesystem::path& path, Mipmap* pMipmap, uint
...
+    uint32_t totalHeight = 0;
+    for (uint32_t i = 0; i < levelCount; ++i) {
+        totalHeight += (baseHeight >> i);
+    }
 
-    *pMipmap = Mipmap(width, height, format, levelCount);
-    if (!pMipmap->IsOk()) {
-        // Something has gone really wrong if this happens
-        return ppx::ERROR_FAILED;
+    // Verify that dimemnsions make sense
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

assets/materials/shaders/EnvDraw.hlsl
```diff
@@ -0,0 +1,50 @@
+// Copyright 2022 Google LLC
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

src/ppx/graphics_util.cpp
chaoticbob: I don't have a strong opinion on this. The original logic was is assert on the condition in the macro name, .e.g. assert on null arg. This maybe different than how others think about the assert if expression inside assert() is false.

apazylbe: Thanks for the explanation! I usually read this as "assert (that) condition/assumption", and I think that's how it's usually defined in other places. Just by looking at the function name I had to look at the definition to see what it was meant to do. It'd be different if it was named THROW_IF_COND, FAIL_IF_COND. Anyway, it's not related to this PR, so not really important.

apazylbe (approved):

Issue #77, opened 2023-04-07
Enable IMGUI for Android


Keenuts:
DearImGUI seems to have support for Android. The initial porting to Android disabled it, but we should find a way to enable it back.

hysw:
See #119 #136 and #132

PR #119, opened 2023-04-20
Enable ImGui on Android.

hysw:
ImGui renders, but the aspect ratio is wrong.
We should get the Android Window size instead of hard code 720p.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -127,6 +127,13 @@ if (PPX_VULKAN)
     )
 endif()
 
+if (PPX_ANDROID)
+    list(APPEND IMGUI_SOURCE_FILES
+        ${IMGUI_DIR}/backends/imgui_impl_android.h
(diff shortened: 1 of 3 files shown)


Keenuts (approved):

apazylbe:
Is this enough to close https://github.com/google/bigwheels/issues/77? If that's the case, could you add `Fixes https://github.com/google/bigwheels/issues/77` in the description (https://github.com/google/bigwheels/blob/main/CONTRIBUTING.md#for-developers-contributing-a-patch)

apazylbe (approved):

hysw:
> Is this enough to close https://github.com/google/bigwheels/issues/77?

Not yet, this only enable the rendering of ImGui, there's at least a few more things to fix:

* The surface size is wrong, so ImGui has a weird aspect ratio
* Still need to add input support
* The bigwheels Android event loop is broken (it's not processing Android input at all)

PR #120, opened 2023-04-20
Add basic passthrough support to XrComponent.

at-ninja:
This will allow future benchmarks or projects to utilize the passthrough features on HMDs.

### DIFF
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -490,9 +490,10 @@ private:
 
 #if defined(PPX_BUILD_XR)
     XrComponent mXrComponent;
-    uint32_t    mDebugCaptureSwapchainIndex = 0;
-    uint32_t    mUISwapchainIndex           = 0;
-    uint32_t    mStereoscopicSwapchainIndex = 0;
+    uint32_t    mDebugCaptureSwapchainIndex  = 0;
+    uint32_t    mUISwapchainIndex            = 0;
+    uint32_t    mStereoscopicSwapchainIndex  = 0;
+    int         mTogglePassthroughFrameCount = 0;
 #endif
 };
 
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -490,10 +490,9 @@ private:
 
 #if defined(PPX_BUILD_XR)
     XrComponent mXrComponent;
-    uint32_t    mDebugCaptureSwapchainIndex  = 0;
-    uint32_t    mUISwapchainIndex            = 0;
-    uint32_t    mStereoscopicSwapchainIndex  = 0;
-    int         mTogglePassthroughFrameCount = 0;
+    uint32_t    mDebugCaptureSwapchainIndex = 0;
+    uint32_t    mUISwapchainIndex           = 0;
+    uint32_t    mStereoscopicSwapchainIndex = 0;
 #endif
 };
 
--- a/src/ppx/application.cpp
+++ b/src/ppx/application.cpp
@@ -1318,6 +1318,8 @@ int Application::Run(int argc, char** argv)
     }
 
 #if defined(PPX_BUILD_XR)
+    mTogglePassthroughFrameCount = mCommandLineParser.GetOptions().GetExtraOptionValueOrDefault<int>("xr-toggle-passthrough-frame-count", -1);
+
     if (mSettings.xr.enable) {
         XrComponentCreateInfo createInfo = {};
         createInfo.api                   = mSettings.grfx.api;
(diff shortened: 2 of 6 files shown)


include/ppx/application.h
```diff
@@ -490,9 +490,10 @@ class Application
 
 #if defined(PPX_BUILD_XR)
     XrComponent mXrComponent;
-    uint32_t    mDebugCaptureSwapchainIndex = 0;
-    uint32_t    mUISwapchainIndex           = 0;
-    uint32_t    mStereoscopicSwapchainIndex = 0;
+    uint32_t    mDebugCaptureSwapchainIndex  = 0;
+    uint32_t    mUISwapchainIndex            = 0;
+    uint32_t    mStereoscopicSwapchainIndex  = 0;
+    int         mTogglePassthroughFrameCount = 0;
```
Keenuts, line 496: Why is this a signed int, since you check for `mTogglePassthroughFrameCount > 0`? Looks like is could be an uint?

src/ppx/application.cpp
```diff
@@ -1438,6 +1440,10 @@ int Application::Run(int argc, char** argv)
                 break;
             }
 
+            if (mTogglePassthroughFrameCount > 0 && mFrameCount % mTogglePassthroughFrameCount == 0) {
+                mXrComponent.TogglePassthrough();
```
Keenuts, line 1444: What's the goal of toggling the passthrough every N frames?

Keenuts (commented):
Don't have XR device on hand to test right now nor know how it works, but looking at the code, seems logical. Just one high-level question.

include/ppx/application.h
at-ninja: I removed the field.

src/ppx/application.cpp
at-ninja: It was useful for testing the feature. Removed for now.

src/ppx/xr_component.cpp
```diff
@@ -253,6 +300,74 @@ void XrComponent::Destroy()
...
+        PPX_ASSERT_MSG(mPassthroughLayer != XR_NULL_HANDLE, "XrPassthroughLayerFB creation failed!");
+    }
+
+    PFN_xrPassthroughLayerResumeFB pfnXrPassthroughLayerResumeFB = nullptr;
+    CHECK_XR_CALL(xrGetInstanceProcAddr(mInstance, "xrPassthroughLayerResumeFB", (PFN_xrVoidFunction*)(&pfnXrPassthroughLayerResumeFB)));
+    PPX_ASSERT_MSG(pfnXrPassthroughLayerResumeFB != nullptr, "Cannot get xrPassthroughLayerResumeFB function pointer!");
+    CHECK_XR_CALL(pfnXrPassthroughLayerResumeFB(mPassthroughLayer));
+#endif // defined(PPX_XR_QUEST)
+
+    mPassthroughEnabled = true;
```
Keenuts, line 335: shouldn't this be set to true only in the PPX_XR_QUEST case? (shouldn't reach that, but to make code coherent) (same in End function)

Keenuts (approved):
Just a small nit, otherwise fine with this!

src/ppx/xr_component.cpp
```diff
@@ -434,11 +549,26 @@ void XrComponent::EndFrame(const std::vector<grfx::SwapchainPtr>& swapchains, ui
...
+        if (mPassthroughSupported && mPassthroughEnabled) {
+#if defined(PPX_XR_QUEST)
+            if (mPassthroughLayer != XR_NULL_HANDLE) {
+                XrCompositionLayerFlags         compositionLayerPassthroughFlags = {};
+                XrCompositionLayerPassthroughFB compositionLayerPassthrough      = {XR_TYPE_COMPOSITION_LAYER_PASSTHROUGH_FB};
+                compositionLayerPassthrough.next                                 = nullptr;
+                compositionLayerPassthrough.flags                                = compositionLayerPassthroughFlags;
+                compositionLayerPassthrough.space                                = XR_NULL_HANDLE;
+                compositionLayerPassthrough.layerHandle                          = mPassthroughLayer;
+                layers.push_back(reinterpret_cast<XrCompositionLayerBaseHeader*>(&compositionLayerPassthrough));
```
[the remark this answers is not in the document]
Keenuts: Completely missed that. Yes that's right! It was probably working because all the local variables are allocated with one sub at the function start, but definitely a nasty bug.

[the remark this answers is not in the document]
at-ninja: Thanks for the catch! Fixed.

apazylbe (approved):
Thanks for updating! lgtm

apazylbe:
Perhaps we should update the description to remove the use of the flag?

at-ninja:
The flag has been removed from the initial comment.

PR #124, opened 2023-04-24
Metrics framework API

mdagois:
* Implemented most of the metrics API
* Added metrics API tests

### DIFF
--- a/include/ppx/metrics.h
+++ b/include/ppx/metrics.h
@@ -149,8 +149,8 @@ class Run final
     friend class Manager;
 
 public:
-    Result       AddMetricGauge(MetricMetadata metadata, MetricGauge** outMetric);
-    Result         AddMetricCounter(MetricMetadata metadata, MetricCounter** outMetric);
+    template <typename T>
+    Result AddMetric(MetricMetadata metadata, T** outMetric);
 
 private:
     Run(const char* name);
@@ -159,13 +159,32 @@ private:
 
     Metric* GetMetric(const char* name) const;
 
-    template <typename T>
-    Result AddMetric(MetricMetadata metadata, T** outMetric);
-
 private:
     std::string                              mName;
     std::unordered_map<std::string, Metric*> mMetrics;
 };
+
+template <typename T>
+Result Run::AddMetric(MetricMetadata metadata, T** outMetric)
+{
+    PPX_ASSERT_MSG(outMetric != nullptr, "The metric pointer must not be null");
+    PPX_ASSERT_MSG(!metadata.name.empty(), "The metric name must not be empty");
+
+    *outMetric = nullptr;
+    if (GetMetric(metadata.name.c_str()) != nullptr) {
+        return ERROR_DUPLICATE_ELEMENT;
+    }
+
+    T* metric = new T(metadata);
+    if (metric == nullptr) {
+        return ERROR_OUT_OF_MEMORY;
+    }
+
+    const auto ret = mMetrics.insert({metadata.name, metric});
+    PPX_ASSERT_MSG(ret.second, "An insertion shall always take place when adding a metric");
+    *outMetric = metric;
+    return SUCCESS;
+}
 
 ////////////////////////////////////////////////////////////////////////////////
 
--- a/include/ppx/metrics.h
+++ b/include/ppx/metrics.h
@@ -99,7 +99,7 @@ public:
     size_t GetEntriesCount() const;
     void   GetEntry(size_t index, double& seconds, double& value) const;
 
-    const GaugeStatistics GetStatistics() const;
+    const GaugeStatistics GetStatistics(bool realtime) const;
 
 private:
     struct TimeSeriesEntry
--- a/include/ppx/metrics.h
+++ b/include/ppx/metrics.h
@@ -193,8 +193,7 @@ public:
     Manager();
     ~Manager();
 
-    Result AddRun(const char* name, Run** outRun);
-    Run*   GetRun(const char* name) const;
+    Run* AddRun(const char* name);
 
 private:
     METRICS_NO_COPY(Manager)
--- a/include/ppx/metrics.h
+++ b/include/ppx/metrics.h
@@ -147,6 +147,9 @@ private:
 
 ////////////////////////////////////////////////////////////////////////////////
 
+// A run gathers metrics relevant to the execution of a benchmark.
+// It is expected that a new run is created each time parameters that affect the
+// metrics measurements are changed.
 class Run final
 {
     friend class Manager;
--- a/include/ppx/metrics.h
+++ b/include/ppx/metrics.h
@@ -63,8 +63,6 @@ enum class MetricType
 
 class Metric
 {
-    friend class Run;
-
 public:
     const MetricMetadata& GetMetadata() const;
     MetricType            GetType() const;
@@ -95,6 +93,8 @@ struct GaugeStatistics
 
 class MetricGauge final : public Metric
 {
+    friend class Run;
+
 public:
     MetricGauge(const MetricMetadata& metadata);
 
@@ -157,26 +157,28 @@ private:
     ~Run();
     METRICS_NO_COPY(Run)
 
-    Metric* GetMetric(const char* name) const;
+    void AddMetric(MetricGauge* metric);
+    void AddMetric(MetricCounter* metric);
+    bool HasMetric(const char* name) const;
 
 private:
     std::string                              mName;
-    std::unordered_map<std::string, Metric*> mMetrics;
+    std::unordered_map<std::string, MetricGauge*> mGauges;
+    std::unordered_map<std::string, MetricCounter*> mCounters;
 };
 
 template <typename T>
 T* Run::AddMetric(MetricMetadata metadata)
 {
     PPX_ASSERT_MSG(!metadata.name.empty(), "The metric name must not be empty");
-    PPX_ASSERT_MSG(GetMetric(metadata.name.c_str()) == nullptr, "Metrics must have unique names (duplicate name detected)");
+    PPX_ASSERT_MSG(!HasMetric(metadata.name.c_str()), "Metrics must have unique names (duplicate name detected)");
 
     T* metric = new T(metadata);
     if (metric == nullptr) {
         return nullptr;
     }
 
-    const auto ret = mMetrics.insert({metadata.name, metric});
-    PPX_ASSERT_MSG(ret.second, "An insertion shall always take place when adding a metric");
+    AddMetric(metric);
     return metric;
 }
 
--- a/include/ppx/metrics.h
+++ b/include/ppx/metrics.h
@@ -150,7 +150,7 @@ class Run final
 
 public:
     template <typename T>
-    Result AddMetric(MetricMetadata metadata, T** outMetric);
+    T* AddMetric(MetricMetadata metadata);
 
 private:
     Run(const char* name);
@@ -165,25 +165,19 @@ private:
 };
 
 template <typename T>
-Result Run::AddMetric(MetricMetadata metadata, T** outMetric)
+T* Run::AddMetric(MetricMetadata metadata)
 {
-    PPX_ASSERT_MSG(outMetric != nullptr, "The metric pointer must not be null");
     PPX_ASSERT_MSG(!metadata.name.empty(), "The metric name must not be empty");
-
-    *outMetric = nullptr;
-    if (GetMetric(metadata.name.c_str()) != nullptr) {
-        return ERROR_DUPLICATE_ELEMENT;
-    }
+    PPX_ASSERT_MSG(GetMetric(metadata.name.c_str()) == nullptr, "Metrics must have unique names (duplicate name detected)");
 
     T* metric = new T(metadata);
     if (metric == nullptr) {
-        return ERROR_OUT_OF_MEMORY;
+        return nullptr;
     }
 
     const auto ret = mMetrics.insert({metadata.name, metric});
     PPX_ASSERT_MSG(ret.second, "An insertion shall always take place when adding a metric");
-    *outMetric = metric;
-    return SUCCESS;
+    return metric;
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/include/ppx/metrics.h
+++ b/include/ppx/metrics.h
@@ -15,6 +15,8 @@
 #ifndef ppx_metrics_h
 #define ppx_metrics_h
 
+#include "ppx/config.h"
+
 #include <limits>
 #include <string>
 #include <unordered_map>
@@ -30,22 +32,22 @@ namespace metrics {
 enum class MetricInterpretation
 {
     NONE,
-	HIGHER_IS_BETTER,
-	LOWER_IS_BETTER,
+    HIGHER_IS_BETTER,
+    LOWER_IS_BETTER,
 };
 
 struct Range
 {
-	double lowerBound = std::numeric_limits<double>::min();
-	double upperBound = std::numeric_limits<double>::max();
+    double lowerBound = std::numeric_limits<double>::min();
+    double upperBound = std::numeric_limits<double>::max();
 };
 
 struct MetricMetadata
 {
-    std::string name;
-	std::string unit;
-	MetricInterpretation interpretation = MetricInterpretation::NONE;
-	Range expectedRange;
+    std::string          name;
+    std::string          unit;
+    MetricInterpretation interpretation = MetricInterpretation::NONE;
+    Range                expectedRange;
 };
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -59,15 +61,15 @@ enum class MetricType
 class Metric
 {
 public:
-	const MetricMetadata& GetMetadata() const;
-    MetricType GetType() const;
+    const MetricMetadata& GetMetadata() const;
+    MetricType            GetType() const;
 
 protected:
     Metric(const MetricMetadata& metadata, MetricType type);
     virtual ~Metric();
 
 private:
-	MetricMetadata mMetadata;
+    MetricMetadata mMetadata;
     MetricType     mType;
 };
 
@@ -75,31 +77,31 @@ private:
 
 struct GaugeStatistics
 {
-	double min;
-	double max;
-	double average;
+    double min;
+    double max;
+    double average;
     double median;
-	double standardDeviation;
-	double timeRatio;
+    double standardDeviation;
+    double timeRatio;
 };
 
 class MetricGauge final : public Metric
 {
 public:
-	MetricGauge(const MetricMetadata& metadata);
+    MetricGauge(const MetricMetadata& metadata);
 
-	void RecordEntry(double seconds, double value);
-	size_t GetEntriesCount() const;
-	void GetEntry(size_t index, double& seconds, double& value) const;
+    void   RecordEntry(double seconds, double value);
+    size_t GetEntriesCount() const;
+    void   GetEntry(size_t index, double& seconds, double& value) const;
 
-	const GaugeStatistics GetStatistics() const;
+    const GaugeStatistics GetStatistics() const;
 
 private:
-	struct TimeSeriesEntry
-	{
-		double seconds;
-		double value;
-	};
+    struct TimeSeriesEntry
+    {
+        double seconds;
+        double value;
+    };
 
 private:
     std::vector<TimeSeriesEntry> mTimeSeries;
@@ -110,12 +112,12 @@ private:
 class MetricCounter final : public Metric
 {
 public:
-	MetricCounter(const MetricMetadata& metadata);
-	uint64_t Increment(uint64_t add);
-	uint64_t Get() const;
+    MetricCounter(const MetricMetadata& metadata);
+    uint64_t Increment(uint64_t add);
+    uint64_t Get() const;
 
 private:
-	uint64_t mCounter;
+    uint64_t mCounter;
 };
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -125,11 +127,11 @@ class Run final
 public:
     Run(const char* name);
     ~Run();
-	MetricGauge* AddMetricGauge(MetricMetadata metadata);
-	MetricCounter* AddMetricCounter(MetricMetadata metadata);
+    MetricGauge*   AddMetricGauge(MetricMetadata metadata);
+    MetricCounter* AddMetricCounter(MetricMetadata metadata);
 
 private:
-    std::string mName;
+    std::string                              mName;
     std::unordered_map<std::string, Metric*> mMetrics;
 };
 
@@ -140,7 +142,8 @@ class Manager final
 public:
     Manager();
     ~Manager();
-    Run* AddRun(const char* name);
+    Result AddRun(Run*& outRun, const char* name);
+    Run*   GetRun(const char* name) const;
 
 private:
     std::unordered_map<std::string, Run*> mRuns;
--- a/include/ppx/metrics.h
+++ b/include/ppx/metrics.h
@@ -59,21 +59,21 @@ struct MetricMetadata
 // They can be retrieved without any significant run-time cost.
 struct GaugeBasicStatistics
 {
-    double min;
-    double max;
-    double average;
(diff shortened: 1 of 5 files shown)


include/ppx/metrics.h
```diff
@@ -0,0 +1,198 @@
...
+
+class Run final
+{
+    friend class Manager;
+
+public:
+    Result       AddMetricGauge(MetricGauge*& outMetric, MetricMetadata metadata);
+    MetricGauge* GetMetricGauge(const char* name) const;
+
+    Result         AddMetricCounter(MetricCounter*& outMetric, MetricMetadata metadata);
```
Keenuts, line 155: Why not return the pointer directly? This combined with a templated `AddMetric` would yield the following code:

```cpp
const auto *metric = AddMetric<MetricGauge>(metadata);
PPX_ASSERT_MSG(metric != nullptr, "Failed to create metric blabla");
```

From the tests I see errors are sometimes used (like ERROR_DUPLICATE_ELEMENT), but also assert are used, like `PPX_ASSERT_MSG(!metadata.name.empty(), "The metric name must not be empty");`.
Is there a real value to add error messages vs asserts?

What I mean is if we provide an API, the user interacts with at runtime, sure, we don't want to assert and crash because they gave the wrong name. But as metrics seem to be hardcoded, wouldn't a simple clean assert be simpler to debug when developing the benchmark?

include/ppx/metrics.h
```diff
@@ -0,0 +1,198 @@
...
+    COUNTER,
+};
+
+class Metric
+{
+    friend class Run;
+
+public:
+    const MetricMetadata& GetMetadata() const;
+    MetricType            GetType() const;
```
Keenuts, line 70: Is this needed?
From what I see, the only cases we check for type are to return the correct metric from the tables. This could be removed by storing each metric in their own table right?

include/ppx/metrics.h
```diff
@@ -0,0 +1,198 @@
...
+
+private:
+    std::vector<TimeSeriesEntry> mTimeSeries;
+    GaugeStatistics              mRealTimeStatistics;
+    double                       mAccumlatedValue;
+};
+
+////////////////////////////////////////////////////////////////////////////////
+
+class MetricCounter final : public Metric
```
Keenuts, line 128: Comments? Assuming you follow the same principle as https://prometheus.io/docs/concepts/metric_types/ ?

include/ppx/metrics.h
```diff
@@ -0,0 +1,198 @@
...
+    double average;
+    double median;
+    double standardDeviation;
+    double timeRatio;
+    double percentile90;
+    double percentile95;
+    double percentile99;
+};
+
+class MetricGauge final : public Metric
```
Keenuts, line 96: Same, comments around this metric characteristics.

include/ppx/metrics.h
```diff
@@ -0,0 +1,198 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+
+enum class MetricType
+{
+    GAUGE,
+    COUNTER,
+};
+
+class Metric
```
Keenuts, line 64: Why inheritance at all?
From what I see, the only common bit is the MetricMetadata, which can be useful for printing, etc. But shouldn't the Metadata bit just be composed?
Seems like the rest of the shared metric API is limited.

include/ppx/metrics.h
```diff
@@ -0,0 +1,198 @@
...
+    ~MetricCounter();
+    METRICS_NO_COPY(MetricCounter)
+
+private:
+    uint64_t mCounter;
+};
+
+////////////////////////////////////////////////////////////////////////////////
+
+class Run final
```
Keenuts, line 147: I'm guessing this Manager/Run/Metric hierarchy is for the following use case:
 - 1 benchmark -> 1 manager
 - this benchmark has 2 things to test: 2 runs
 - each run as N metrics to monitor: N metrics
 
 What bugs me is the following:

* runs are independent.
  They don't need to share anything. Can have a different set of metrics attached to it. Should be name be different? (run feels like doing multiple runs of the same benchmark). Maybe something more generic, like Collection/Group?
  If the goal is indeed to have the same metric set, shouldn't the manager value be to check that? 

* No concept of "live/archived" run:
 We run a first test with Run#1. Then run another test with Run#2. If my metrics pointer are mishandled, I can still modify metrics of Run#1, no way to enforce this. Meaning the only value Run/Manager brings is allocation handling. Could it be beneficial for the manager to mark metrics/runs are immutable, and assert if a metric is updated outside of a run? (Creating a new run marks the old one immutable by ex).

include/ppx/metrics.h
```diff
@@ -0,0 +1,198 @@
...
+
+    Result         AddMetricCounter(MetricCounter*& outMetric, MetricMetadata metadata);
+    MetricCounter* GetMetricCounter(const char* name) const;
+
+private:
+    Run(const char* name);
+    ~Run();
+    METRICS_NO_COPY(Run)
+
+    Metric* GetMetric(const char* name) const;
```
Keenuts, line 163: related to my comment around the `Get` function existence. Going further, shouldn't the name simply be part of the metadata, which seem to be centered around human interpretation of the metric?
The only handle being the pointer returned by AddMetric?

Keenuts (commented):
Hi! Thanks for the contribution! I can see value to add a metric framework for ppx, as this could be used more broadly. Just some high-level comments around the API design 😊

include/ppx/metrics.h
```diff
@@ -0,0 +1,198 @@
...
+};
+
+class MetricGauge final : public Metric
+{
+public:
+    MetricGauge(const MetricMetadata& metadata);
+
+    void   RecordEntry(double seconds, double value);
+    size_t GetEntriesCount() const;
+    void   GetEntry(size_t index, double& seconds, double& value) const;
```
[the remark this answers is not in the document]
mdagois: Fixed!

include/ppx/metrics.h
```diff
@@ -0,0 +1,198 @@
...
+};
+
+////////////////////////////////////////////////////////////////////////////////
+
+class Run final
+{
+    friend class Manager;
+
+public:
+    Result       AddMetricGauge(MetricGauge*& outMetric, MetricMetadata metadata);
```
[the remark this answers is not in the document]
mdagois: Fixed!

include/ppx/metrics.h
```diff
@@ -0,0 +1,198 @@
...
+    MetricGauge* GetMetricGauge(const char* name) const;
```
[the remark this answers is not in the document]
mdagois: True, I am not using it yet, so I removed it. I'll re-add it in the future if that's necessary.

include/ppx/metrics.h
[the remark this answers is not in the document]
mdagois: Interestingly, I had a template as private method. I changed the code so that the public API is easier to use with a template.

include/ppx/metrics.h
mdagois: Agreed, I changed it for both metrics and runs creation. I added unit tests to check that the asserts trigger.

include/ppx/metrics.h
mdagois: I added comments in the metrics.h to explain what the metrics do, among other comments.

include/ppx/metrics.h
mdagois: Comments added.

include/ppx/metrics.h
mdagois: So, I was using it to have only one container, but as I switched to two containers, it became useless. No inheritance anymore.

include/ppx/metrics.h
mdagois: GetMetric is gone after the refactoring of the API.

include/ppx/metrics.h
```diff
@@ -0,0 +1,198 @@
...
+class MetricGauge final : public Metric
+{
+public:
+    MetricGauge(const MetricMetadata& metadata);
+
+    void   RecordEntry(double seconds, double value);
+    size_t GetEntriesCount() const;
+    void   GetEntry(size_t index, double& seconds, double& value) const;
+
+    const GaugeStatistics GetStatistics(bool realtime) const;
```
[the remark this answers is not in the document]
mdagois: I changed this, please see comment below.

src/ppx/metrics.cpp
```diff
@@ -0,0 +1,278 @@
...
+    mRealTimeStatistics.average = mAccumlatedValue / entriesCount;
+    if (entriesCount > 1) {
+        const double timeSpan         = mTimeSeries.back().seconds - mTimeSeries.front().seconds;
+        mRealTimeStatistics.timeRatio = mAccumlatedValue / timeSpan;
+    }
+    else {
+        mRealTimeStatistics.timeRatio = mTimeSeries.front().value;
+    }
+
+    // TODO Implement median, standard deviation and percentiles approximations if necessary at runtime
```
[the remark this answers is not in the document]
mdagois: I really like the idea of splitting the statistics. I split into GaugeBasicStatistics and GaugeComplexStatistics. I also split the method to get them (using compute for the complex statistics to highlight the need to compute them). Please let me know what you think.

include/ppx/metrics.h
```diff
@@ -0,0 +1,198 @@
...
+    double percentile95;
+    double percentile99;
+};
+
+class MetricGauge final : public Metric
+{
+public:
+    MetricGauge(const MetricMetadata& metadata);
+
+    void   RecordEntry(double seconds, double value);
```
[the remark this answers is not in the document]
mdagois: I added comments about the seconds and also in several places in the header.

include/ppx/metrics.h
```diff
@@ -0,0 +1,198 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+
+class Manager final
+{
+public:
+    Manager();
+    ~Manager();
+
+    Result AddRun(Run*& outRun, const char* name);
```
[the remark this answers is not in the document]
mdagois: Yes, swapped.

include/ppx/metrics.h
mdagois: In the context of the benchmark, the expected flow is:
1. set the knobs to particular values
2. create a run
3. add metrics to the run that are relevant to the knobs
4. record time series for each metrics
5. repeat for as long as needed

Yes, runs are independent. To be honest, I could not find a better name. I thought about benchmark, but that would be even more confusing for obvious reasons. Collection/Group feel a bit too generic. I will add a comment over the class to try and make it more obvious what it is.

It might definitely be beneficial to block editing... that should be easy. I will think about it in a subsequent PR, once I implement the rest of the system and start seeing more use cases.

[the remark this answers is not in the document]
mdagois: I added a comment. PTAL.

src/ppx/metrics.cpp
```diff
@@ -0,0 +1,219 @@
...
+    }
+    mRuns.clear();
+}
+
+Run* Manager::AddRun(const char* name)
+{
+    PPX_ASSERT_MSG(name != nullptr, "A run name must not be null");
+    PPX_ASSERT_MSG(mRuns.find(name) == mRuns.end(), "Runs must have unique names (duplicate name detected)");
+
+    Run* run = new Run(name);
```
Keenuts, line 206: curious: WDYT about unique_ptr for this? (just to avoid having to handle/check new/delete). Since we return a raw pointer anyway.

src/ppx/metrics.cpp
```diff
@@ -0,0 +1,219 @@
...
+    if (run == nullptr) {
+        return nullptr;
+    }
+
+    const auto ret = mRuns.insert({name, run});
+    PPX_ASSERT_MSG(ret.second, "An insertion shall always take place when adding a run");
```
Keenuts, line 212: Does this causes a "unused variable ret" warning in release builds?

Keenuts (commented):
Curious, for the IO, do you plan on having a "standard" format for metrics, like the line format for influxDB/prometheus?

mdagois:
>

mdagois:
> Curious, for the IO, do you plan on having a "standard" format for metrics, like the line format for influxDB/prometheus?

@Keenuts The original design was to use protobuf and a custom export. However, I am thinking about exporting to something perfetto-friendly. I have not done much research on those formats, so any suggestion is very welcome.

src/ppx/metrics.cpp
```diff
@@ -0,0 +1,219 @@
...
+void Run::AddMetric(MetricCounter* metric)
+{
+    PPX_ASSERT_NULL_ARG(metric);
+    const auto ret = mCounters.insert({metric->mMetadata.name, metric});
+    PPX_ASSERT_MSG(ret.second, "An insertion shall always take place when adding a metric");
+}
+
+bool Run::HasMetric(const char* name) const
+{
+    return mGauges.find(name) != mGauges.end() || mCounters.find(name) != mCounters.end();
```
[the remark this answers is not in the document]
mdagois: I am still not sure I will keep the map in the long run to be honest. As I add the rest of the functionalities, I will re-evaluate this container.
That's why on insertion, I assert that the metric name is unique over all containers. It keeps the option to merge the metrics into a single container later on.

src/ppx/metrics.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+void MetricGauge::UpdateBasicStatistics(double seconds, double value)
+{
+    mAccumulatedValue += value;
+
+    mBasicStatistics.min = std::min(mBasicStatistics.min, value);
+    mBasicStatistics.max = std::max(mBasicStatistics.max, value);
+
+    const size_t entriesCount = mTimeSeries.size();
+    mBasicStatistics.average  = mAccumulatedValue / entriesCount;
```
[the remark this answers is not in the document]
mdagois: Yeah, exactly. That's why I use double everywhere. Besides the space consumption, they are usually faster (though the speed boost is probably completely negligeable) and the added precision just eases everything. However, once we have use case for actual metrics, maybe it is worth checking.
Feel free to change it too in a CL. I don't think there would be any drawbacks to using **value / count**.

src/ppx/metrics.cpp
mdagois: Good point, but no. It seems the BW assert macro always uses the condition to do some logging. Whether the assert triggers or not is all from cassert.

src/ppx/metrics.cpp
```diff
@@ -0,0 +1,219 @@
...
+}
+
+const GaugeBasicStatistics MetricGauge::GetBasicStatistics() const
+{
+    return mBasicStatistics;
+}
+
+const GaugeComplexStatistics MetricGauge::ComputeComplexStatistics() const
+{
+    GaugeComplexStatistics statistics   = {};
```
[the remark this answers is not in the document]
mdagois: I added initialization values to statistics. Fixed.

src/ppx/metrics.cpp
```diff
@@ -0,0 +1,219 @@
...
+        delete metric;
+    }
+    mGauges.clear();
+    for (auto [name, metric] : mCounters) {
+        delete metric;
+    }
+    mCounters.clear();
+}
+
+void Run::AddMetric(MetricGauge* metric)
```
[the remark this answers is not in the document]
mdagois: I prefer the non-templated funtions. If we ever add another type, it will break at compile time 100% of time. It is straightforward, so I like it better. :)

include/ppx/metrics.h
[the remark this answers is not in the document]
mdagois: Oops, forgot those. I turned them into pointers. As for the naming, let me do it in another CL. I don't think I used any "p" prefix anywhere.
Is the naming a BW convention? I'd rather not mingle the type with the name of variables precisely because when you refactor, you have to rename variables.
I will comply to whatever BW rules there is though.

mdagois:
Note: sorry I think I closed the PR by mistake. Re-opened it.

src/ppx/metrics.cpp
mdagois: Please let me keep it in mind for a subsequent CL. I might actually change the container, etc., once I implement more of the design.

include/ppx/metrics.h
mdagois: Yes, the metric super class is gone.

include/ppx/metrics.h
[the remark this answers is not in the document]
Keenuts: It's a bigwheels convention (mostly observed across the project as it's not enforced by tools). It's to follow Microsoft's convention:
 - pMyPointer
 - ppMyPointerOfPointer
 - pppMyPointerOfPointerOfPointer

src/ppx/metrics.cpp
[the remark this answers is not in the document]
Keenuts: Might be out of scope for the design, but what about doing like the public metric systems out there:
 - client only exports raw metrics
 - some other service handles the statistics and dashboards.
 
 If you look at prometheus, each producer just outputs raw data over TCP/something, and then you have a tool like grafana to visualize metrics and see human-readable statistics.
 This way no need to worry about performance of statistical analysis on the producer, and minimize impact on the benchmark.

src/ppx/metrics.cpp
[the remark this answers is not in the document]
Keenuts: Fair

src/ppx/metrics.cpp
Keenuts: Thanks for checking :)

Keenuts:
> 



> > Curious, for the IO, do you plan on having a "standard" format for metrics, like the line format for influxDB/prometheus?
> 
> @Keenuts The original design was to use protobuf and a custom export. However, I am thinking about exporting to something perfetto-friendly. I have not done much research on those formats, so any suggestion is very welcome.

Perfetto seems to be a good option for this kind of project. In that case, shouldn't the whole statistic computations be left for perfetto, and this API only allow exporting basic traces?

include/ppx/metrics.h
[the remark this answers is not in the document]
mdagois: OK, let me fix all these in a separate, dedicated, CL then.

src/ppx/metrics.cpp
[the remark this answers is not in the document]
mdagois: Actually, that's in the plan to export raw data in a proprietary format (using protobuf) and the purpose is exactly to have it easily grabbed by external tools.
The inclusion of metrics in the API is for a more direct feedback.

mdagois:
> > 
> 
> > > Curious, for the IO, do you plan on having a "standard" format for metrics, like the line format for influxDB/prometheus?
> > 
> > 
> > @Keenuts The original design was to use protobuf and a custom export. However, I am thinking about exporting to something perfetto-friendly. I have not done much research on those formats, so any suggestion is very welcome.
> 
> Perfetto seems to be a good option for this kind of project. In that case, shouldn't the whole statistic computations be left for perfetto, and this API only allow exporting basic traces?

I think it's worth keeping the functionality in the API, as the cost of maintaining it is quite minimal.

src/test/metrics_test.cpp
```diff
@@ -0,0 +1,414 @@
...
+    constexpr const char* RUN_NAME = "run";
+    {
+        metrics::Run* run = manager.AddRun(RUN_NAME);
+        EXPECT_NE(run, nullptr);
+    }
+    {
+        EXPECT_DEATH({
+            manager.AddRun(RUN_NAME);
+        },
+                     "");
```
Keenuts, line 106: This is clang format going wild I suppose?

src/test/metrics_test.cpp
```diff
@@ -0,0 +1,414 @@
...
+    EXPECT_EQ(metric->GetEntriesCount(), 3U);
+    metric->RecordEntry(0.0349, 10.8);
+    EXPECT_EQ(metric->GetEntriesCount(), 4U);
+    metric->RecordEntry(0.0457, 11.1);
+    EXPECT_EQ(metric->GetEntriesCount(), 5U);
+
+    double seconds;
+    double value;
+    metric->GetEntry(0, &seconds, &value);
+    EXPECT_EQ(seconds, 0.0000);
```
Keenuts, line 240: EXPECT_NEAR? (input values above cannot be represented as doubles (exactly), so might want to add some margins?

src/test/metrics_test.cpp
```diff
@@ -0,0 +1,414 @@
...
+    metric->RecordEntry(1.1423114793963198, 11.310652174786224);
+
+    const metrics::GaugeBasicStatistics basicStatistics = metric->GetBasicStatistics();
+    EXPECT_EQ(basicStatistics.min, 10.50685559634115);
+    EXPECT_EQ(basicStatistics.max, 12.465256618853836);
+    EXPECT_DOUBLE_EQ(basicStatistics.average, 11.53622131571106);
+    EXPECT_DOUBLE_EQ(basicStatistics.timeRatio, 1009.9015482018648);
+
+    const metrics::GaugeComplexStatistics complexStatistics = metric->ComputeComplexStatistics();
+    EXPECT_EQ(complexStatistics.median, 11.526757950183075);
```
Keenuts, line 407: Seems like below you used the approximate EXPECT_DOUBLE_EQ, so why not here?

Keenuts (approved):
Assuming their will be a follow-up PR to fix the format/naming issues.
Most comments lefts are around the float precision in testing.

src/test/metrics_test.cpp
mdagois: Yeah :(

src/test/metrics_test.cpp
mdagois: There are no computation involved, so I expect the values to be exactly the same.

src/test/metrics_test.cpp
mdagois: Ah good catch. This one does actually lead to a computation when the number of entries is even. Fixed.

mdagois:
> Assuming their will be a follow-up PR to fix the format/naming issues. Most comments lefts are around the float precision in testing.

Next PR on my list. Trust me. ( ¬ ‿¬)

src/ppx/metrics.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+    UpdateBasicStatistics(seconds, value);
+}
+
+size_t MetricGauge::GetEntriesCount() const
+{
+    return mTimeSeries.size();
+}
+
+void MetricGauge::GetEntry(size_t index, double* seconds, double* value) const
```
apazylbe, line 52: Should we validate that index is in range?

include/ppx/metrics.h
```diff
@@ -0,0 +1,215 @@
...
+// A counter metric represents a value that only goes up, e.g. the number
+// of stutters or pipeline cache misses.
+class MetricCounter final
+{
+    friend class Run;
+
+public:
+    MetricCounter(const MetricMetadata& metadata);
+
+    uint64_t Increment(uint64_t add);
```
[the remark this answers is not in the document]
mdagois: Not that useful. I removed it.

include/ppx/metrics.h
```diff
@@ -0,0 +1,215 @@
...
+    // Record a measurement for the metric at a particular point in time.
+    // Each entry must have a positive 'seconds' that is greater than the
+    // 'seconds' of the previous entry (i.e. 'seconds' must form a stricly
+    // increasing positive function).
+    // Note however that the system does NOT assume that the 'seconds' of
+    // the first entry equal zero.
+    void RecordEntry(double seconds, double value);
+
+    size_t GetEntriesCount() const;
+    void   GetEntry(size_t index, double* seconds, double* value) const;
```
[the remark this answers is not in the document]
mdagois: I added a comment over the two functions.

src/ppx/metrics.cpp
mdagois: I added an assert and a unit test to check the assert triggers.

src/test/metrics_test.cpp
Keenuts: What about the values which are not exactly representable as floats? Nor sure if the conversion to the closest representation is defined.

mdagois: Yeah, I doubt we still have those around. :)

PR #128, opened 2023-04-24
Fix --resolution flag for XR devices.

at-ninja:
If a target resolution is specified via the `--resolution` command line flag, the value will be passed through to XrComponent.

### DIFF
--- a/include/ppx/xr_component.h
+++ b/include/ppx/xr_component.h
@@ -49,18 +49,24 @@ enum struct XrRefSpace
     XR_STAGE,
 };
 
+struct XrComponentResolution
+{
+    uint32_t width  = 0;
+    uint32_t height = 0;
+};
+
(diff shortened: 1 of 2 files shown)


include/ppx/xr_component.h
```diff
@@ -94,12 +101,16 @@ class XrComponent
     {
         if (mConfigViews.empty())
             return 0;
+        if (mCreateInfo.resolution.width > 0)
```
Keenuts, line 104: shouldn't that be returned only while there are no view configured?
(Not familiar with how XR manages view, but I'd guess XR takes the size hint, but might return a view with a different size once created no?)

Keenuts: oh ok, thanks

Keenuts (approved):
Thanks for the PR! 😊

PR #129, opened 2023-04-25
android: switch back to NativeActivity.

Keenuts:
NativeActivity is supported since a long time. GameActivity is the new recommended method.
THe issue is GameActivity is not supported by Android, and being brand new means lack of support in other domains.

Moving back to NativeActivity is a risk, if it suddenly gets deprecated. But the cost of the switch being lower than the cost of fixing dependencies to use GameActivity means it's an acceptable risk.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -195,7 +195,7 @@ if (NOT PPX_ANDROID)
 endif()
 
 # ------------------------------------------------------------------------------
-# Add game-activity for ANDROID
+# Add native activity for ANDROID
 # ------------------------------------------------------------------------------
 if (PPX_ANDROID)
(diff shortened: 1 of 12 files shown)


Keenuts:
@at-ninja : are you OK with this PR? (cannot add you as reviewer)

CMakeLists.txt
```diff
@@ -198,7 +198,11 @@ endif()
 # Add game-activity for ANDROID
```
apazylbe, line 198: Should we update this comment?

apazylbe:
I think @hysw had a similar change, so adding in case he has any comments

hysw (approved):
This is essentially the same change I had

at-ninja (commented):
LGTM, thanks for the heads up.

PR #135, opened 2023-04-25
Add vs/ps shaders for benchmarks pipeline project

RenfengLiu:
This implemented the 2 VS shaders and 3 PS shaders described in the benchmark doc.

### DIFF
--- a/benchmarks/graphics_pipeline/main.cpp
+++ b/benchmarks/graphics_pipeline/main.cpp
@@ -185,7 +185,7 @@ void ProjApp::Config(ppx::ApplicationSettings& settings)
     settings.window.width               = 1920;
     settings.window.height              = 1080;
     settings.grfx.api                   = kApi;
-    settings.grfx.enableDebug           = true;
+    settings.grfx.enableDebug           = false;
     settings.grfx.swapchain.depthFormat = grfx::FORMAT_D32_FLOAT;
 }
 


assets/benchmarks/shaders/Benchmark_PsMemBound.hlsl
```diff
@@ -0,0 +1,102 @@
...
+    const float4 albedo4 = AlbedoTexture.Sample(AlbedoSampler, uv4).rgba;
+    const float4 albedo5 = AlbedoTexture.Sample(AlbedoSampler, uv5).rgba;
+
+    const float4 albedo = (albedo1 + albedo2 + albedo3 + albedo4 + albedo5)/5.0;
+    
+    const float roughness1 = MetalRoughness.Sample(MetalRoughnessSampler, uv1).g;
+    const float roughness2 = MetalRoughness.Sample(MetalRoughnessSampler, uv2).g;
+    const float roughness3 = MetalRoughness.Sample(MetalRoughnessSampler, uv3).g;
+    const float roughness4 = MetalRoughness.Sample(MetalRoughnessSampler, uv4).g;
+    const float roughness5 = MetalRoughness.Sample(MetalRoughnessSampler, uv4).g;
```
wangra-google, line 81: the compiler will probably remove the 5th sample here since it is using the same uv as roughness4

assets/benchmarks/shaders/Benchmark_VsAluBound.hlsl
```diff
@@ -0,0 +1,57 @@
...
+      0, 0, 0, 1,
+  };
+
+  float4 jointWeights = {
+    frac(uv.x), 1 - frac(uv.x), 0, 0
+  };
+
+  float4 jointIndices = {0, 1, 0, 0};
+
+  float4x4 skinMat = jointWeights.x * jointMat[int(jointIndices.x)] +
```
wangra-google, line 42: the compiler probably is most likely able to optimize the case where the weight is 0 (in this case, we probably need to make the shader more complex to make sure the compiler is not doing any unwanted optimizations). can you put some comments here to check this later?

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -637,18 +650,24 @@ void ProjApp::Setup()
         PPX_CHECKED_CALL(GetDevice()->CreateDescriptorPool(&poolCreateInfo, &mDescriptorPool));
     }
 
-    for (size_t i = 0; i < kAvailableShaders.size(); i++) {
-        const std::string shaderBaseName = kAvailableShaders[i];
+    uint32_t pipeline_index = 0;
+    for (size_t i = 0; i < kAvailableVsShaders.size(); i++) {
+        for (size_t j = 0; j < kAvailablePsShaders.size(); j++) {
```
apazylbe, line 655: It seems like we are creating the same VS and PS shaders multiple times. Do we expect to be adding more shaders? I am curious if there a reason not to create each unique one once and then reference it

apazylbe (commented):
I think it's preferable to add new commits instead of amending the commit to make it easier to review according to the [contributing page](https://github.com/google/bigwheels/blob/main/CONTRIBUTING.md#for-developers-contributing-a-patch)

benchmarks/graphics_pipeline/main.cpp
wangra-google: the shaders for benchmark are supposed to be unique, we also might need to tune them in the future.

wangra-google (approved):

apazylbe (approved):

PR #136, opened 2023-04-25
Add abstraction for platform window operations.

hysw:
- So we can remove 20 of `#if defined`
- Remove platform specific clutter from Application

### DIFF
--- a/include/ppx/window.h
+++ b/include/ppx/window.h
@@ -1,0 +1,98 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ppx_window_h
+#define ppx_window_h
+
+#include "ppx/config.h"
+
+#include <memory>
+
+namespace ppx {
+
+namespace grfx {
+struct SurfaceCreateInfo;
+} // namespace grfx
+
+class Application;
+
+struct WindowSize
+{
+    uint32_t width  = 0;
+    uint32_t height = 0;
+
+    WindowSize() = default;
+    WindowSize(uint32_t width_, uint32_t height_)
+        : width(width_), height(height_) {}
+
+    bool operator==(const WindowSize&) const = default;
+};
+
+class Window
+{
+public:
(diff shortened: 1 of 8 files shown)


hysw:
- d17a434 is just copying part of application.cpp so later diff actually make sense
- 1cfd0fc contains the actual diffs

include/ppx/window.h
```diff
@@ -0,0 +1,94 @@
...
+    WindowSize(uint32_t width_, uint32_t height_)
+        : width(width_), height(height_) {}
+
+    bool operator==(const WindowSize&) const = default;
+};
+
+class Window
+{
+public:
+    static Window* New(Application* app);
```
wangra-google, line 45: when i first see this function, I am confused: should i call `New` or `Create`? Also normally i would assume `Init` being called after `Create` which is not the case here (seems need to call `Init` before `Create`). Maybe it is just me, but I feel `New` is not commonly used as a function name. Not sure what would be a better name here though, but it would be better to have at least some comments here to explain a how `Init`, `New`, `Create` should be called. Can we merge `Init` and `Create` into 1 function since they are called together? (if so, it might be better to name this merged function as `Init` and change the `New` function to `Create`.)

chaoticbob: The typical pattern that BigWheels has is if `Create*(...)` is static - the function will allocate a new object or return a copyable object. Otherwise `Create(...)` as a member function is the first function use for creating the object internals. `Create*` and `Initialize*` are typically used to create internal objects, eg API objects. I've tried to avoid functions that look like `new` or `delete` unless there's an explicit need to allocate or free something.

hysw: Init and Create should be able to be merged, the Init part used to be in InitializePlatform, and it seems to be possible to move it into CreatePlatformWindow after poking at it for a bit.

The New part is a dispatch for different windowing system, and it's basically a `new WindowImpl{Platform}(Application*)` call

hysw: I guess I'll keep one of those `#if` statement around so we don't have this mess

hysw: Updated.

The platform window implementation should really be passed from main/android_main to Application instead of been created by Application. But that's a much more complicated change and I'll not attempt it here

wangra-google (approved):

include/ppx/window.h
```diff
@@ -0,0 +1,98 @@
...
+    WindowSize(uint32_t width_, uint32_t height_)
+        : width(width_), height(height_) {}
+
+    bool operator==(const WindowSize&) const = default;
+};
+
+class Window
+{
+public:
+    static std::unique_ptr<Window> GetImplHeadless(Application*);
```
chaoticbob, line 45: Why is the ownership of this changed to `unique_ptr`? This implies that either the object is self owned or is dependent on a function calling reset? The general design is to use explicit create/destroy idioms. This breaks that design.

chaoticbob (changes requested):

chaoticbob (approved):
LGTM

wangra-google (approved):

Issue #118, opened 2023-04-20
Android: check Android build in CI


vettoreldaniele:
A new CI workflow should be added to test that BigWheels for Android builds correctly.

apazylbe:
If anyone else is already looking at or planning to, please let me know!

PR #139, opened 2023-04-26
Add a new workflow to check if Android builds

apazylbe:
Builds a single sample for mobile - Gbuffer

I think we could also combine all CI checks (windows, linux, android) into a single workflow with multiple jobs. But since the existing linux and windows are 2 different workflows, I created a new one for now 

Fixes https://github.com/google/bigwheels/issues/118

### DIFF
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -1,0 +1,35 @@
+name: Android build
+
+on:
+  pull_request:
+
+# Cancel previous runs if a more recent commit is pushed.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
+
+permissions: read-all
+
+jobs:
+  android-build:
+    name: Check Android build with gradle
+    runs-on: "ubuntu-20.04"
+    strategy:
+      fail-fast: true
+    steps:
+      - name: Checkout the code
+        uses: actions/checkout@v3
+        with:
+          submodules: 'true'
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@v2
+      - name: Setup Vulkan SDK
+        run: |
+          wget -q https://sdk.lunarg.com/sdk/download/1.3.216.0/linux/vulkansdk-linux-x86_64-1.3.216.0.tar.gz
+          mkdir "${HOME}/vulkan-sdk"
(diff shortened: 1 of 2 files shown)


Keenuts:
What's the impact to combine all CI builds into one workflow?
What is one job fails (ex linux), are the other cancelled? If yes seems counterproductive vs having all workflows run to completion so we show multiple failures no?

.github/workflows/android-build.yml
```diff
@@ -0,0 +1,35 @@
...
+        uses: gradle/gradle-build-action@v2
+      - name: Setup Vulkan SDK
+        run: |
+          wget -q https://sdk.lunarg.com/sdk/download/1.3.216.0/linux/vulkansdk-linux-x86_64-1.3.216.0.tar.gz
+          mkdir "${HOME}/vulkan-sdk"
+          tar -xf vulkansdk-linux-x86_64-1.3.216.0.tar.gz -C "${HOME}/vulkan-sdk"
+          echo "VULKAN_SDK=${HOME}/vulkan-sdk/1.3.216.0/x86_64" >> $GITHUB_ENV
+      - name: Build Cube sample for Android
+        run: |
+          ./gradlew assembleMobileGbufferDebug
```
Keenuts, line 34: Why only build 1 sample? is there a way to query what rules are available to build all the samples?

apazylbe: I updated it to build all samples for mobile only. This job seems to take about ~15 minutes which is slightly longer than what windows and linux builds take. Including Openxr targets increases the total time to ~30 minutes. Quest targets rely on the oculus OpenXR loader, which we'll have to figure something out for if we wanted to test those as well. 
We could do mobile and openXr now, but I kinda don't like the idea of increasing the CI time. Thoughts?

apazylbe:
> What's the impact to combine all CI builds into one workflow? What is one job fails (ex linux), are the other cancelled? If yes seems counterproductive vs having all workflows run to completion so we show multiple failures no?

From what I read so far there is really no reason to use one over the other for Bigwheels right now. A single workflow with multiple jobs would still report multiple failures, the output looks identical. I think we can keep multiple workflows for now since that's what we already have, and change it if necessary later.

Keenuts (approved):
Not familiar with the concurrency clause, thought PR runs were already cancelled on push. But we'll see 😊

PR #141, opened 2023-04-27
android: change how we replace manifest values

Keenuts:
The initial method I found was using resource strings. But we can actually replace them at compile-time using those placeholder variables. This would simplify integration with other build-systems.


Keenuts:
Was able to remove most manifest changes leading to the same effect. Also cleaned up some useless manifest infos.

apazylbe (approved):

PR #142, opened 2023-04-28
Android: specify project specific targets in the build

apazylbe:
So that only the library needed for the sample is built and packaged


Keenuts (approved):
Oh nice, thanks for this!

PR #143, opened 2023-04-28
Fixed warnings and some style issues

chaoticbob:
- Fixed cast warnings for some pointers and integers
- Moved multiline function(s) from header to source
- Renamed mSwapchain to mSwapchains since it can store more than one swapchain
- Fixed some styling issues (overall style guidance will be addressed later)


apazylbe (changes requested):
Looks like there are still places where mSwapchain needs to be renamed in application.cpp https://github.com/google/bigwheels/actions/runs/4832779848/jobs/8611983129#step:6:366

chaoticbob:
> Looks like there are still places where mSwapchain needs to be renamed in application.cpp https://github.com/google/bigwheels/actions/runs/4832779848/jobs/8611983129#step:6:366

Thanks. Fixed (hopefully).

apazylbe (approved):

PR #145, opened 2023-05-01
Update clang-format version and one-time format

apazylbe:
Switch to using an available github action for clang-format CI check. This method checks the entire codebase as opposed to just changes from the PR, so this also includes a one time formatting for the entire project to bring it a "clean" state.


apazylbe:
If we do want to upgrade clang-format version, this is one way to do it. We could also consider doing this after we establish formatting style, in case we need to apply more 1-time format changes to the project

PR #147, opened 2023-05-01
Ignore android window size when we are in XR.

hysw:


at-ninja (approved):

Keenuts (approved):
Seems fine, except why? The commit message should definitely explain why such change is needed.

hysw:
> Seems fine, except why? The commit message should definitely explain why such change is needed.

On quest2 the window size is the entire display :(
BigWheels uses window size for openxr swapchain size.

PR #148, opened 2023-05-01
Use Application::{GetScissor,GetViewport}().

hysw:
It's used by some samples already.

Some benchmark has more complicated logic regarding viewport size, so they are left alone.


chaoticbob (approved):
LGTM

Keenuts (approved):
Thanks for this small improvement 😊

PR #150, opened 2023-05-02
Explicitly set the gradle build action version

apazylbe:
There seems to be a vulnerability fixed in 2.4.2.

### DIFF
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -22,7 +22,7 @@ jobs:
         with:
           submodules: 'true'
       - name: Setup Gradle
-        uses: gradle/gradle-build-action@v2
+        uses: gradle/gradle-build-action@v2.4.2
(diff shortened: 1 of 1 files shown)


Keenuts:
Isn't the security report saying `v2` is fine? (as it's automatically updated to the latest v2.x.x ?)

Keenuts:
(I don't understand why the error is still showing TBH, the report says "no action required", and we have no cache)

Keenuts:
Let's try merge, see if it solves, and then revert?
I don't want to fix the version as we won't be able to benefit from those patches in the future if we hardcode `v2.4.2`.

apazylbe:
Makes sense, it looks like this was published recently, so maybe there will be a better automated check for it later? agree, that it's better to have v2 so we get the latest

Keenuts (approved):

Keenuts:
Alert is closed.

PR #153, opened 2023-05-03
Add command line arguments for fishtornado_xr

Slumpwuffle:
This change updates the XR version of fish tornado to accept command-line arguments controlling the number of fish, whether to render ocean, shark and fish, and whether to enable head tracking


Keenuts (approved):

projects/fishtornado_xr/FishTornado.cpp
```diff
@@ -407,6 +407,30 @@ void FishTornadoApp::SetupScene()
 
 void FishTornadoApp::Setup()
 {
+    const auto& cl_options        = GetExtraOptions();
```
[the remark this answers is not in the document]
Slumpwuffle: Ah yeah - other bits of code were using the underscore; likely just cleanup we'll need to do later.

projects/fishtornado/Flocking.h
```diff
@@ -64,8 +65,9 @@ class Flocking
         bool renderedWithAsyncCompute = false;
     };
 
-    uint32_t mResX;
-    uint32_t mResY;
+    const FishTornadoSettings* pSettings;
```
apazylbe, line 68: Is there any particular reason to hold on to the settings instead of fishThreadsX and fishThreadsY?

projects/fishtornado/Ocean.h
```diff
@@ -22,14 +22,15 @@ using namespace ppx;
...
+struct FishTornadoSettings;
 
 class Ocean
 {
 public:
     Ocean();
     ~Ocean();
 
-    void Setup(uint32_t numFramesInFlight);
+    void Setup(uint32_t numFramesInFlight, const FishTornadoSettings& settings);
```
apazylbe, line 33: Why do ocean and shark need settings?

apazylbe:
A few comments, and adding @chaoticbob for another look as well

- It looks like the PR now updates the non-XR version as well, could you update the title and description?
- I was going to say we should update the README for the project, but it looks like we don't have one, perhaps we can add one later and include the command line options.
- I am having some issues while adjusting these settings (disabling fish), does everything run fine for you?

projects/fishtornado/FishTornado.cpp
```diff
@@ -398,6 +398,29 @@ void FishTornadoApp::SetupScene()
...
+    bool forceSingleCommandBuffer = clOptions.HasExtraOption("ft-use-single-command-buffer");
+    bool useAsyncCompute          = !clOptions.HasExtraOption("ft-use-async-compute");
+    if (forceSingleCommandBuffer && useAsyncCompute) {
+        PPX_LOG_WARN("Single command buffer selected WITH async compute! Disabling async compute!");
+        useAsyncCompute = false;
+    }
+    mSettings.forceSingleCommandBuffer = forceSingleCommandBuffer;
+    mSettings.useAsyncCompute          = useAsyncCompute;
+    mSettings.renderFish               = !clOptions.HasExtraOption("ft-disable-fish");
+    mSettings.renderOcean              = !clOptions.HasExtraOption("ft-disable-ocean");
```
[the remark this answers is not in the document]
Slumpwuffle: yes, we probably should. But probably in a separate PR?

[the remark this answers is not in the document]
apazylbe: Yeah, definitely separate PR, I think I was more so thinking that we should use here whatever we are likely to use in the future, but no action here for now

projects/fishtornado/Flocking.h
Slumpwuffle: It's actually not the threads that's problematic - it's the enable/disable head tracking specifically. Because that's a live-updatable setting in the main FishTornado.cpp, it can't be stored off. I could do a pointer specifically for that setting, but I also figured it would be easy enough to keep a pointer to the entire settings structure. Thoughts?

projects/fishtornado/Ocean.h
Slumpwuffle: They don't yet. Can eliminate this part; I just liked the symmetry.

Slumpwuffle:
Will double-check.

projects/fishtornado/Flocking.h
apazylbe: Ah, yes, I thought threads that are updatable but its the head tracking, I was looking at the non-xr code.
I guess we could have had something like a method on flocking to set/update parameters, but I don't know if that's better. This is fine then, perhaps a comment somewhere to document this might help?

projects/fishtornado/Ocean.h
apazylbe: Do you anticipate they would use it in the future? I'd probably keep them if that's the case and remove them otherwise, but up to you

Slumpwuffle: No value in code that anticipates for a future that doesn't exist. I'll clean it up, I think.

projects/fishtornado/Flocking.h
Slumpwuffle: I think it's probably better to separately track head info, given that the non-XR version doesn't need it. Let me clean this up a bit.

projects/fishtornado/FishTornado.cpp
```diff
@@ -398,6 +398,29 @@ void FishTornadoApp::SetupScene()
...
+    mSettings.usePCF              = !clOptions.HasExtraOption("ft-disable-pcf-shadows");
+    bool forceSingleCommandBuffer = clOptions.HasExtraOption("ft-use-single-command-buffer");
+    bool useAsyncCompute          = !clOptions.HasExtraOption("ft-use-async-compute");
+    if (forceSingleCommandBuffer && useAsyncCompute) {
+        PPX_LOG_WARN("Single command buffer selected WITH async compute! Disabling async compute!");
+        useAsyncCompute = false;
+    }
+    mSettings.forceSingleCommandBuffer = forceSingleCommandBuffer;
+    mSettings.useAsyncCompute          = useAsyncCompute;
+    mSettings.renderFish               = !clOptions.HasExtraOption("ft-disable-fish");
```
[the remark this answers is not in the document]
Slumpwuffle: Correct. There appear to be other arguments around that use this approach. I think I'd rather be flexible about it, though... this has been bugging me. I'll adjust it so --ft-disable-fish insinuates --ft-disable-fish=true but you can still flag it differently.

Slumpwuffle:
Had some bad logic in the actual arguments (re: disabling vs. enabling) and fixed an issue with a semaphore wait. Seems to work well now.

apazylbe (approved):

apazylbe:
Thank you! I updated the commit message to include addition on non-XR project flags.

By the way, the [contributing](https://github.com/google/bigwheels/blob/main/CONTRIBUTING.md#for-developers-contributing-a-patch) page recommends adding new commits, instead of amending so that it's easier to review changes.

PR #154, opened 2023-05-03
fix 01_triangle GetViewport()/GetScissor()

Keenuts:
Those members were removed in #148, but forgotten here. Fixing this.


apazylbe (approved):
I was confused as to why this wasn't caught in CI check, but it turns out the changes were out of date on the branch when it was tested, so it's a corner case we don't catch now. Wondering if we should run CI on push, or if there is anything else we can do to catch these cases.

hysw:
Sorry, I did not merge change #148 due to it have build conflicting with #144 , and were about to rebase this morning. Probably should have commented on it.

PR #162, opened 2023-05-05
Fix debug output of fishtornado_xr

Slumpwuffle:
This change reenables the extended debug output of fishtornado_xr. To do so, it tracks metrics for each view needed by the XR layer, and then presents the sum (e.g. for GPU time, vertices, etc.). It also sets the window position for the diagnostic output to anchor to windowWidth/4 and windowHeight/4, with a pivot point at 0.5x0.5. This theoretically serves to center the diagnostic output on XR as the imgui surface is the same size as each eye, but spans across both. (This doesn't explain the vertical difference well, though...)


projects/fishtornado_xr/FishTornado.h
```diff
@@ -127,8 +127,9 @@ class FishTornadoApp
     Ocean                        mOcean;
     Shark                        mShark;
     bool                         mUsePCF                   = true;
-    uint64_t                     mTotalGpuFrameTime        = 0;
-    grfx::PipelineStatistics     mPipelineStatistics       = {};
+    int                                   mViewCount                = 1;
```
chaoticbob, line 130: Odd, the format checker didn't catch the aligned assignment with the lines above and below this section.

Slumpwuffle: Fixed this, I think...

src/ppx/application.cpp
```diff
@@ -1400,6 +1400,8 @@ void Application::DrawDebugInfo(std::function<void(void)> drawAdditionalFn)
     }
     uint32_t minWidth  = std::min(kImGuiMinWidth, GetWindowWidth() / 2);
     uint32_t minHeight = std::min(kImGuiMinHeight, GetWindowHeight() / 2);
+    // This is roughly centered for XR, upper-left corner for non-XR.
+    ImGui::SetNextWindowPos({GetWindowWidth() / 4.f, GetWindowHeight() / 4.f}, 0, {0.5f, 0.5f});
```
apazylbe, line 1404: I think the pivot value (0.5, 0.5) does not make sense for non-XR, and I also personally like the initial position better now (though that's subjective). You can try opening a non-XR application and resizing the debug window, it will resize from the center and the top and left side might go out of the window range. Maybe we should be setting these for XR only?

apazylbe (changes requested):

apazylbe:
Also, would you be able to update the branch to latest to resolve conflict?

src/ppx/application.cpp
Slumpwuffle: Reworked this a good bit. Found the actual culprit so now it's doing something significantly more correct.

Slumpwuffle:
Done!

Keenuts (approved):

apazylbe (changes requested):
Based on the quick testing, the position for non-XR samples is in the middle. As I understand that's not intended?

src/ppx/application.cpp
```diff
@@ -1400,6 +1398,16 @@ void Application::DrawDebugInfo(std::function<void(void)> drawAdditionalFn)
...
     uint32_t minHeight = std::min(kImGuiMinHeight, GetWindowHeight() / 2);
+#if defined(PPX_BUILD_XR)
+    // For XR, force the diagnostic window to the center with automatic sizing for legibility and since control is limited.
+    static ImVec2 lastWindowSize(0.f, 0.f);
+    ImGui::SetNextWindowPos({(GetWindowWidth() - lastWindowSize.x) / 2, (GetWindowHeight() - lastWindowSize.y) / 2}, 0, {0.0f, 0.0f});
+    ImGui::SetNextWindowSize({0, 0});
+#else
+    // For Non-XR, spawn the window in the upper left region, but allow it to be resized and repositioned.
+    // Note that the FirstUseEver conditional means any value from previous runs (stored in imgui.ini) will be used if available.
+    ImGui::SetNextWindowPos({10.f, 10.f}, ImGuiCond_FirstUseEver, {0.0f, 0.0f});
```
apazylbe, line 1409: What is the default behaviour here if we don't set this? It looks like it is already in the upper left region without this code, do we need to set anything here?

apazylbe (approved):
Thanks!

PR #169, opened 2023-05-09
Unify XR builds.

at-ninja:
Combine the `openxr` and `quest` targets into a single `xr` target.


src/android/AndroidManifest.XR.xml
```diff
@@ -9,19 +9,35 @@
...
+
   <application
       android:allowBackup="true"
       android:icon="@android:drawable/ic_info"
       android:label="${appLabel}"
       android:theme="@style/Theme.AppCompat.NoActionBar"
       tools:targetApi="31">
+    <!-- Needed for Oculus devices -->
+      <uses-native-library
+          android:name="libopenxr_forwardloader.oculus.so"
```
Keenuts, line 26: What happens if we forget to include this on an oculus device? Do we have a clear error with some undefined symbol or something handled by the XR runtime, or is it crashing in a weird way?

Keenuts (commented):
Fine with those changes. I like the fact that we now compile all paths for XR, as long as it doesn't require additional requirements for basic OpenXR builds (seems like no).
Just a question around the error handling in case of the missing oculus lib, but overall, I'm fine with those changes! Thanks!

apazylbe (approved):
Thanks! There seems to be a merge conflict, would you be able to update to latest?

at-ninja:
> Thanks! There seems to be a merge conflict, would you be able to update to latest?

I rebased this, the merge conflict should be resolved now.

Keenuts (approved):
Thanks!

PR #171, opened 2023-05-17
benchmark: add multiple sphere models.

Keenuts:
This commits adds the option to change the default model/scene from the altimeter, to a sphere, with 6 LOD levels. From 16k faces to 16.

To implement this, we could either swap meshes, or whole scenes. The reason to swap scenes is to support complex scenes without adding complexity.
Drawback is the skybox is not configurable at the scene-level, but can be removed by adding a new scene like "sphere_0_no_skybox".

The materials are available in the GLTF files, but the current shaders implementations don't switch to emissive/unlit the the skybox.


benchmarks/graphics_pipeline/main.cpp
```diff
@@ -116,13 +142,12 @@ class ProjApp
     grfx::DescriptorSetLayoutPtr                    mSetLayout;
     std::array<grfx::ShaderModulePtr, kAvailableVsShaders.size()> mVsShaders;
     std::array<grfx::ShaderModulePtr, kAvailablePsShaders.size()> mPsShaders;
+    std::array<Scene, kAvailableScenes.size()>                    mScenes;
     PerspCamera                                     mCamera;
     float3                                          mLightPosition = float3(10, 100, 10);
     BenchmarkSettings                               mBenchmarkSettings;
+    size_t                                                        mCurrentSceneIndex;
```
apazylbe, line 149: Indentation looks off, I guess our formatting check didn't catch this

apazylbe (commented):
Overall, lgtm, but I'll let @shanminchao review

benchmarks/graphics_pipeline/main.cpp
Keenuts: Ah yes,thanks!

Keenuts: Looks like the format was doing that on purpose. Maybe a mixed tab/space thing, changed the lines above so it looks consistent.

Keenuts:
note for licensing: the skybox texture was made using krita, and the sphere blender, both allowing all usages from the assets produced.

shanminchao:
How do I get the skybox to show up? I'm assuming the skybox texture is mapped to the sphere, right? 
image

shanminchao:
Spheres 1-5 seem to have a discontinuity in their UVs at the north pole:
image

Well, I'm assuming that's UV

Is this expected?

shanminchao:
Couple of additional questions:

1. Would it be easy for others to generate additional spheres? I ask because I suspect the tessellation factors chosen here might not be what we will eventually need. We might want to adjust the tessellation factors for each sphere, but not sure to what yet.
2. Is it easy at load time to read in each of the attribute streams per sphere and split them and re-arrange them as we see fit? Our knob design calls for re-arranging the vertex streams for **all** the spheres into **one** giant vertex buffer in a custom order. I want to make sure direct access to the binary stream data is possible and simple.

Keenuts:
> 1. Would it be easy for others to re-generate the spheres and/or skybox? I ask because I suspect the tessellation factors chosen here might not be what we will eventually need. We might want to adjust the tessellation factors for each sphere, but not sure to what yet. Same goes with the resolution of the skybox texture.


The sphere is a Blender basic UV sphere. So adding a new specific model means
 - starting blender
 - loading one of those GLTF files
 - removing the existing sphere
 - adding a sphere with the desired number of subdivisions
 - exporting the new GLTF file

> 2. Is it easy at load time to read in each of the attribute streams per sphere and split them and re-arrange them as we see fit? Our knob design calls for re-arranging the vertex streams for **all** the spheres into **one** giant vertex buffer in a custom order. I want to make sure direct access to the binary stream data is possible and simple.

GLTF file format gives the start index for a vertex buffer, count, and stride. Right now, we load each one of those objects as in their own buffer, as it's simpler. But if we have the need to create our own buffers with data ordered in a specific way, that's possible yes.

Keenuts:
> I don't understand this sentence. Is this trying to say that the skybox doesn't actually show up in the rendering right now?

Yes, that's correct: as is, the sample doesn't apply a specific shader to the skybox. The default one is used:
If you want to show the skybox texture, we can add a new shader after this PR to correctly show the texture in the sphere. Right now this PR only adds the models and the ability to switch the model.

Keenuts:
> Spheres 1-5 seem to have a discontinuity in their UVs at the north pole: image
> 
> Well, I'm assuming that's UV. But the red color probably means it isn't. Maybe it's a diffuse color calculation, in which case this might mean an issue with the normals. In any case, something seems to be off here.

This is UV, on R and G channels. So yes the line is expected as we are at the end of the unwraped plane.

But looks like the modifier I used on blender was messing up some normals. Recreated the spheres to have a proper normal. The spike at the pole is expected due to the sphere topology (top vertice will have a (0, 1, 0) normal, also visible on blender).

elviscapiaq:
Yes, the basic-skybox and the other 6 spheres are fine for me. Thank you @Keenuts for adding them.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -184,7 +209,7 @@ void ProjApp::Config(ppx::ApplicationSettings& settings)
     settings.window.width               = 1920;
     settings.window.height              = 1080;
     settings.grfx.api                   = kApi;
-    settings.grfx.enableDebug           = false;
+    settings.grfx.enableDebug           = true;
```
apazylbe, line 212: Was this intentional?

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -299,10 +327,10 @@ void ProjApp::LoadMaterial(
 
             // FIXME: assuming all primitives provides POSITION, UV, NORMAL and TANGENT. Might not be the case.
             gpCreateInfo.vertexInputState.bindingCount      = 4;
-            gpCreateInfo.vertexInputState.bindings[0]       = mPrimitives[0].mesh->GetDerivedVertexBindings()[0];
-            gpCreateInfo.vertexInputState.bindings[1]       = mPrimitives[0].mesh->GetDerivedVertexBindings()[1];
-            gpCreateInfo.vertexInputState.bindings[2]       = mPrimitives[0].mesh->GetDerivedVertexBindings()[2];
-            gpCreateInfo.vertexInputState.bindings[3]       = mPrimitives[0].mesh->GetDerivedVertexBindings()[3];
+            gpCreateInfo.vertexInputState.bindings[0]       = mScenes[0].primitives[0].mesh->GetDerivedVertexBindings()[0];
```
apazylbe, line 330: Should this take into account the scene? Or the FIXME above applies where we assume all primitives in all scenes are the same

apazylbe (approved):
Seems reasonable to me! Just a couple of questions.

benchmarks/graphics_pipeline/main.cpp
Keenuts: no! fixed thanks!

benchmarks/graphics_pipeline/main.cpp
Keenuts: Yes, FIXME applies. Handling varying attributes means we would have to add logic to generate tangents, etc, and would make the application more complex.
For now, don't seem to be an issue.

PR #172, opened 2023-05-17
android: fix lifecycle handling

Keenuts:
Android activities lifecycles are not what we expect on desktop platforms:
 - we can be paused/stopped/restarted.
 - the window can be destroyed & recreated during execution.

BigWheels assumes you start once, and exit once. So the whole swapchain, initialization cannot handle android lifecycle events. Handling Android lifecycle events could either mean a deep, fine-grained integration, or what I'm implementing here.

This implementation hides the whole lifecycle concept to the application. Meaning we can have a simple, linear code when writing samples.

The downside is we cannot handle PAUSE like a real app does: our whole state is reset. I believe this downside is OK for the simplification this solution brings.

For now, I hard-assert on all messages handled in the default handler so we quickly find out new edge cases, and tried to keep the state tracking dead-simple: are we running, or waiting to be restarted.

This commit also allows resetting some global variables:
 the library lifecycle is not linked to the activity lifecycle, but to
 the class loader's. This means we can re-enter android_main multiple
 time! And globals won't be reset.
 This means we need to either stop using them, or just reinitialize
 them.

### DIFF
--- a/src/android/main.cpp
+++ b/src/android/main.cpp
@@ -85,7 +85,16 @@ int32_t defaultInputHandler(struct android_app*, AInputEvent*)
     return 0;
 }
 
-void RunAndroidEventLoop(struct android_app* pApp)
+// On Android, the app can go through multiple states, which we simplify in 3 categories:
+//  - idle state:
+//      onCreate/onStart/onPause/onStop: application should mostly be waiting.
+//  - running state:
+//      onResume: the app is in the foreground, running -> give control to the BigWheels' app.
+//  - destroyed state:
+//      onDestroy: the app is getting destroyed, we should return.
+//
+// This function will process android events and not return until the state either 'destroyed' or 'running'.
+void WaitForNonIdleState(struct android_app* pApp)
 {
     pApp->onAppCmd     = defaultCommandHandler;
     pApp->onInputEvent = defaultInputHandler;
@@ -134,7 +143,7 @@ extern "C"
                 gApplicationState = READY;
             }
             else {
-                RunAndroidEventLoop(pApp);
+                WaitForNonIdleState(pApp);
             }
         }
     }
--- a/src/android/main.cpp
+++ b/src/android/main.cpp
@@ -1,4 +1,3 @@
-#include "ppx/log.h"
 #include "ppx/profiler.h"
 
 #include <android_native_app_glue.h>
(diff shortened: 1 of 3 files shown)


src/android/main.cpp
```diff
@@ -36,20 +38,101 @@ static std::vector<std::string> get_java_args(android_app* pApp)
     return args;
 }
 
+// The Android activity can go by many more states, like PAUSED.
+// Right now, we just need to be able not to crash if we are stopped and resumed.
+enum ApplicationState
+{
+    // The JNI code is loaded and running, but the activity is not started.
+    READY,
+    // The activity is started, the Applicaiton/Window code should handle events.
```
apazylbe, line 47: typo in Application

src/android/main.cpp
```diff
@@ -36,20 +38,101 @@ static std::vector<std::string> get_java_args(android_app* pApp)
...
+
+// NOTE: JNI libraries can outlive the activity. Meaning we can re-enter the android_main
+// without re-running global constructors or reseting the BSS.
+static ApplicationState gApplicationState = READY;
+
+// When the activity not running, we have no window. To simplify BigWheels code, we
+// handle those states here, and restart the application once we are back live.
+static void defaultCommandHandler(struct android_app* app, int32_t cmd)
+{
+    PPX_ASSERT_MSG(cmd == APP_CMD_START || cmd == APP_CMD_STOP || cmd == APP_CMD_DESTROY || cmd == APP_CMD_SAVE_STATE || cmd == APP_CMD_LOST_FOCUS || cmd == APP_CMD_GAINED_FOCUS, "Handled in the default-handler a message we shouldn't. This is a bug.");
```
apazylbe, line 61: Would it help readability if we used a switch here?

apazylbe (commented):
Overall, looks good to me. I think two potential follow ups are:
- Properly handle global state. Might mean we remove use of global variables altogether since they might cause other problems on Android 
- Consider deeper integration of the lifecycle events - i.e., not resetting state when window is closed on Android. This we can just document and look at if we actually have a use case that needs this

Keenuts:
Thanks, the switch is indeed better. Also changed the event set to only forbid the windows test, as the input change can be reached through instrumentation tests.

apazylbe (approved):

src/android/main.cpp
```diff
@@ -36,20 +37,105 @@ static std::vector<std::string> get_java_args(android_app* pApp)
...
+    }
+}
+
+int32_t defaultInputHandler(struct android_app*, AInputEvent*)
+{
+    PPX_ASSERT_MSG(false, "Handled an input message without a window. This is a bug.");
+    return 0;
+}
+
+void RunAndroidEventLoop(struct android_app* pApp)
```
chaoticbob, line 88: Are there any cases where this wouldn't use the default command handler? If not, it should be renamed to disambiguate it from the "Android event loop".

Keenuts: What do you mean? Like outside of the in-app event loop?
If it was just the "AndroidEventLoop" name which was too generic?
What about "WaitForNonIdleState"? Which is what we do, as the 2 only exists are either running, or the transient exit state.

PR #177, opened 2023-05-27
Add xr-ui-resolution flag

Slumpwuffle:
This change includes the following:
1. Introduces the xr-ui-resolution command-line flag, which allows users to select a separate resolution for the UI quad when in XR mode.
2. Address an issue where the vulkan swapchain creation was ignoring the provided arguments in favor of xrComponent sizes.
3. Fixes incorrect swapchain references inside the UI component.
4. Includes the XR call result in the output of CHECK_XR_CALL.

### DIFF
--- a/include/ppx/command_line_parser.h
+++ b/include/ppx/command_line_parser.h
@@ -41,6 +41,9 @@ struct StandardOptions
     std::pair<int, int> resolution         = {-1, -1};
     int                 frame_count        = -1;
     uint32_t            stats_frame_window = 300;
+#if defined(PPX_BUILD_XR)
+    std::pair<int, int> xrUIResolution = {-1, -1};
+#endif
 
     int         screenshot_frame_number                  = -1;
     std::string screenshot_path                          = "";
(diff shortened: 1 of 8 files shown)


src/ppx/imgui_impl.cpp
```diff
@@ -187,7 +187,7 @@ Result ImGuiImplDx12::InitApiObjects(ppx::Application* pApp)
     bool result = ImGui_ImplDX12_Init(
         grfx::dx12::ToApi(pApp->GetDevice())->GetDxDevice(),
         static_cast<int>(pApp->GetNumFramesInFlight()),
-        grfx::dx::ToDxgiFormat(pApp->GetSwapchain()->GetColorFormat()),
+        grfx::dx::ToDxgiFormat(pApp->GetUISwapchain()->GetColorFormat()),
```
apazylbe, line 190: Looks like this doesn't compile, where is GetUISwapchain defined?

Slumpwuffle: Sorry about that, didn't realize it wasn't available for non-XR. Fixed.

include/ppx/command_line_parser.h
```diff
@@ -164,6 +167,26 @@ class CommandLineParser
 
 private:
     CliOptions        mOpts;
+#if defined(PPX_BUILD_XR)
```
apazylbe, line 170: I think this and some other code that uses the new flag should also check for applications settings xr.enable, otherwise if the support for XR is compiled in, it just advertises the flag but does nothing with it. 
I don't know if there is a good way to do this now, perhaps after we refactor Application it will make this cleaner.

apazylbe (approved):

include/ppx/command_line_parser.h
Slumpwuffle: Yeah, I was thinking about that, too. There really isn't a good way to propagate that information at this stage of analysis. But at least the new flag won't interfere with anything in non-XR apps.

PR #178, opened 2023-06-05
Converting benchmark graphics_pipeline parameters to knobs

angela28chen:
Converting benchmark graphics_pipeline parameters (Vertex Shader, Pixel Shader) to knobs, also removing placeholder knob --alpha_blend

Example usage:
--vs "Benchmark_VsAluBound"
--ps "Benchmark_PsMemBound"

### DIFF
--- a/benchmarks/graphics_pipeline/main.cpp
+++ b/benchmarks/graphics_pipeline/main.cpp
@@ -123,7 +123,6 @@ private:
 private:
     std::shared_ptr<KnobDropdown<std::string>> pKnobVs;
     std::shared_ptr<KnobDropdown<std::string>> pKnobPs;
-    std::shared_ptr<KnobCheckbox>              pKnobAlphaBlend; // alpha blend functionality unimplemented
 
 private:
     void LoadScene(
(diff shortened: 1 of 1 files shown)


benchmarks/graphics_pipeline/main.cpp
```diff
@@ -120,15 +114,16 @@ class ProjApp
...
     std::vector<Material>  mMaterials;
     std::vector<Primitive> mPrimitives;
     std::vector<Object>    mObjects;
     TextureCache           mTextureCache;
 
 private:
-    std::shared_ptr<KnobCheckbox> pKnobAlphaBlend; // alpha blend functionality unimplemented
+    std::shared_ptr<KnobDropdown<std::string>> pKnobVs;
+    std::shared_ptr<KnobDropdown<std::string>> pKnobPs;
+    std::shared_ptr<KnobCheckbox>              pKnobAlphaBlend; // alpha blend functionality unimplemented
```
Keenuts, line 126: Could this be removed since it's not implemented (and since we have real knobs now?)

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -196,6 +191,12 @@ void ProjApp::Config(ppx::ApplicationSettings& settings)
 
 void ProjApp::InitKnobs()
 {
+    pKnobVs = GetKnobManager().CreateKnob<ppx::KnobDropdown<std::string>>("vs", 0, kAvailableVsShaders);
```
Keenuts, line 194: This will break internal testing in an invisible manner. (Not your fault, due to the way we handle flags in BigWheels).

Could you do the following:
Add a check to crash if cl_options contains "vs-shader-index" and "ps-shader-index" saying it's has been removed, and no user should rely on those?
This way it will crash all internal CIs, and we won't miss this failure.

Keenuts (changes requested):
Thanks! LGTM, except for an issue due to the way we test internally.
Adding this crash would help make sure we don't miss it.

apazylbe (approved):

Keenuts (approved):

PR #179, opened 2023-06-06
Add runtime parameter to limit test in realtime; fix frame-count argument to use 0 as infinite.

Slumpwuffle:
This change introduces the 'run-time-ms' parameter which dictates how long to run a bigwheels application for before automatically quitting. It also cleans up a little bit around the use of timestamps and MaxFrames.

### DIFF
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -466,7 +466,8 @@ private:
 private:
     CommandLineParser               mCommandLineParser;
     StandardOptions                 mStandardOptions;
-    uint64_t                        mMaxFrame;
+    uint64_t                        mMaxFrames;
+    float                           mRunTimeSeconds;
     ApplicationSettings             mSettings = {};
     std::string                     mDecoratedApiName;
     Timer                           mTimer;
--- a/include/ppx/command_line_parser.h
+++ b/include/ppx/command_line_parser.h
@@ -40,6 +40,7 @@ struct StandardOptions
     int                 gpu_index          = -1;
     std::pair<int, int> resolution         = {-1, -1};
     int                 frame_count        = -1;
+    int                 run_time_ms        = -1;
     uint32_t            stats_frame_window = 300;
 #if defined(PPX_BUILD_XR)
     std::pair<int, int> xrUIResolution = {-1, -1};
@@ -177,6 +178,7 @@ private:
 
 --deterministic               Disable non-deterministic behaviors, like clocks.
 --frame-count <N>             Shutdown the application after successfully rendering N frames.
+--run-time-ms <N>             Shutdown the application after N milliseconds.
 --gpu <index>                 Select the gpu with the given index. To determine the set of valid indices use --list-gpus.
 --headless                    Run the sample without creating windows.
 --list-gpus                   Prints a list of the available GPUs on the current system with their index and exits (see --gpu).
@@ -196,6 +198,7 @@ private:
 
 --deterministic               Disable non-deterministic behaviors, like clocks.
 --frame-count <N>             Shutdown the application after successfully rendering N frames.
+--run-time-ms <N>             Shutdown the application after N milliseconds.
 --gpu <index>                 Select the gpu with the given index. To determine the set of valid indices use --list-gpus.
 --headless                    Run the sample without creating windows.
 --list-gpus                   Prints a list of the available GPUs on the current system with their index and exits (see --gpu).
--- a/include/ppx/command_line_parser.h
+++ b/include/ppx/command_line_parser.h
@@ -197,8 +197,8 @@ private:
 --help                        Prints this help message and exits.
 
 --deterministic               Disable non-deterministic behaviors, like clocks.
---frame-count <N>             Shutdown the application after successfully rendering N frames.
---run-time-ms <N>             Shutdown the application after N milliseconds.
+--frame-count <N>             Shutdown the application after successfully rendering N frames. Default: 0 (infinite).
+--run-time-ms <N>             Shutdown the application after N milliseconds. Default: 0 (infinite).
 --gpu <index>                 Select the gpu with the given index. To determine the set of valid indices use --list-gpus.
 --headless                    Run the sample without creating windows.
 --list-gpus                   Prints a list of the available GPUs on the current system with their index and exits (see --gpu).
--- a/include/ppx/command_line_parser.h
+++ b/include/ppx/command_line_parser.h
@@ -39,8 +39,8 @@ struct StandardOptions
     // Options
     int                 gpu_index          = -1;
     std::pair<int, int> resolution         = {-1, -1};
-    int                 frame_count        = -1;
-    int                 run_time_ms        = -1;
+    int                 frame_count        = 0;
+    int                 run_time_ms        = 0;
     uint32_t            stats_frame_window = 300;
 #if defined(PPX_BUILD_XR)
     std::pair<int, int> xrUIResolution = {-1, -1};
--- a/src/ppx/application.cpp
+++ b/src/ppx/application.cpp
@@ -22,6 +22,7 @@
 #include <unordered_map>
 #include <optional>
 #include <filesystem>
+#include <limits>
 
 namespace ppx {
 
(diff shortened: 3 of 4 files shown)


include/ppx/application.h
```diff
@@ -465,7 +465,8 @@ class Application
 private:
     CommandLineParser               mCommandLineParser;
     StandardOptions                 mStandardOptions;
-    uint64_t                        mMaxFrame;
+    uint64_t                        mMaxFrames;
+    float                           mRunDuration;
```
apazylbe, line 469: Would you be able to add the measurement unit (seconds I think?) either as a comment or in the name?

include/ppx/command_line_parser.h
```diff
@@ -40,6 +40,7 @@ struct StandardOptions
     int                 gpu_index          = -1;
     std::pair<int, int> resolution         = {-1, -1};
     int                 frame_count        = -1;
+    int                 run_time_ms        = -1;
```
apazylbe, line 43: Could you add it to the usage message below? For both xr and non-xr

include/ppx/application.h
Slumpwuffle: Sure!

include/ppx/command_line_parser.h
Slumpwuffle: Ah yep, right.

Slumpwuffle:
Ah, sorry, forgot to make the updates a separate CL :/. Hopefully it's small enough to be obvious...

Keenuts:
> This change introduces the 'run-time-ms' parameter which dictates how long to run a bigwheels application for before automatically quitting. It also cleans up a little bit around the use of timestamps and MaxFrames.

What is the motivation for this PR?
Time is quite unreliable for automated testing, meaning this seems to be for manual usage?
If it's for manual usage is it to replace a ctrl-c? (or a `./vk_01_sample & sleep 1 && pkill vk_01_sample`)

Slumpwuffle:
Hmm... is time unreliable for automated testing? My desire is to limit the test time to avoid situations where a bad system image causes framerates that are remarkably slower than before. The argument can be used in conjunction with frame count to basically be a worst-case timeout if needed.

To your point, we could have a command line option, but I much prefer a clean shutdown to guarantee (per my other PR) that metrics are written out at exit.

Keenuts:
> Hmm... is time unreliable for automated testing?

Usually yes, since even the same exact binary will have different results if using time somewhere. (That's why we added the `--deterministic` flag, which even replaces the time function with a `getFrameIndex` equivalent.
I thought the motivation was to always have something you have time to look at, even if it's running at 20000fps.


> My desire is to limit the test time to avoid situations where a bad system image causes framerates that are remarkably slower than before.

Isn't that the role of the test harness?
Using frame count means you can measure over time how long it took to run 100 frames by ex.
(same can be done the other way around, but frame-count is already available).
Then, the test harness/driver/runner can add an additional timeout, if a frame freezes by ex, to kill a test after N seconds.
Using the time-based timeout like you did means if the frame never completes (for any reason), this timeout won't kick either, since it's handled in the main thread. While a test runner could handle that case.
Am I missing something?

src/ppx/application.cpp
```diff
@@ -1007,10 +1008,15 @@ int Application::Run(int argc, char** argv)
...
-    mMaxFrame = UINT64_MAX;
+    mMaxFrames = UINT64_MAX;
     // If command line provided a maximum number of frames to draw
     if (mStandardOptions.frame_count != -1) {
-        mMaxFrame = mStandardOptions.frame_count;
+        mMaxFrames = mStandardOptions.frame_count;
+    }
+
+    mRunTimeSeconds = std::numeric_limits<float>::max();
+    if (mStandardOptions.run_time_ms > 0) {
```
Keenuts, line 1018: >= to 0 right? (otherwise 0 also means infinite)

include/ppx/command_line_parser.h
```diff
@@ -177,6 +178,7 @@ class CommandLineParser
 
 --deterministic               Disable non-deterministic behaviors, like clocks.
 --frame-count <N>             Shutdown the application after successfully rendering N frames.
+--run-time-ms <N>             Shutdown the application after N milliseconds.
```
Keenuts, line 181: Maybe also document that `-1` means no limit?

Keenuts:
> Right, but having a time cutoff for testing is pretty nominal, right?

Sure, wasn't quite convinced of the added value since we already have `frame-count` which is roughly the same assuming you don't freeze.
It's just that frame-count is more absolute than time.
With frame-count, you can easily separate the warmup frame from the rest by ex.
Time is messier.

> this is a method of guaranteeing the application can still e.g. dump data after the completion of the test.

> The situation where I expect it to hit the time limit in-app would be performance degradation

Still not convinced of the usefulness TBH. But I won't block the PR on my sole opinion, so if you want to get this merge, we can move forward.

If you use frame-count, and report runtime, you can still track performance over time. Also, I'd expect a performance benchmark to use the metrics framework, and ignore the first frames (or separate it from the actual run).
Then, from the reports, track the variation overtime, and determine if there is a regression either by the delta, or by tracking evolution over time on dashboard (for small, accumulating regressions being under the noise threshold).

The whole logic to link time to frames seems to be the job of the test runner, using the exported metrics etc.
Cutting the application with the implemented time limit means cutting at an arbitrary moment, after an arbitrary number of frames, and seems IMO a bit weird.

Once again, if you are convinced this solves your usecase, we can go ahead. Just want to make sure this is not a "redundant" option vs the `frame-count`.

Slumpwuffle:
> > Right, but having a time cutoff for testing is pretty nominal, right?
> 
> Sure, wasn't quite convinced of the added value since we already have `frame-count` which is roughly the same assuming you don't freeze. It's just that frame-count is more absolute than time. With frame-count, you can easily separate the warmup frame from the rest by ex. Time is messier.
> 
> > this is a method of guaranteeing the application can still e.g. dump data after the completion of the test.
> 
> > The situation where I expect it to hit the time limit in-app would be performance degradation
> 
> Still not convinced of the usefulness TBH. But I won't block the PR on my sole opinion, so if you want to get this merge, we can move forward.
> 
> If you use frame-count, and report runtime, you can still track performance over time. Also, I'd expect a performance benchmark to use the metrics framework, and ignore the first frames (or separate it from the actual run). Then, from the reports, track the variation overtime, and determine if there is a regression either by the delta, or by tracking evolution over time on dashboard (for small, accumulating regressions being under the noise threshold).
> 
> The whole logic to link time to frames seems to be the job of the test runner, using the exported metrics etc. Cutting the application with the implemented time limit means cutting at an arbitrary moment, after an arbitrary number of frames, and seems IMO a bit weird.
> 
> Once again, if you are convinced this solves your usecase, we can go ahead. Just want to make sure this is not a "redundant" option vs the `frame-count`.

I'm confident this won't be redundant, but I'm happy to talk through it sometime if you're still not convinced. I agree with your point about avoiding the first few frames worth of data, though I think there's a longer discussion to be had about the metrics we capture regardless.

src/ppx/application.cpp
```diff
@@ -1007,10 +1008,16 @@ int Application::Run(int argc, char** argv)
     }
 #endif
 
-    mMaxFrame = UINT64_MAX;
-    // If command line provided a maximum number of frames to draw
-    if (mStandardOptions.frame_count != -1) {
-        mMaxFrame = mStandardOptions.frame_count;
+    mMaxFrames = UINT64_MAX;
+    // If command line provided a maximum number of frames to draw.
+    if (mStandardOptions.frame_count > 0) {
```
Keenuts, line 1013: If you change frame_count behavior in that PR too you should update the title to reflect this change.

src/ppx/command_line_parser.cpp
```diff
@@ -118,6 +118,12 @@ std::optional<CommandLineParser::ParsingError> CommandLineParser::Parse(int argc
             }
             mOpts.standardOptions.frame_count = opt.GetValueOrDefault<int>(-1);
         }
+        else if (opt.GetName() == "run-time-ms") {
+            if (!opt.HasValue()) {
+                return std::string("Command-line option --run-time-ms requires a parameter");
+            }
+            mOpts.standardOptions.run_time_ms = opt.GetValueOrDefault<int>(-1);
```
Keenuts, line 125: default should be 0 then

src/ppx/command_line_parser.cpp
```diff
@@ -118,6 +118,12 @@ std::optional<CommandLineParser::ParsingError> CommandLineParser::Parse(int argc
             }
             mOpts.standardOptions.frame_count = opt.GetValueOrDefault<int>(-1);
```
Keenuts, line 119: same here, default to 0.

Keenuts (approved):
Some changes in the PR title/comment and nits in the default values, otherwise fine with this.

src/ppx/command_line_parser.cpp
Slumpwuffle: Yeah, fair. Done. Some of the other -1 values in this file probably ought to be reexamined at some point.

PR #181, opened 2023-06-06
build: bump android build ram usage limit

Keenuts:
The Android build started to fail due to out of heap memory errors.
The Github default runners have 7Gb or Ram and 14Gb of SSD.
Bumping to 4G should still allow the build to run on the default runners while fixing local builds.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -6,7 +6,7 @@
 # http://www.gradle.org/docs/current/userguide/build_environment.html
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx40986m -Dfile.encoding=UTF-8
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -6,7 +6,7 @@
 # http://www.gradle.org/docs/current/userguide/build_environment.html
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx40986m -Dfile.encoding=UTF-8
```
apazylbe, line 9: Is this supposed to be 4096?

Keenuts: Oh yes, thanks 😊 40GB would be a lot.

apazylbe (approved):

PR #182, opened 2023-06-06
ci: fix android CI

Keenuts:
The default GitHub runners have 14GB of hard drive.
The previous workflow built all the targets in one go, which reached a total of 18Gb in size.

Why is the `build/` directory that large?
- Each target requires assets, which are stored in a target specific 'assets' folder.
- Those assets are a copy of the full `assets/` and `/third_party/assets` folders.
- Size of the full assets copy: 377Mb
- Additionally, the build also creates an additional 'compressed assets' directory.
- Size of the assets + compressed: 688Mb

Then, compressed assets are stored in the APK, bringing the build size to at least 999Mb (just counting assets, no code).
Adding the code twice (once in the `intermediate/` folder, once in the APK (150M of native libs, because we support 4 archs + few MB of java).
This brings us to 1.5GB to build 1 sample.

We have 10 targets (5 targets, Debug+Release), this means 10 * 1.5Gb -> 15Gb.

This PR builds each target individually, forcing a cleanup in between.
The main drawback is we won't reuse artifacts from previous target builds, but the build directory now a more manageable size: 1.5GB.


Keenuts:
I'm trying to see if we can run those in different workflows, as running them sequentially is VERY slow. But at least this fixes the issue with the CI.

Keenuts:
Closing this since #183 will be merged.

PR #183, opened 2023-06-06
ci: fix android CI (test jobs)

Keenuts:


apazylbe:
Thanks for taking a look at this! I am also looking at reorganizing the build now, but haven't found a good solution for the space problem. I am not really a fan of having 5 checks for android build.. Do you think there is value in building all of the available mobile samples? The mobile sample selection is already somewhat arbitrary. Could we just build fishtornado and another one for now so we can have a single job that doesn't run for too long? In the meantime, I will try to look into a better long term solution. What do you think?

Keenuts:
> but haven't found a good solution for the space problem.

As an initial fix, #182 can be merged so the CI is restored. (Or this one).
Then we could work on a more long-term fix.

> I am not really a fan of having 5 checks for android build

Why that?

> Do you think there is value in building all of the available mobile samples?
Yes, more coverage. Each target we don't build is a broken target.

> so we can have a single job that doesn't run for too long?
Well, that's why I prefer this PR over #182 as it's faster. I don't quite see the issue with having multiple jobs. But maybe your rationale for not wanting 5 android checks will explain it.

apazylbe:
Well, I don't have strong reasons, I just think a single job is cleaner and more obvious than scrolling through a list of checks each potentially only adding marginal value (especially for the samples that aren't under development). We also should mark all as "required", and adjust that if we change which samples are built every time. Again no strong reasons. I am fine with merging this, and try to find a better long term solution so that we can also build more samples.

apazylbe (approved):

Keenuts:
I'll move forward with merging this one to we have a working CI again (this one is faster then #182, so might be better).

>  and more obvious than scrolling through a list of checks each potentially only adding marginal value

I'd suggest we see in action while you find a replacement? Maybe it's actually not that bad as we only look at broken ones?

PR #187, opened 2023-06-06
Add callback support for knob framework

angela28chen:
Adding functionality for application writers to add a callback when initializing the knobs. 

This callback will be triggered when:
* The value of the knob is changed with `SetValue()`, this includes when commandline flags are parsed
* The knob is changed in the UI

Intended use cases for this callback are for immediate actions, for example adjusting the value of another knob (to prevent invalid state of multiple dependent knobs).

### DIFF
--- a/include/ppx/knob.h
+++ b/include/ppx/knob.h
@@ -219,7 +219,7 @@ public:
         }
     }
 
-    void AddCallback(typename OnChangeCallback<T> callback)
+    void AddCallback(OnChangeCallback<T> callback)
     {
         mCallback = callback;
     }


hysw:
I wonder if a `valueUpdated` flag, and a `Ack()` to clear the flag is more useful than callback.
Having callbacks in the middle of the rendering loop is a bit awkward (i.e. can't re-create any resources).

Keenuts:
> I wonder if a `valueUpdated` flag, and a `Ack()` to clear the flag is more useful than callback. Having callbacks in the middle of the rendering loop is a bit awkward (i.e. can't re-create any resources).

Yes, for this reason I'd be more in favor of the `GetValue()` pattern. Since most of our usage will be to change resources/rendering options, hence we'd need to do that at a specific time.

PR #189, opened 2023-06-06
Promote ImGui debug info callback to lifecycle method.

at-ninja:
Instead of `ppx::Application::DrawDebugInfo` accepting a callback to display additional info through ImGui, the callback has been converted into a method called `DrawGui`. This makes it quick to override, and keeps the projects consistent.


apazylbe (approved):

at-ninja:
Rebased with main branch.

PR #190, opened 2023-06-07
Adding value updated/acknowledged flag to knobs

angela28chen:
Adding a value updated flag to all knobs to help with signaling when their value changes and they should be addressed in the user application code.

1. Flag will be raised whenever the value of the flag is set, or when the corresponding widget is adjusted in the UI.
2. The application can check/clear the flag with `DigestUpdate()`

Unit tests have also been updated and refactored into smaller pieces.


hysw:
Feel like we could do something similar to ImGui here:

```cpp
bool Knob::DigestUpdate() {
  if(IsValueUpdated()) {
    AckUpdatedFlag();
    return true;
  }
  return false;
}

void App::Render() {
  // ...
  if (knob->DigestUpdate()) {
    // stuff when knob updated
    UpdateBuffers(knob->GetValue());
  }
  // ...
}
```

include/ppx/knob.h
```diff
@@ -54,10 +54,12 @@ class Knob
     const std::string& GetFlagName() const { return mFlagName; }
     const std::string& GetFlagHelp() const { return mFlagDesc; }
     size_t             GetIndent() const { return mIndent; }
+    bool               IsValueUpdated() const { return mUpdatedFlag; }
 
     void SetDisplayName(const std::string& displayName) { mDisplayName = displayName; }
     void SetFlagDesc(const std::string& flagDesc) { mFlagDesc = flagDesc; }
     void SetIndent(size_t indent) { mIndent = indent; }
+    void AckUpdatedFlag() { mUpdatedFlag = false; }
```
hysw, line 62: Maybe `AckValueUpdated()`?

mdagois:
> This is an alternative to implementing callbacks for the knob framework (#187)
> 
> Callbacks
> 
> * Pro: Defined when the knob is initialized, all in one location
> * Con: Immediately triggered when a value is updated, so this could be a bad choice for kicking off operations that will take a comparatively longer time to complete (creating resources?)
> 
> Value updated flag
> 
> * Pro: More flexibility for when the new knob values are used
> * Con: If there's chained dependencies between the knobs, there may be some loops where the global state of the knob values is invalid depending on what order the knob logic is laid out in the rendering loop. This is likely way more complicated than the average use case for knobs though.
> 
> Any thoughts on which implementation would be better for current/future bigwheels applications?

I think both implementations are valid.

The callback's con can easily be avoided on the application side by just raising flags inside the callbacks and treat them later as a whole. I think the flag's con can also be entirely mitigated on the application side.

Basically, the application should be responsible for making sure the state is correct. I think it is a very reasonable expectation from the knob's framework perspective.

To sum up, I'd let you decide which implementation you like best. :)
One comment though. I would not implement the function inline inside the header. It makes it really hard to grasp the intent of a class, because the code is mingled with the interface (which is the only thing the user would care about).
You can either use a cpp file or move the function implementation outside the classes (if you want to keep a header-only approach).

include/ppx/knob.h
```diff
@@ -69,11 +69,18 @@ class Knob
 
     virtual void ResetToDefault() = 0;
 
+    // Returns true if there has been an update to the knob value and lowers mUpdatedFlag
```
Keenuts, line 72: nit: remove the mUpdatedFlag form the comment. Comment here should be about usage. mUpdatedFlag is an internal implementation detail.

Keenuts:
As @mdagois said, both could be valid, but since our use-case will mostly be changing some resources/rendering options, we would need to handle callback+flag in each app. So also in favor of this implementation.

Keenuts (approved):

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -123,6 +123,10 @@ class ProjApp
 private:
     std::shared_ptr<KnobDropdown<std::string>> pKnobVs;
     std::shared_ptr<KnobDropdown<std::string>> pKnobPs;
+    std::shared_ptr<KnobCheckbox>              pKnobPlaceholder1;
```
hysw, line 126: imo, usage examples are should be in tests,  not in actual benchmark/application.

hysw (approved):

apazylbe (approved):
Thanks Angela! Just one comment, and would you be able to update the PR description so that it can act as a commit message?

Keenuts (approved):

PR #193, opened 2023-06-14
application: disable imgui on deterministic runs.

Keenuts:
ImGUI itself is not non-deterministic. But ImGUI is used to display metrics or data which is not deterministic.
I don't see a benefit in keeping imgui in deterministic mode, so the simplest option seem to disable it.


apazylbe (approved):
What about having a flag to turn it off instead/in addition? Cause I can see that being useful too. Either way sounds reasonable to me

PR #195, opened 2023-06-15
Add SkyBox like the brackground scene

elviscapiaq:
- A checkbox knob was added to enable or not the SkyBox.


elviscapiaq:
Please take a look of my PR. I am not totally sure if these are the expect results but I will be happy to hear your feedback to iterate if it is necessary. With this PR we can visualize a Skybox as the background scene only for the Altimeter scene, since the others spheres are totally covering the background.

skybox_altimeter

skybox_altimeter_disable

PR #197, opened 2023-06-16
fs: tries to uniformize Android/Desktop FS ops.

Keenuts:
Android has 2 ways to access assets:
 - the Assets NDK API
 - normal disk read. Desktop only has the latter.

The FS API built today was preventing Android to load assets from the disk, and API diverged a bit (ex memory-mapped usage was Android only).

This commit tried to re-uniformize this. To limit the PR scope, the memory mapping is not added for desktop, but is trivial to add now.

Adding this code would also unblock Android application to load assets not from the APK folder, but from the disk. The issue I had was APK/disk duality is only valid on desktop.
I solved this by assuming relative path on Android means in-APK, and absolute means "I know what I'm doing, disk access".

This seems to work for most use-cases I found, but we might need to revisit this in the future if need be.. Maybe we should stop try to guess path between 3p and 1p assets, and maybe we could add some `apk:/some/path` or `storage:/some/other/path` indicator? but that would require apps to know in advance what we do with their assets..

### DIFF
--- a/include/ppx/fs.h
+++ b/include/ppx/fs.h
@@ -90,13 +90,13 @@ public:
 
     // Returns true if the file is readable.
     bool IsValid() const;
-    // Returns true if the file is mapped in memory. See `File::GetPointer()`.
+    // Returns true if the file is mapped in memory. See `File::GetMappedData()`.
     bool IsMapped() const;
 
     // Returns the total size in bytes of the file from the start.
     size_t GetLength() const;
     // Returns a readable pointer to a beginning of the file. Behavior undefined if `File::IsMapped()` is false.
-    const void* GetPointer() const;
+    const void* GetMappedData() const;
 
 private:
 #if !defined(PPX_ANDROID)
--- a/include/ppx/fs.h
+++ b/include/ppx/fs.h
@@ -46,27 +46,69 @@ namespace ppx::fs {
 void set_android_context(android_app* androidContext);
 #endif
 
-//! @class File
-//!
-//!
+// Abstract a static, regular file on all platforms.
+// This class doesn't handle sockets, or file which content is not constant
+// for the lifetime of this class.
 class File
 {
+    // Internal enum to know which handle is currently valid.
+    enum FileHandleType
+    {
+        // Default handle, no file associated.
+        BAD_HANDLE = 0,
+        // The file is accessible throught a stream.
+        STREAM_HANDLE = 1,
+        // The file is accessible through an Android asset handle.
+        ASSET_HANDLE = 3
+    };
+
 public:
-    bool   Open(const std::filesystem::path& path);
-    bool   IsOpen() const;
-    size_t Read(void* buf, size_t count);
-    size_t GetLength();
-    void   Close();
-#if defined(PPX_ANDROID)
-    const void* GetBuffer();
-#endif
+    File();
+    File(const File& other)  = delete;
+    File(const File&& other) = delete;
+    ~File();
+
+    // Open a file given a specific path.
+    // path: the path of the file to open.
+    //  - On desktop, loads the regular file at `path`. (memory-mapping availability is implementation defined).
+    //  - On Android, relative path are assumed to be loaded from the APK, those are memory mapped.
+    //                absolute path are loaded as regular files (mapping availability is implementation defined).
+    //
+    // - This API only supports regular files.
+    // - This API expects the file not to change size or content while this handle is open.
+    // - This class supports RAII. File will be closed on destroy.
+    bool Open(const std::filesystem::path& path);
+
+    // Read `size` bytes from the file into `buffer`.
+    // buffer: a pointer to a buffer with at least `count` writable bytes.
+    // count: the maximum number of bytes to write to `buffer`.
+    // Returns the number of bytes written to `buffer`.
+    //
+    // - The file has an internal cursor, meaning the next read will start at the end of the last read.
+    // - If the file size is larger than `count`, the read stops at `count` bytes.
+    size_t Read(void* buffer, size_t count);
+
+    // Is this file readable?
+    bool IsValid() const;
+    // Is this file mapped in memory? See `File::GetPointer()`.
+    bool IsMapped() const;
+
+    // Gets the total size in bytes of the file from the start.
+    size_t GetLength() const;
+    // Gets a readable pointer to a beginning of the file. Behavior undefined is `File::IsMapped()` is false.
+    const void* GetPointer() const;
 
 private:
-#if defined(PPX_ANDROID)
-    AAsset* mFile = nullptr;
-#else
-    std::ifstream mStream;
+#if !defined(PPX_ANDROID)
+    typedef void AAsset;
 #endif
+
+    FileHandleType mHandleType = BAD_HANDLE;
+    AAsset*        mAsset      = nullptr;
+    const void*    mBuffer     = nullptr;
+    std::ifstream  mStream;
+    size_t         mFileSize   = 0;
+    size_t         mFileOffset = 0;
 };
 
 class FileStream : public std::streambuf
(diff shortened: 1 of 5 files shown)


include/ppx/fs.h
```diff
@@ -46,27 +46,69 @@ namespace ppx::fs {
...
+// This class doesn't handle sockets, or file which content is not constant
+// for the lifetime of this class.
 class File
 {
+    // Internal enum to know which handle is currently valid.
+    enum FileHandleType
+    {
+        // Default handle, no file associated.
+        BAD_HANDLE = 0,
+        // The file is accessible throught a stream.
```
apazylbe, line 59: typo, through without a "t" at the end

include/ppx/fs.h
```diff
@@ -46,27 +46,69 @@ namespace ppx::fs {
...
+        STREAM_HANDLE = 1,
+        // The file is accessible through an Android asset handle.
+        ASSET_HANDLE = 3
```
apazylbe, line 62: Any reason this is a 3 and not a 2?

include/ppx/fs.h
```diff
@@ -46,27 +46,69 @@ namespace ppx::fs {
...
-    void   Close();
-#if defined(PPX_ANDROID)
-    const void* GetBuffer();
-#endif
+    File();
+    File(const File& other)  = delete;
+    File(const File&& other) = delete;
+    ~File();
+
+    // Open a file given a specific path.
```
apazylbe, line 71: Just for consistency can we make function comments implicitly follow "This function...", so
"Opens a file given..."

include/ppx/fs.h
```diff
@@ -46,27 +46,69 @@ namespace ppx::fs {
...
+    //  - On desktop, loads the regular file at `path`. (memory-mapping availability is implementation defined).
+    //  - On Android, relative path are assumed to be loaded from the APK, those are memory mapped.
+    //                absolute path are loaded as regular files (mapping availability is implementation defined).
+    //
+    // - This API only supports regular files.
+    // - This API expects the file not to change size or content while this handle is open.
+    // - This class supports RAII. File will be closed on destroy.
+    bool Open(const std::filesystem::path& path);
+
+    // Read `size` bytes from the file into `buffer`.
```
apazylbe, line 82: For consistency, change to Reads?

include/ppx/fs.h
```diff
@@ -46,27 +46,69 @@ namespace ppx::fs {
...
+    // buffer: a pointer to a buffer with at least `count` writable bytes.
+    // count: the maximum number of bytes to write to `buffer`.
+    // Returns the number of bytes written to `buffer`.
+    //
+    // - The file has an internal cursor, meaning the next read will start at the end of the last read.
+    // - If the file size is larger than `count`, the read stops at `count` bytes.
+    size_t Read(void* buffer, size_t count);
+
+    // Is this file readable?
```
apazylbe, line 91: "Returns true if the file is readable"?

include/ppx/fs.h
```diff
@@ -46,27 +46,69 @@ namespace ppx::fs {
...
+    bool IsValid() const;
+    // Is this file mapped in memory? See `File::GetPointer()`.
```
apazylbe, line 93: "Returns true.."?

include/ppx/fs.h
```diff
@@ -46,27 +46,69 @@ namespace ppx::fs {
...
+    bool IsMapped() const;
+
+    // Gets the total size in bytes of the file from the start.
+    size_t GetLength() const;
+    // Gets a readable pointer to a beginning of the file. Behavior undefined is `File::IsMapped()` is false.
```
apazylbe, line 98: "undefined is" -> "undefined if"

include/ppx/fs.h
```diff
@@ -46,27 +46,69 @@ namespace ppx::fs {
...
+    const void* GetPointer() const;
```
apazylbe, line 99: Personally I think GetDataIfMapped would make it more readable, but up to you

include/ppx/fs.h
```diff
@@ -78,8 +120,19 @@ class FileStream : public std::streambuf
     std::vector<char> mBuffer;
 };
 
+// Opens a regular file and returns its content if the read succeeded.
+// `path`: the path of the file to load.
+// The path is handled differently depending on the platform:
+//  - desktop: all path are treated the same.
```
apazylbe, line 126: "all path" -> "all paths"

include/ppx/fs.h
```diff
@@ -46,27 +46,69 @@ namespace ppx::fs {
...
+    File();
+    File(const File& other)  = delete;
+    File(const File&& other) = delete;
+    ~File();
+
+    // Open a file given a specific path.
+    // path: the path of the file to open.
+    //  - On desktop, loads the regular file at `path`. (memory-mapping availability is implementation defined).
+    //  - On Android, relative path are assumed to be loaded from the APK, those are memory mapped.
+    //                absolute path are loaded as regular files (mapping availability is implementation defined).
```
apazylbe, line 75: Do we need to do anything special to read those? For example add READ_EXTERNAL_STORAGE to manifest?

src/test/filesystem_test.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+constexpr std::string_view kDefaultFileContent = "some content";
+
+// Counts the number of filedescriptors open by the current process.
+static size_t getOpenFDCount()
+{
+    DIR* fdDir = opendir("/proc/self/fd");
+    assert(fdDir != nullptr);
+
+    size_t fd_count = 0;
```
apazylbe, line 40: nit: fdCount?

apazylbe (approved):
Lgtm, just some nits. Thanks for adding tests!

src/ppx/bitmap.cpp
```diff
@@ -583,30 +585,23 @@ Result Bitmap::GetFileProperties(const std::filesystem::path& path, uint32_t* pW
...
-    char* pResult = nullptr;
-#if defined(PPX_ANDROID)
     ppx::fs::File file;
     if (!file.Open(path)) {
-        return pResult;
-    }
-    if (format == Bitmap::FORMAT_RGBA_FLOAT) {
-        pResult = reinterpret_cast<char*>(stbi_loadf_from_memory(
-            reinterpret_cast<const stbi_uc*>(file.GetBuffer()), file.GetLength(), pWidth, pHeight, pChannels, desiredChannels));
+        return nullptr;
```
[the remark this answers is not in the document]
Keenuts: Yeah, I agree the assert without the path is frustrating. Adding the error bellow in the `LoadFile` function.

src/ppx/bitmap.cpp
```diff
@@ -583,30 +585,23 @@ Result Bitmap::GetFileProperties(const std::filesystem::path& path, uint32_t* pW
...
+
+    std::vector<stbi_uc> buffer(0);
+    if (!file.IsMapped()) {
+        buffer.resize(file.GetLength());
+        file.Read(buffer.data(), buffer.size());
     }
-    file.Close();
-#else
+    const stbi_uc* readPtr = file.IsMapped() ? reinterpret_cast<const stbi_uc*>(file.GetPointer()) : buffer.data();
+    ;
```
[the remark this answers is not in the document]
Keenuts: missed it, thanks!

include/ppx/fs.h
Keenuts: Forgot to clean after removing linux file-mapping form the PR. Fixed!

include/ppx/fs.h
Keenuts: Looks like this will depend on the data location?
I just tried loading the CubeTextured's texture from `/data/local/tmp/` instead of the APK, and it works with checked-in manifests.

include/ppx/fs.h
Keenuts: Renamed to `GetMappedData()`. The `IfMapped` parts seems like I'd be allowed to call this function to check if the pointer is valid. While the file-mapping is not implemented on all platforms, I'd prefer to split `IsMapped()` and `GetMappedData()` logics.
(Open to refactorize that in a single (GetMappedData()) and check for null if no platform gives us a constraint making this harder to implement.

PR #198, opened 2023-06-16
utils: add ScopedTimer class.

Keenuts:
A lot of timer patterns we have is start, do something, and stop in the same function to measure some process.

The Scopedtimer wraps this logic in a class using RAII. An alternative implementation could be to not print, but use a float& which is set to the elapsed time in the destructor. This would allow more control over the format, and print additional info not available on timer initialization.
But for now, this is a simple pattern we can start with.

### DIFF
--- a/include/ppx/timer.h
+++ b/include/ppx/timer.h
@@ -22,6 +22,7 @@
 //   Use PPX_TIMER_FORCE_MONOTONIC to force CLOCK_MONOTONIC.
 //
 
+#include <string>
 #include <cstdint>
 
 // clang-format off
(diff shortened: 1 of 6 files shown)


src/ppx/graphics_util.cpp
```diff
@@ -1189,34 +1169,28 @@ Result CreateIBLTexturesFromFile(
...
-    double fnStartTime = timer.SecondsSinceStart();
-
     // Load IBL environment map - this is stored as a bitmap on disk
     std::filesystem::path envFilePath = path.parent_path() / envFile;
     Mipmap                mipmap      = {};
-    ppxres                            = Mipmap::LoadFile(envFilePath, baseWidth, baseHeight, &mipmap, levelCount);
-    if (Failed(ppxres)) {
-        return ppxres;
+    {
+        ScopedTimer timer("Texture creation from mipmap file '" + envFilePath.string() + "'");
```
apazylbe, line 1185: I think this doesn't actually measure the time for CreateTextureFromMipMap call below? Which is different from the previous behaviour. Is this what we want?

include/ppx/timer.h
```diff
@@ -90,6 +91,21 @@ class Timer
...
 };
 
+// Helper class to display a timer result using RAII pattern.
+class ScopedTimer
+{
+    Timer             mTimer;
+    const std::string mMessage;
+
+public:
+    // Create a timer. The logged message will have the format "<message>: %f ms elapsed".
```
apazylbe, line 101: nit: Creates?

src/ppx/graphics_util.cpp
Keenuts: Yes right, fixing this!

apazylbe (approved):

PR #200, opened 2023-06-16
Refactoring commandline parsing to make use of knob framework (proof of concept)

angela28chen:
This PR is pretty messy, I just want to get some feedback on the refactoring portion of it and whether it's worth cleaning that up and implementing or not.

New features supported
* Support for JSON config file parsing
* "KnobConstants" which are basically knobs that aren't displayed in the UI and cannot be changed while application is running. This is intended for settings that are only to be specified by commandline/config file
* Boolean flags / options can now be specified in more ways
```
==============================
Options can be turned on with:
  --flag_name true
  --flag_name

And turned off with:
  --flag_name false
  --no-flag_name
==============================
```

Refactoring
* Replace `StandardOptions` with `KnobConstant`
* Changing implementation of `CliOptions` but not changing the interface, so existing applications using `ExtraOptions` are not affected. Basically no more standard/extra options, all options are just stored as strings and then flags are updated with their values.
* Minor change to knob flag text where parameters are stored separately and the KnobManager assembles the flag help message

By implementing standard options as knobs:
* Adding new flags is less error-prone
    * Flag usage message is automatically generated
    * Flag names cannot be the same as other flags specified with knobs
* `CommandlineParser::Parse()` logic can be reduced
* (Big if) If all applications move to using knobs then commandline parsing can return error and exit early when invalid flags are specified, which would make it easier to use

### DIFF
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -344,8 +344,8 @@ public:
     void Quit();
 
     std::vector<const char*> GetCommandLineArgs() const;
-    const StandardOptions    GetStandardOptions() const;
-    const CliOptions&        GetExtraOptions() const;
+    //const StandardOptions    GetStandardOptions() const;
+    const CliOptions& GetExtraOptions() const;
 
     const ApplicationSettings* GetSettings() const { return &mSettings; }
     uint32_t                   GetWindowWidth() const { return mSettings.window.width; }
@@ -487,7 +487,6 @@ private:
 
 private:
     CommandLineParser               mCommandLineParser;
-    StandardOptions                 mStandardOptions;
     uint64_t                        mMaxFrames;
     float                           mRunTimeSeconds;
     ApplicationSettings             mSettings = {};
@@ -514,6 +513,29 @@ private:
     float             mAverageFrameTime  = 0;
     double            mFirstFrameTime    = 0;
     std::deque<float> mFrameTimesMs;
+
+    // Standard Options
+    struct
+    {
+        // Flags
+        std::shared_ptr<KnobConstant<bool>> pListGpus;
+        std::shared_ptr<KnobConstant<bool>> pUseSoftwareRenderer;
+        std::shared_ptr<KnobConstant<bool>> pHeadless;
+        std::shared_ptr<KnobConstant<bool>> pDeterministic;
+
+        // Options
+        std::shared_ptr<KnobConstant<int>>          pGpuIndex;
+        std::shared_ptr<KnobConstantPair<int, int>> pResolution;
+#if defined(PPX_BUILD_XR)
+        std::shared_ptr<KnobConstantPair<int, int>> pXrUiResolution;
+#endif
+        std::shared_ptr<KnobConstant<int>>         pFrameCount;
+        std::shared_ptr<KnobConstant<int>>         pRunTimeMs;
+        std::shared_ptr<KnobConstant<int>>         pStatsFrameWindow;
+        std::shared_ptr<KnobConstant<int>>         pScreenshotFrameNumber;
+        std::shared_ptr<KnobConstant<std::string>> pScreenshotPath;
+        std::shared_ptr<KnobConstant<std::string>> pJsonConfigPath;
+    } mKnobOptions;
 
     // Metrics
     struct
--- a/include/ppx/knob.h
+++ b/include/ppx/knob.h
@@ -54,15 +54,14 @@ public:
     const std::string& GetFlagName() const { return mFlagName; }
     const std::string& GetFlagHelp() const { return mFlagDesc; }
     size_t             GetIndent() const { return mIndent; }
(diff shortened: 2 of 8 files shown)


include/ppx/knob.h
```diff
@@ -361,6 +337,131 @@ class KnobDropdown final
...
+        SetValue(defaultValue);
+    }
+
+    void Draw() override {}
+
+    void ResetToDefault() override {}
+
+    void UpdateFromFlags(const CliOptions& opts) override
+    {
+        SetValue(opts.GetIntPairValueOrDefault(GetFlagName(), mValue));
```
apazylbe, line 435: Does this mean the type has to be int?

include/ppx/knob.h
```diff
@@ -361,6 +337,131 @@ class KnobDropdown final
...
+public:
+    KnobConstant(const std::string& flagName, T defaultValue)
+        : Knob(flagName)
+    {
+        static_assert(std::is_same_v<T, bool> || std::is_same_v<T, std::string> || std::is_same_v<T, int> || std::is_same_v<T, float>, "KnobConstant must be of type bool, std::string, int, or float");
+
+        SetValue(defaultValue);
+    }
+
+    KnobConstant(const std::string& flagName, T defaultValue, T minValue, T maxValue)
```
apazylbe, line 355: I think this only makes sense for numeric types?

src/ppx/knob.cpp
```diff
@@ -69,25 +69,57 @@ void KnobManager::DrawAllKnobs(bool inExistingWindow)
...
+    for (const auto& knobPtr : mKnobConstants) {
+        size_t flagLength = knobPtr->GetFlagName().size() + knobPtr->GetHelpParams().size() + 2;
+        if (flagLength > maxFlagLength) {
+            maxFlagLength = flagLength;
+        }
+    }
+    for (const auto& knobPtr : mKnobConstants) {
+        std::string flagText = "--" + knobPtr->GetFlagName() + knobPtr->GetHelpParams();
+        size_t      spaces   = maxFlagLength - flagText.size() + 2;
+        for (size_t i = 0; i < spaces; i++) {
```
apazylbe, line 85: You can use `std::string(spaces, ' ')` instead of the loop?

apazylbe (commented):
At a high level, I am good with changes, and I think they can provide value. Personally, I like keeping some of the options related code in the command line parser, so we don't have to bloat the application.

include/ppx/application.h
```diff
@@ -344,8 +344,8 @@ class Application
     void Quit();
 
     std::vector<const char*> GetCommandLineArgs() const;
-    const StandardOptions    GetStandardOptions() const;
-    const CliOptions&        GetExtraOptions() const;
+    //const StandardOptions    GetStandardOptions() const;
```
MichaelAnttila, line 347: Was this line meant to be deleted instead of just commented out?

src/ppx/application.cpp
```diff
@@ -650,6 +650,65 @@ void Application::DestroyPlatformWindow()
...
+#endif
+    mKnobOptions.pHeadless->SetFlagDesc("Run the sample without creating windows.");
+
+    mKnobOptions.pDeterministic = GetKnobManager().CreateKnobConstant<bool>("deterministic", false);
+    mKnobOptions.pDeterministic->SetFlagDesc("Disable non-deterministic behaviors, like clocks.");
+
+    mKnobOptions.pGpuIndex = GetKnobManager().CreateKnobConstant<int>("gpu", -1, -1, INT_MAX);
+    mKnobOptions.pGpuIndex->SetFlagDesc("Select the gpu with the given index. To determine the set of valid indices use --list-gpus.");
+
+    mKnobOptions.pResolution = GetKnobManager().CreateKnobConstantPair<int, int>("resolution", std::make_pair(0, 0));
```
hysw, line 674: e.g.
```
GetKnobManager().CreateKnob<KnobConstant<std::pair<int, int>>>("resolution", std::make_pair(0,0));
```

include/ppx/application.h
```diff
@@ -515,6 +514,29 @@ class Application
     double            mFirstFrameTime    = 0;
     std::deque<float> mFrameTimesMs;
 
+    // Standard Options
+    struct
+    {
+        // Flags
+        std::shared_ptr<KnobConstant<bool>> pListGpus;
```
Slumpwuffle, line 521: In general, shared_ptr should be avoided. Is there a reason it's being used here rather than unique_ptr? Or even just a KnobConstant?

hysw: It should just be a "raw" pointer here with KnobManager keep ownership of the knobs. The usage was from #155 since proper memory management is tricky for people who are new to C++.

hysw: `shared_ptr` is thread safe and uses extra memory for reference counting, which makes it quite expensive.

The vector should be holding `unique_ptr`, and a reference of the object should be returned.

PR #201, opened 2023-06-16
Build Android APK for arm64-v8a only

apazylbe:
I don't think there currently is a good reason to include all ABIs. Limiting to one reduces APK size and build times, which is better for development and CI checks.


Keenuts:
Don't see issues with that. @at-ninja @mdagois , any ABI you require which is not arm64-v8?

Keenuts (approved):

at-ninja:
> Don't see issues with that. @at-ninja @mdagois , any ABI you require which is not arm64-v8?

No issues here either!

PR #203, opened 2023-06-20
assets,28_gltf: add skybox to the altimeter

Keenuts:
This commit fixes the skybox material to properly use the "unlit" GLTF extension (blender doesn't seem to export it?).

To showcase how such material could be rendered, this sample also adds the Unlit basic material, compatible with the fake-PBR for the 28_gltf sample.

28_gltf sample rendering now: 
skybox


projects/28_gltf/main.cpp
```diff
@@ -259,7 +259,8 @@ void ProjApp::LoadMaterial(
 
     grfx::GraphicsPipelineCreateInfo2 gpCreateInfo = {};
     gpCreateInfo.VS                                = {mVertexShader.Get(), "vsmain"};
-    gpCreateInfo.PS                                = {mPixelShader.Get(), "psmain"};
+    ppx::grfx::ShaderModule* psModule              = material.unlit ? mUnlitPixelShader.Get() : mPbrPixelShader.Get();
```
apazylbe, line 262: Just to clarify, right now it wouldn't use PBR path, since the scene is hardcoded and the material is unlit in it?

apazylbe (approved):

projects/28_gltf/main.cpp
Keenuts: It would use both actually :)
The `material.unlit` depends on the GLTF file material definition. Only the skybox has the "unlit" marker, while the altimeter is classic PBR.
Each object is assigned 1 material, meaning we have 1 unlit material (skybox) and 1 'PBR' material ('PBR' just because this one is an ugly 'pbr' shader, not the real stuff)

PR #204, opened 2023-06-20
Minor improvements for knob framework

angela28chen:
* Replace `GetFlagHelpText()` with storing help parameters separately and assembling in `KnobManager::GetUsageMsg()`
* Move `KnobCheckbox` member functions to `knob.cpp`
* When listing `KnobDropdown` choices in usage message, only add quotation marks if there are spaces
* Made KnobManager a friend class of Knob and removed some of the getters/setters, preferring to have protected/private members where possible
* Renamed `SetFlagDesc` -> `SetFlagDescription`

### DIFF
--- a/include/ppx/knob.h
+++ b/include/ppx/knob.h
@@ -61,12 +61,14 @@ public:
     // Returns true if there has been an update to the knob value
     bool DigestUpdate();
 
+    virtual void ResetToDefault() = 0;
+
 protected:
     void RaiseUpdatedFlag() { mUpdatedFlag = true; }
 
 private:
-    virtual void Draw()           = 0;
-    virtual void ResetToDefault() = 0;
+    // Only called from KnobManager
+    virtual void Draw() = 0;
 
     // Updates knob value from commandline flag
     virtual void UpdateFromFlags(const CliOptions& opts) = 0;
@@ -126,7 +128,7 @@ public:
     {
         PPX_ASSERT_MSG(minValue < maxValue, "invalid range to initialize slider");
         PPX_ASSERT_MSG(minValue <= defaultValue && defaultValue <= maxValue, "defaultValue is out of range");
-        SetFlagParameters(" <" + std::to_string(mMinValue) + "~" + std::to_string(mMaxValue) + ">");
+        SetFlagParameters("<" + std::to_string(mMinValue) + "~" + std::to_string(mMaxValue) + ">");
         RaiseUpdatedFlag();
     }
 
@@ -215,7 +217,7 @@ public:
         if (!choiceStr.empty()) {
             choiceStr.pop_back();
         }
-        choiceStr = " <" + choiceStr + ">";
+        choiceStr = "<" + choiceStr + ">";
         SetFlagParameters(choiceStr);
         RaiseUpdatedFlag();
     }
--- a/include/ppx/knob.h
+++ b/include/ppx/knob.h
@@ -54,15 +54,14 @@ public:
     const std::string& GetFlagName() const { return mFlagName; }
     const std::string& GetFlagHelp() const { return mFlagDesc; }
     size_t             GetIndent() const { return mIndent; }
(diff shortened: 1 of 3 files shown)


include/ppx/knob.h
```diff
@@ -54,16 +54,15 @@ class Knob
     const std::string& GetFlagName() const { return mFlagName; }
     const std::string& GetFlagHelp() const { return mFlagDesc; }
     size_t             GetIndent() const { return mIndent; }
+    const std::string& GetHelpParams() const { return mHelpParams; }
```
apazylbe, line 57: `GetHelpParams` does not sound very descriptive. It would help to have a more descriptive name or at least a comment. Perhaps we could also group the three flag functions together and give them related names. `GetFlagHelp` above is also a bit confusing, and potentially misleading as well (since there was `GetFlagHelpText` below.
Perhaps `GetFlagName`, `GetFlagParameters`/`GetFlagValues` and `GetFlagDescription`?

include/ppx/knob.h
```diff
@@ -54,16 +54,15 @@ class Knob
...
 
     void SetDisplayName(const std::string& displayName) { mDisplayName = displayName; }
     void SetFlagDesc(const std::string& flagDesc) { mFlagDesc = flagDesc; }
     void SetIndent(size_t indent) { mIndent = indent; }
+    void SetHelpParams(const std::string& helpParams) { mHelpParams = helpParams; }
```
apazylbe, line 62: Looking at just changes here and not how this will be used later, how do you feel about overriding `GetHelpParams` to build the string instead of having setter/getter. Perhaps later usage of SetHelpParams changes things though

include/ppx/knob.h
```diff
@@ -54,16 +54,15 @@ class Knob
...
 
     virtual void Draw() = 0;
 
-    // Returns the flag description used in usage message
-    virtual std::string GetFlagHelpText() const = 0;
```
apazylbe, line 65: Are `GetFlagName`, `GetFlagHelp` and `GetHelpParams` only used by the KnobManager to build help text?
I  think you moved this function out of Knob because you want to align the description?
You could consider making the KnobManager a friend and keeping some of this interface private, if it doesn't need to be public.

src/test/knob_test.cpp
```diff
@@ -371,15 +365,15 @@ TEST_F(KnobManagerTestFixture, KnobManager_GetUsageMsg)
...
     std::shared_ptr<KnobDropdown<std::string>> strKnobPtr1(km.CreateKnob<KnobDropdown<std::string>>("flag_name4", 1, choices1.cbegin(), choices1.cend()));
 
     std::string usageMsg = R"(
-Application-specific flags
+Application-Specific Flags:
 --flag_name1 <true|false>
 --flag_name2 <true|false>
 --flag_name3 <0~10>
---flag_name4 <"c1"|"c2"|"c3">
+--flag_name4 <c1|c2|"c3 and more">
```
apazylbe, line 376: Curious, why this change?

include/ppx/knob.h
```diff
@@ -243,6 +202,21 @@ class KnobDropdown final
         : Knob(flagName), mIndex(defaultIndex), mDefaultIndex(defaultIndex), mChoices(choicesBegin, choicesEnd)
     {
         PPX_ASSERT_MSG(defaultIndex < mChoices.size(), "defaultIndex is out of range");
+        std::string choiceStr = "";
+        for (const auto& choice : mChoices) {
+            bool hasSpace = choice.find_first_of("\t ") != std::string::npos;
+            if (hasSpace) {
+                choiceStr += R"(")" + choice + R"("|)";
```
Keenuts, line 209: nit: why use the more complex `R"( )"` notation vs `"\""` ?

Keenuts (approved):
Ok with this, @apazylbe has covered remaining issues, delegating approval 😊

apazylbe (changes requested):

apazylbe (approved):

PR #205, opened 2023-06-21
Create KnobFlag

angela28chen:
Create knobs that can have their value set by the commandline flags but not through the UI or by the application during rendering.

* Add Knob member for visibility and hide constant knobs by default.
* Minor fix, make `KnobDropdown::Draw` private

Will be implemented in follow-up PRs:
* Int Pair KnobFlag which will be used for resolution flags, will be supported and unit tested once they are supported by the commandline parser.


include/ppx/knob.h
```diff
@@ -357,6 +357,95 @@ class KnobDropdown final
...
+// They will not be displayed in the UI and cannot have their values changed by the application
+template <typename T>
+class KnobConstant final
+    : public Knob
+{
+public:
+    KnobConstant(const std::string& flagName, T defaultValue)
+        : Knob(flagName)
+    {
+        static_assert(std::is_same_v<T, bool> || std::is_same_v<T, std::string> || std::is_same_v<T, int> || std::is_same_v<T, float> || std::is_same_v<T, std::pair<int, int>>, "KnobConstant must be of type bool, std::string, int, float, or paired int");
```
hysw, line 370: Usually it's better to assert on the property of a type, not what type it is. i.e. concept, well if the clang version in NDK is newer we could actually use proper language feature instead of static assert

include/ppx/knob.h
```diff
@@ -369,6 +458,9 @@ class KnobManager
     // Knobs are added on creation and never removed
     std::vector<std::shared_ptr<Knob>> mKnobs;
 
+    // Knobs which are not dislayed and that cannot have their values changed
```
hysw, line 461: Have a IsVisible function on Knob might be a better idea.

include/ppx/knob.h
```diff
@@ -314,6 +316,67 @@ class KnobDropdown final
...
+        });
+
+        SetValue(defaultValue);
+    }
+
+    void Draw() override {} //TODO(chenangela: normally turned off but if turned on will just draw the values of const knobs)
+
+    T    GetValue() const { return mValue; }
+    void ResetToDefault() override {} // KnobConstant is always the "default" value
+    void SetIsValidFunc(std::function<bool(T)> isValidFunc) { mIsValidFunc = isValidFunc; }
```
hysw, line 353: personal opinion: it's usually called a "validator", "IsValidFunc" sounds a bit weird to me.

include/ppx/knob.h
```diff
@@ -314,6 +316,67 @@ class KnobDropdown final
...
+    {
+        if (!mIsValidFunc) {
+            return true;
+        }
+        return mIsValidFunc(val);
+    }
+
+    void SetValue(T newValue)
+    {
+        PPX_ASSERT_MSG(IsValidValue(newValue), "invalid default value");
```
hysw, line 371: default?

include/ppx/knob.h
```diff
@@ -314,6 +316,67 @@ class KnobDropdown final
...
+            if (newValue < minValue || newValue > maxValue) {
+                return false;
+            }
+            return true;
+        });
+
+        SetValue(defaultValue);
+    }
+
+    void Draw() override {} //TODO(chenangela: normally turned off but if turned on will just draw the values of const knobs)
```
hysw, line 349: fyi: `void Draw() final` also works here, although given your class is marked as final nothing else would override this function.

src/ppx/knob.cpp
```diff
@@ -88,7 +88,10 @@ void KnobManager::DrawAllKnobs(bool inExistingWindow)
     }
 
     for (const auto& knobPtr : mKnobs) {
-        for (size_t i = 0; i < knobPtr->mIndent; i++) {
+        if (!knobPtr->mVisible) {
+            continue;
+        }
+        for (size_t i = 0; i < knobPtr->mVisible; i++) {
```
hysw: typo? Do you mean mIndent here?

hysw (commented):
LGTM, except one typo

hysw (approved):

include/ppx/knob.h
```diff
@@ -46,8 +46,8 @@ namespace ppx {
 class Knob
 {
 public:
-    Knob(const std::string& flagName)
-        : mFlagName(flagName), mDisplayName(flagName), mIndent(0), mUpdatedFlag(false) {}
+    Knob(const std::string& flagName, bool visible)
```
Keenuts, line 49: nit: set the default value of visible to false?

include/ppx/knob.h
```diff
@@ -124,7 +126,7 @@ class KnobSlider final
     static_assert(std::is_same_v<T, int>, "KnobSlider must be created with type: int");
 
     KnobSlider(const std::string& flagName, T defaultValue, T minValue, T maxValue)
-        : Knob(flagName), mValue(defaultValue), mDefaultValue(defaultValue), mMinValue(minValue), mMaxValue(maxValue)
+        : Knob(flagName, true), mValue(defaultValue), mDefaultValue(defaultValue), mMinValue(minValue), mMaxValue(maxValue)
```
Keenuts, line 129: I'd expose the visible boolean in the constructor params (with default value to true)

include/ppx/knob.h
```diff
@@ -314,6 +316,67 @@ class KnobDropdown final
...
+    : public Knob
+{
+public:
+    KnobConstant(const std::string& flagName, T defaultValue)
+        : Knob(flagName, false)
+    {
+        SetValue(defaultValue);
+    }
+
+    KnobConstant(const std::string& flagName, T defaultValue, T minValue, T maxValue)
```
Keenuts, line 332: I feel like having the validator function is a good idea (since those would replace flags AFAIK).
But seems like the minValue/maxValue constraints this constant to be numerical, vs being some string value by ex.
Do you think it would be useful to have the generic KnobConstant() take a unary function for validator, and maybe have a template specialization for numerical values with the minValue/maxValue constructor you have here?

src/ppx/knob.cpp
```diff
@@ -36,7 +36,7 @@ bool Knob::DigestUpdate()
 // -------------------------------------------------------------------------------------------------
 
 KnobCheckbox::KnobCheckbox(const std::string& flagName, bool defaultValue)
-    : Knob(flagName), mDefaultValue(defaultValue), mValue(defaultValue)
+    : Knob(flagName, true), mDefaultValue(defaultValue), mValue(defaultValue)
```
Keenuts, line 39: same here, I'd expose the visible bit in a param with default=true

Keenuts (commented):
Thanks for this!
Some good ideas, we might have something even better by:
 - exposing the visible bit on all knobs constructors, and have it default to `true` for non-constant knobs.
  - maybe allow the validator function to be passed instead of min/max to handle more complex values than ranges/numeric values.

include/ppx/knob.h
Keenuts: Fair point, fine with that.

include/ppx/knob.h
```diff
@@ -314,6 +316,67 @@ class KnobDropdown final
...
+            if (newValue < minValue || newValue > maxValue) {
+                return false;
+            }
+            return true;
+        });
+
+        SetValue(defaultValue);
+    }
+
+    void Draw() override {} // KnobConstant is invisible in the UI
```
apazylbe, line 349: Should we add an assert in the functions that shouldn't be called? like `Draw`, `ResetToDefault` or is it valid to call them and expect nothing happens?

include/ppx/knob.h
```diff
@@ -124,7 +126,7 @@ class KnobSlider final
     static_assert(std::is_same_v<T, int>, "KnobSlider must be created with type: int");
 
     KnobSlider(const std::string& flagName, T defaultValue, T minValue, T maxValue)
```
apazylbe, line 128: Not directly relevant to this PR, but what was the rational for naming the knob classes with the UI name, instead of the value they contain (KnobBoolean, KnobRange, KnobOptions)? Do you there is a use case for them to be drawn differently? We could always add  a parameter used in the `Draw` function if that is the case. 
Just thought of this, cause now the knobs can be invisible too

include/ppx/knob.h
```diff
@@ -314,6 +316,67 @@ class KnobDropdown final
     std::vector<T> mChoices;
 };
 
+// KnobConstant is intended for parameters that cannot be adjusted when the application is run
+// They will not be displayed in the UI and cannot have their values changed by the application
+template <typename T>
+class KnobConstant final
```
apazylbe, line 322: Completely up to you, but how about naming these KnobFlag? That way it is clear what they are used for, unless you think there could be another use case for these.

include/ppx/knob.h
apazylbe, line 332: Just wondering do we need this constructor, or could we just use the constructor above and add a validator function?

include/ppx/knob.h
```diff
@@ -314,6 +317,75 @@ class KnobDropdown final
     std::vector<T> mChoices;
 };
 
+// KnobFlag is intended for parameters that cannot be adjusted when the application is run
+// They will not be displayed in the UI and cannot have their values changed by the application
```
apazylbe, line 321: It is currently possible to display those in UI with SetVisible(true). I think that's fine and maybe could have uses, but if we keep that behaviour, we should not call out that they will not be displayed in the UI, perhaps only that they are hidden by default. Alternatively we don't allow them to be visible and leave this comment as is.

include/ppx/knob.h
```diff
@@ -314,6 +317,75 @@ class KnobDropdown final
...
+    }
+
+    T GetValue() const { return mValue; }
+
+    void SetValidator(std::function<bool(T)> validatorFunc) { mValidatorFunc = validatorFunc; }
+
+private:
+    void Draw() override
+    {
+        std::stringstream ss;
```
apazylbe, line 358: Related to the other comment, but this could be used by knob framework users by setting SetVisible(true). If that's intended we can leave it, if this is just for debugging while developing knobs perhaps we should remove it.

apazylbe (approved):
Just one comment, otherwise LGTM.

I am wondering if KnobFlag could have been handled by other Knob classes with a setting that sets them to be only editable through flags. Especially if they can be drawn as read only, then the read only property is the only difference with other knobs. Anyway no action for here, just maybe a potential idea if the use cases evolve further where for example we want to draw a KnobFlag<bool> as a read only checkbox.

Keenuts (approved):

PR #207, opened 2023-06-22
Cleanup Metrics - Pull Report internal to Manager and use unique_ptrs

Slumpwuffle:
This set of changes includes two key pieces:
1. The report object is now internal to the manager, as the lifetime of strings used by the json object cannot be guaranteed outside of the scope of the manager regardless. This cleans up the application-side piece of the report writing.
2. Unique_ptrs are now used for the metrics. This helps clean up several pieces of the metrics code, especially around destructors and constructors.

### DIFF
--- a/include/ppx/metrics.h
+++ b/include/ppx/metrics.h
@@ -18,6 +18,7 @@
 #include "nlohmann/json.hpp"
 #include "ppx/config.h"
 
+#include <filesystem>
 #include <limits>
 #include <string>
 #include <unordered_map>
@@ -52,26 +53,6 @@ struct MetricMetadata
     std::string          unit;
     MetricInterpretation interpretation = MetricInterpretation::NONE;
     Range                expectedRange;
-};
-
-////////////////////////////////////////////////////////////////////////////////
-
-// A report contains runs and metrics information.
-// It is meant to be saved to disk.
-class Report final
-{
-    friend class Manager;
-
-public:
-    static constexpr const char* kFileExtension = ".json";
-
-public:
-    Report();
-    ~Report();
-    void WriteToFile(const char* pFilepath) const;
-
-private:
-    nlohmann::json mContent;
 };
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -239,11 +220,37 @@ public:
 
     Run* AddRun(const char* pName);
 
-    // Exports all the runs and metrics information into a report.
-    Report Export(const char* pName) const;
+    // Exports all the runs and metrics information into a report to disk.
+    void ExportToDisk(const std::string& baseReportName) const;
 
 private:
     METRICS_NO_COPY(Manager)
+
+    // A report contains runs and metrics information meant to be saved to disk.
+    // Because the json object at its core relies on strings-as-pointers, the
+    // lifecycle of a report is tied to the lifecycle of the runs and metrics
+    // owned by the Manager. But this is opaque. To protect misuse, the class
+    // is a private member.
+    class Report final
+    {
+    public:
+        static constexpr const char* kFileExtension = ".json";
+
+    public:
+        // Copy constructor for content.
+        Report(const nlohmann::json& content, const std::string& baseReportName);
+        // Move constructor for content.
+        Report(nlohmann::json&& content, const std::string& baseReportName);
+
+        void WriteToDisk(bool overwriteExisting = false) const;
+
+    private:
+        void SetReportName(const std::string& baseReportName);
+
+        nlohmann::json        mContent;
+        std::string           mReportName;
+        std::filesystem::path mFilePath;
+    };
 
 private:
     std::unordered_map<std::string, Run*> mRuns;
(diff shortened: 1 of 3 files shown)


include/ppx/metrics.h
```diff
@@ -195,58 +177,77 @@ class Run final
...
 public:
     template <typename T>
-    T* AddMetric(MetricMetadata metadata);
+    T* AddMetric(MetricMetadata metadata)
+    {
+        PPX_ASSERT_MSG(!metadata.name.empty(), "The metric name must not be empty");
+        PPX_ASSERT_MSG(!HasMetric(metadata.name), "Metrics must have unique names (duplicate name detected)");
+
+        auto* pMetric = new T(metadata);
+        auto  metric  = std::unique_ptr<T>(pMetric);
```
Keenuts, line 186: The goal of an `unique_ptr` is to enforce this pointer is unique. Returning a raw pointer after sound like adding constraints internally while keeping the unsafe aspect.

Keenuts: I'd be fine to move raw pointers to shared_ptr (same as the knob framwork), but not this half-mesure.

include/ppx/metrics.h
```diff
@@ -141,7 +123,8 @@ class MetricGauge final
     };
 
 private:
-    ~MetricGauge();
+    MetricGauge(const MetricMetadata& metadata)
+        : mMetadata(metadata), mAccumulatedValue(0) {}
```
Keenuts, line 127: isn't the mAccumulatedValue redundant with the other default value below?

src/ppx/metrics.cpp
```diff
@@ -215,104 +175,112 @@ uint64_t MetricCounter::Get() const
...
-    for (auto [name, pRun] : mRuns) {
-        content["runs"] += pRun->Export();
+    if (!overwriteExisting && std::filesystem::exists(mFilePath)) {
+        PPX_LOG_ERROR("Metrics report file cannot be written to disk. Path [" << mFilePath << "] already exists.");
+        return;
     }
-    return report;
+
+    std::filesystem::create_directories(mFilePath.parent_path());
+    std::ofstream outputFile(mFilePath.c_str(), std::ofstream::out);
```
Keenuts, line 266: should you check if the file is good before writing?
I'm fine returning void and assert, as it's the standard behavior for BigWheels when we fail, but we might need to check errors.

Keenuts (commented):
some comments.
Btw, I see 2 PRs here:
 - changes around export/report lifecycle.
 - change in API around pointer usage.

include/ppx/metrics.h
Slumpwuffle: Unique_ptr is more about ownership than it is about true uniqueness. Returning a raw pointer like this generally indicates that the caller does not OWN the memory in question, and that memory management is handled elsewhere; therein lies the uniqueness. Shared ptr has some very gnarly consequences and is dangerous for a variety of reasons; it should be avoided as much as possible. I will reach out to you specifically about this, because it's surprising that you would recommend it, and it seems like there's some misunderstanding of the value of unique_ptr.

src/ppx/metrics.cpp
Slumpwuffle: Oh, yeah that's probably a good call. Note that this was a copy/paste and the previous code did not check.

include/ppx/metrics.h
Slumpwuffle: yes! good catch, thanks.

include/ppx/metrics.h
Keenuts: Yeah, you are right. It's just emplaced in owned storage, and the manager lifetime is currently bounded to the external references (`mMetrics` in `application.h`).

I was worried about dangling pointers, as yes, `shared_ptr`/`weak_ptr` does means atomicity overhead, but:
 - is this really a hot path? Doesn't seem like the overhead would be significant vs the compute/graphics load.
 - it allows us to actually catch dangling pointers. Which seems to be the main benefit to have an easy-to-use/safe API for users. (We don't have `raw_ptr` construct like in chrome, hence the suggestion to use shared+weak and not export the raw pointer)
 (We can talk about than in half an hour)

Slumpwuffle: Per discussion - we think the current solution is likely inadequate depending on the situation where this gets used. In particular, we have some use cases where this pointer might reach a bit beyond the application itself, and that leads to difficulty where you can't expire the pointer when you need to (without some separate signalling, etc.)

For now this is still an improvement over what was previously here, so I'll put in a few TODOs mentioning the situation, then clean it up as our use case becomes more mature.

I suspect long-term, we may want to avoid pointers altogether. The MANAGER is guaranteed to have the same lifetime as the application, but everything else has limits. Something like a LUT or a map with function calls may be more valuable, but all of this hinges on whether the access to these components remains single-threaded.

Keenuts (approved):

Keenuts:
@apazylbe : sorry! I clicked merge but you didn't had a chance to review! Let me know if you want to revert. (sorry!)

PR #211, opened 2023-06-23
Fluid simulation - First fully functional version

dnovillo:
This is the first fully functional version of the fluid simulator I started writing several months ago.

Many thanks to @chaoticbob and @Keenuts for their guidance during this implementation. I've learned quite a bit!

This PR is quite big, so I split it in 7 commits.  Basically:

- **Initial version of the simulation loop**: This completed the code conversion from the original WebGL implementation. This is followed by 4 commits that fixed various bugs.
- **Introduce repeat samplers**: This added a clamp and a repeat sampler. The repeat sampler is using when computing the final image to present. Clamp sampler is used by all the compute passes.
- **Simplify dispatch record generation - Fix DirectX12 failure** and **Add debugging code to draw all computed textures** are various fixes and debug code that I added when trying to figure out the bug that was producing [odd output](https://drive.google.com/file/d/15vt_wyCk3WieDu-bvh4be2mXODEUS_y6/view?usp=sharing).
- **Fix texture formats to use floating point instead of ints** fixes the odd output. @Keenuts observed that several of the compute textures in the code were using the display color format (uint8), instead of floating point values.  With that fix, the simulation produced a single splash that looked [more accurate](https://drive.google.com/file/d/1_myYIheiSPJKN1HORNZNTQPRExm6kuHw/view?usp=sharing).
- **Add command line flags and simplify some logic**: This moves the configuration data into a structure that can be filled-in via command-line arguments. I hope to hook this to the knobs framework once @angela28chen makes it available.
- **Add documentation on the simulation parameters**: Documents all the parameters that can be altered to produce different simulation output.

The default execution of the simulator should produce an output [similar to this](https://drive.google.com/file/d/1Tnk43OoTKOQE46otmDh9PAAAARkgoaxn/view?usp=sharing).


chaoticbob (approved):
LGTM

Looks good for a first version. There are some elements that should change to turn the sim into a reusable component.  But in follow up PRs though.

projects/fluid_simulation/sim.cpp
```diff
@@ -415,8 +421,128 @@ void FluidSimulation::DrawDisplay()
...
+    // Queue up some splats at random. But limit the amount of outstanding splats so it doesn't get too busy.
+    if (Random().Float() <= GetConfig().splatFrequency) {
+        MultipleSplats(1);
+    }
+
+    Step(delta);
+    Render();
+}
+
+float FluidSimulation::CalcDeltaTime()
```
Keenuts, line 487: It would be nice for this sample to handle `--deterministic`, so we could use it for screendiff tests.

Keenuts (approved):
LGTM for a first iteration, not let's make is 4k@60fps 😊

projects/fluid_simulation/sim.cpp
dnovillo: Yes.  I was thinking about `--deterministic` recently and thought that maybe it may be enough with giving  `ppx::Random` a fixed seed when `--deterministic` is used.  WDYT?

dnovillo:
> LGTM for a first iteration, not let's make is 4k@60fps 😊

Ha!  Yeah, it's actually pretty easy to peg the GPU by changing some of the resolution parameters.

projects/fluid_simulation/sim.cpp
Keenuts: yes, seeding rand to 0 is fine. But time-based effect should also be avoided.
ppx::Application has `GetElapsedSeconds` which returns the frame count when `--deterministic` is passed. But timers don't.

Keenuts: AFrame-count / 60.f actually, since we assume 60fps.

PR #214, opened 2023-06-26
Let KnobManager holds ownership of knobs.

hysw:


hysw:
@Slumpwuffle Not sure why I can't add you as a reviewer, but here's a fix for the shared_ptr comment on #200

Keenuts (commented):
This is something we discusses with @Slumpwuffle 
 - shared_ptr do makes ownership ambiguous, but solves dangling pointers issues. Not ideal, but prevent use-after-free in this case.
 - unique_ptr + export to raw/wrapped-raw does makes the ownership more clear, but adds the dangling pointers issue.

shared+weak does makes the ownership clearer, but has its drawbacks.
AFAIK he is working on a strong/weak pointer type we could use to sanitize not only those 2 APIs, but also the internal BigWheels weird pointer wrapping stuff.

As for this PR, I'm not convinced this is the solution:
 - adds a new type the user doesn't know about, which is specific for this API -> is it owned or not? not very clear.
 - doesn't solves use-after-free issues: unique gets deleted? still have a dangling pointer.

PR #215, opened 2023-06-27
Add a GPU FPS display

elviscapiaq:
- Also it was added a GPU work duration display.

### DIFF
--- a/benchmarks/graphics_pipeline/main.cpp
+++ b/benchmarks/graphics_pipeline/main.cpp
@@ -941,11 +941,8 @@ void ProjApp::UpdateGUI()
     }
 
     // GUI
-    ImGui::Begin("Debug");
+    ImGui::Begin("Knobs");
     GetKnobManager().DrawAllKnobs(true);
-    ImGui::End();
-
-    ImGui::Begin("Debug");
     DrawExtraInfo();
     ImGui::End();
 }
--- a/benchmarks/graphics_pipeline/main.cpp
+++ b/benchmarks/graphics_pipeline/main.cpp
@@ -906,7 +906,6 @@ void ProjApp::Render()
                     frame.cmd->DrawIndexed(renderable.pPrimitive->mesh->GetIndexCount());
                 }
             }
-
             DrawImGui(frame.cmd);
         }
         frame.cmd->EndRenderPass();
@@ -941,8 +940,9 @@ void ProjApp::UpdateGUI()
     }
 
     // GUI
-    ImGui::Begin("Knobs");
+    ImGui::Begin("Debug Window");
     GetKnobManager().DrawAllKnobs(true);
+    ImGui::Separator();
     DrawExtraInfo();
     ImGui::End();
 }
--- a/benchmarks/graphics_pipeline/main.cpp
+++ b/benchmarks/graphics_pipeline/main.cpp
@@ -82,6 +82,7 @@ private:
         grfx::FencePtr         imageAcquiredFence;
         grfx::SemaphorePtr     renderCompleteSemaphore;
         grfx::FencePtr         renderCompleteFence;
+        grfx::QueryPtr         timestampQuery;
     };
 
     struct Texture
@@ -141,6 +142,7 @@ private:
     float3                                                        mLightPosition = float3(10, 100, 10);
     std::array<Scene, kAvailableScenes.size()>                    mScenes;
     size_t                                                        mCurrentSceneIndex;
+    uint64_t                                                      mGpuWorkDuration;
 
     TextureCache mTextureCache;
 
@@ -200,6 +202,8 @@ private:
         std::vector<Material>*                                    pMaterials) const;
 
     void UpdateGUI();
+
+    void DrawExtraInfo();
 };
 
 void ProjApp::Config(ppx::ApplicationSettings& settings)
@@ -772,6 +776,12 @@ void ProjApp::Setup()
         fenceCreateInfo = {true}; // Create signaled
         PPX_CHECKED_CALL(GetDevice()->CreateFence(&fenceCreateInfo, &frame.renderCompleteFence));
 
+        // Timestamp query
+        grfx::QueryCreateInfo queryCreateInfo = {};
+        queryCreateInfo.type                  = grfx::QUERY_TYPE_TIMESTAMP;
+        queryCreateInfo.count                 = 2;
+        PPX_CHECKED_CALL(GetDevice()->CreateQuery(&queryCreateInfo, &frame.timestampQuery));
+
         mPerFrame.push_back(frame);
     }
 
@@ -798,6 +808,15 @@ void ProjApp::Render()
     PPX_CHECKED_CALL(frame.imageAcquiredFence->WaitAndReset());
     // Wait for and reset render complete fence
     PPX_CHECKED_CALL(frame.renderCompleteFence->WaitAndReset());
+
+    // Read query results
+    if (GetFrameCount() > 0) {
+        uint64_t data[2] = {0, 0};
+        PPX_CHECKED_CALL(frame.timestampQuery->GetData(data, 2 * sizeof(uint64_t)));
+        mGpuWorkDuration = data[1] - data[0];
+    }
+    // Reset query
+    frame.timestampQuery->Reset(0, 2);
 
     // Update camera(s)
     mCamera.LookAt(float3(2, 2, 2), float3(0, 0, 0));
@@ -860,6 +879,9 @@ void ProjApp::Render()
     // Build command buffer
     PPX_CHECKED_CALL(frame.cmd->Begin());
     {
+        // Write start timestamp
+        frame.cmd->WriteTimestamp(frame.timestampQuery, grfx::PIPELINE_STAGE_TOP_OF_PIPE_BIT, 0);
+
         grfx::RenderPassPtr renderPass = swapchain->GetRenderPass(imageIndex);
         PPX_ASSERT_MSG(!renderPass.IsNull(), "render pass object is null");
 
@@ -889,6 +911,12 @@ void ProjApp::Render()
         }
         frame.cmd->EndRenderPass();
         frame.cmd->TransitionImageLayout(renderPass->GetRenderTargetImage(0), PPX_ALL_SUBRESOURCES, grfx::RESOURCE_STATE_RENDER_TARGET, grfx::RESOURCE_STATE_PRESENT);
+
+        // Write end timestamp
+        frame.cmd->WriteTimestamp(frame.timestampQuery, grfx::PIPELINE_STAGE_TOP_OF_PIPE_BIT, 1);
+
+        // Resolve queries
+        frame.cmd->ResolveQueryData(frame.timestampQuery, 0, 2);
     }
     PPX_CHECKED_CALL(frame.cmd->End());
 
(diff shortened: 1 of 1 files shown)


mdagois:
> Is it possible to make use of the metrics framework for tracking GPU time? GPU work duration seems like a useful metric in future automated tests, which would need exported data (presumably through the metrics framework).

Yes, when used through the Application class, the metrics already track framerate, CPU frame time and frame count. It is possible to add "GPU frame time" too, but I believe it would require a bit more work to make sure the calculation is accurate, especially if there are more than 1 frame in flight, or if you are on a mobile architecture (very likely).

> Maybe @mdagois can also comment on if having a way for metrics to be visualized in the UI window is useful. I feel like there's some overlap between the goals here.

If the metrics are already used (or will be in the future), then you can get the N latest FPS measurements by using MetricGauge::GetEntriesCount and MetricGauge::GetEntry. Every application using metrics gathers framerate by default. You might need to add some API to access the data from Application::mMetrics::pFramerate.
However, the metrics might be overkill if you don't actually need them all the time. In that case, I'd use Application::GetAverageFPS directly. The Application class also has a rolling average mechanics if you want a more stable number.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -913,7 +942,29 @@ void ProjApp::UpdateGUI()
     }
 
     // GUI
-    GetKnobManager().DrawAllKnobs();
+    GetKnobManager().DrawAllKnobs(true);
```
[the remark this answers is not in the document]
elviscapiaq: Thansk for pointing that. Updated.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -913,7 +942,29 @@ void ProjApp::UpdateGUI()
...
+    // Draw additional elements
+    DrawGui();
+}
+
+void ProjApp::DrawGui()
```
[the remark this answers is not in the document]
elviscapiaq: I agree, Done.

elviscapiaq:
Ready for review.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -913,7 +941,32 @@ void ProjApp::UpdateGUI()
     }
 
     // GUI
-    GetKnobManager().DrawAllKnobs();
+    ImGui::Begin("Debug");
+    GetKnobManager().DrawAllKnobs(true);
+    ImGui::End();
```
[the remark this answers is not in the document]
elviscapiaq: Done.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -913,7 +941,32 @@ void ProjApp::UpdateGUI()
     }
 
     // GUI
-    GetKnobManager().DrawAllKnobs();
+    ImGui::Begin("Debug");
```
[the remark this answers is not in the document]
elviscapiaq: Done.

elviscapiaq:
I added a new commit to apply the feedback.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -799,6 +809,15 @@ void ProjApp::Render()
     // Wait for and reset render complete fence
     PPX_CHECKED_CALL(frame.renderCompleteFence->WaitAndReset());
 
+    // Read query results
+    if (GetFrameCount() > 0) {
+        uint64_t data[2] = {0, 0};
+        PPX_CHECKED_CALL(frame.timestampQuery->GetData(data, 2 * sizeof(uint64_t)));
```
Keenuts, line 815: `sizeof(data)`, compiler knows its length, so you'll get `sizeof(data[0])*2`

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -799,6 +809,15 @@ void ProjApp::Render()
...
+        mGpuWorkDuration = data[1] - data[0];
+    }
+    // Reset query
+    frame.timestampQuery->Reset(0, 2);
```
Keenuts, line 819: nit:
```cpp
frame.timestampQuery->Reset(/* firstQuery= */ 0, frame.timestampQuery->GetCount());
```

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -889,6 +911,12 @@ void ProjApp::Render()
         }
         frame.cmd->EndRenderPass();
         frame.cmd->TransitionImageLayout(renderPass->GetRenderTargetImage(0), PPX_ALL_SUBRESOURCES, grfx::RESOURCE_STATE_RENDER_TARGET, grfx::RESOURCE_STATE_PRESENT);
+
+        // Write end timestamp
+        frame.cmd->WriteTimestamp(frame.timestampQuery, grfx::PIPELINE_STAGE_TOP_OF_PIPE_BIT, 1);
+
+        // Resolve queries
+        frame.cmd->ResolveQueryData(frame.timestampQuery, 0, 2);
```
Keenuts, line 919:
```cpp
frame.cmd->ResolveQueryData(frame.timestampQuery, /* startIndex= */ 0, frame.timestampQuery->GetCount());
```

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -860,6 +879,9 @@ void ProjApp::Render()
     // Build command buffer
     PPX_CHECKED_CALL(frame.cmd->Begin());
     {
+        // Write start timestamp
+        frame.cmd->WriteTimestamp(frame.timestampQuery, grfx::PIPELINE_STAGE_TOP_OF_PIPE_BIT, 0);
```
Keenuts, line 883:
```cpp
frame.cmd->WriteTimestamp(frame.timestampQuery, grfx::PIPELINE_STAGE_TOP_OF_PIPE_BIT, /* queryIndex = */ 0);
```

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -889,6 +911,12 @@ void ProjApp::Render()
         }
         frame.cmd->EndRenderPass();
         frame.cmd->TransitionImageLayout(renderPass->GetRenderTargetImage(0), PPX_ALL_SUBRESOURCES, grfx::RESOURCE_STATE_RENDER_TARGET, grfx::RESOURCE_STATE_PRESENT);
+
+        // Write end timestamp
+        frame.cmd->WriteTimestamp(frame.timestampQuery, grfx::PIPELINE_STAGE_TOP_OF_PIPE_BIT, 1);
```
Keenuts, line 916:
```cpp
frame.cmd->WriteTimestamp(frame.timestampQuery, grfx::PIPELINE_STAGE_TOP_OF_PIPE_BIT, /* queryIndex = */ 1);
```

Keenuts:
> Maybe @mdagois can also comment on if having a way for metrics to be visualized in the UI window is useful. I feel like there's some overlap between the goals here.

I'd suppose having some idea about the pipeline execution duration in the UI could be useful for driver investigators (even if not completely accurate). Since once they figured a potentially interesting scenario, they create a benchmark from it to track it over time and confirm/dismiss their guess.

Keenuts:
> Yes, when used through the Application class, the metrics already track framerate, CPU frame time and frame count. It is possible to add "GPU frame time" too, but I believe it would require a bit more work to make sure the calculation is accurate, especially if there are more than 1 frame in flight, or if you are on a mobile architecture (very likely).

Would that be an issue to have the GPU time computed like that for now, and refine the computations later on when we have the need to be more accurate?

Keenuts (commented):
some changes regarding the code as it is today.

Now, more design-wise:
@mdagois my understanding was that the metrics framework not only bring the ability to export metrics, but would also be the canonical way to compute metrics.
So I'd say the GPU work duration should be a gauge and not handled like we did with some legacy metrics like frame count?
(I assumed framerate/timePerFrame was still handled application just because we haven't completed the transition to the metrics framework)

benchmarks/graphics_pipeline/main.cpp
elviscapiaq: Done.

benchmarks/graphics_pipeline/main.cpp
elviscapiaq: Done.

benchmarks/graphics_pipeline/main.cpp
elviscapiaq: Done.

benchmarks/graphics_pipeline/main.cpp
elviscapiaq: Done.

benchmarks/graphics_pipeline/main.cpp
elviscapiaq: Done.

elviscapiaq:
Ready for review!

mdagois:
> Would that be an issue to have the GPU time computed like that for now, and refine the computations later on when we have the need to be more accurate?

It should be easy to get an over estimation of the GPU frame time (that's what CPU frame time is with only 1 image in flight). I definitely think the metrics should be added to the framework side, so that we can improve the computation later and every app would benefit from it automatically. The only metrics that should be on the application side are the ones that are very specific to that application, like "image quality" for a particular algorithm.

mdagois:
> some changes regarding the code as it is today.
> 
> Now, more design-wise: @mdagois my understanding was that the metrics framework not only brings the ability to export metrics, but would also be the canonical way to compute metrics. So I'd say the GPU work duration should be a gauge and not handled like we did with some legacy metrics like frame count? (I assumed framerate/timePerFrame was still handled application just because we haven't completed the transition to the metrics framework)

That makes sense yes, but, as of today, metrics do not come for free. They consume memory over time, as the main goal is to output all the data (along with statistics computed over all that data) to a file at the end of the application execution. If the goal is just to display framerate, I am not sure it is worth using metrics in their current form.

elviscapiaq:
I was wondering if we can merge this PR. Based on https://github.com/google/bigwheels/blob/main/CONTRIBUTING.md, it seems only the reviewer who approved the PR can merge the PR?

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -913,7 +941,29 @@ void ProjApp::UpdateGUI()
...
+    ImGui::Text("GPU Work Duration");
+    ImGui::NextColumn();
+    ImGui::Text("%f ms ", gpuWorkDuration);
+    ImGui::NextColumn();
+
+    ImGui::Columns(2);
+    const float gpuFPS = static_cast<float>(frequency / static_cast<double>(mGpuWorkDuration));
+    ImGui::Text("GPU FPS");
+    ImGui::NextColumn();
+    ImGui::Text("%f fps ", gpuFPS);
```
apazylbe, line 966: Just a note: these numbers change every frame, and are not very stable, so it might be hard to read and interpret. 
I am good with this especially as a first iteration or if it serves the needs, but perhaps we could also consider adding average/plot/or other ways to make this easier to understand.

apazylbe (commented):
Overall lgtm just a couple of comments, I will pass it to Nathan to do approval and merging.

Keenuts:
> That makes sense yes, but, as of today, metrics do not come for free. They consume memory over time, as the main goal is to output all the data (along with statistics computed over all that data) to a file at the end of the application execution. If the goal is just to display framerate, I am not sure it is worth using metrics in their current form.

So far my understanding is:
 - those metrics are not reliable, but can be an hint. Might need some additional work in the future.
 - long term, we should move this to the metrics framework, but it's not ready right now for this use case.


So far I'm fine merging this as-is, it's still an improvement vs today, even if it's incomplete. And I assume this could be replaced in the future as the metrics framework becomes the main way to gather metrics.

Keenuts:
> I was wondering if we can merge this PR. Based on https://github.com/google/bigwheels/blob/main/CONTRIBUTING.md, it seems only the reviewer who approved the PR can merge the PR?

Yes, was waiting on action on https://github.com/google/bigwheels/pull/215/files#r1244290570
(Seems like you still needed to add a separator?)

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -913,7 +941,29 @@ void ProjApp::UpdateGUI()
     }
 
     // GUI
-    GetKnobManager().DrawAllKnobs();
+    ImGui::Begin("Knobs");
+    GetKnobManager().DrawAllKnobs(true);
+    DrawExtraInfo();
```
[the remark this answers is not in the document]
elviscapiaq: Done.

elviscapiaq:
Ready for review.

Keenuts (approved):

PR #218, opened 2023-06-29
remove std::move from return

Keenuts:
Remove superfluous std::move (-Werror,-Wpessimizing-move).


dnovillo (approved):
Thanks!  Remnants from a previous iteration of this.

PR #221, opened 2023-06-29
fluidsim: remove sources of non-determism

Keenuts:
GetElapsedTime() works well with --deterministic.
Timers don't.
But in this case, maybe removing deltatime computation entirely is fine?

### DIFF
--- a/projects/fluid_simulation/sim.cpp
+++ b/projects/fluid_simulation/sim.cpp
@@ -20,7 +20,7 @@ namespace FluidSim {
 
 // In a normal game, animations are linked to the frame delta-time to make then run
 // as a fixed perceptible speed. For our use-case (benchmarking), determinism is important.
-// Targeting 60ips.
+// Targeting 60 images per second.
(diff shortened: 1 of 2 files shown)


projects/fluid_simulation/sim.cpp
```diff
@@ -12,13 +12,17 @@
...
-#include "ppx/timer.h"
 
 #include <algorithm>
 #include <cmath>
 
 namespace FluidSim {
 
+// In a normal game, animations are linked to the frame delta-time to make then run
+// as a fixed perceptible speed. For our use-case (benchmarking), determinism is important.
+// Targeting 60ips.
```
dnovillo, line 23: ips? iterations per second?

dnovillo (approved):
Thanks!  It was in my queue as well.  At all times delta was a very constant .016f so this makes perfect sense.  Just a minor question.

projects/fluid_simulation/sim.cpp
Keenuts: images per seconds. I'll expand the abbreviation

PR #222, opened 2023-06-29
Add WASD functionality to keyboard

elviscapiaq:
- It allows to navigate into the scene where W = forward, S = backward, A = left and D = Right.

### DIFF
--- a/benchmarks/graphics_pipeline/main.cpp
+++ b/benchmarks/graphics_pipeline/main.cpp
@@ -47,19 +47,12 @@ public:
 
     FreeCamera()
     {
-        Setup();
+        mEyePosition = float3(0, 0, -5);
+        mTarget      = float3(0, 0, -4);
     }
 
     // Moves the location of the camera in dir direction for distance units.
     void Move(MovementDirection dir, float distance);
-
-private:
-    // Initializes the eye position and target of the camera.
-    void Setup()
-    {
-        mEyePosition = float3(0, 0, -5);
-        mTarget      = float3(0, 0, -4);
-    }
 };
 
 #if defined(USE_DX12)
@@ -176,7 +169,7 @@ private:
     float3                                                        mLightPosition = float3(10, 100, 10);
     std::array<Scene, kAvailableScenes.size()>                    mScenes;
     size_t                                                        mCurrentSceneIndex;
-    std::set<ppx::KeyCode>                                        mPressedKeys;
+    std::array<bool, TOTAL_KEY_COUNT>                             mPressedKeys = {0};
     uint64_t                                                      mGpuWorkDuration;
 
     TextureCache mTextureCache;
@@ -238,14 +231,32 @@ private:
 
     void SetupCamera();
 
-    void UpdateCamera();
-
     void ProcessInput();
 
     void UpdateGUI();
 
     void DrawExtraInfo();
 };
+
+void FreeCamera::Move(MovementDirection dir, float distance)
+{
+    switch (dir) {
+        case MovementDirection::FORWARD:
+            mEyePosition += float3(0, 0, distance);
+            break;
+        case MovementDirection::LEFT:
+            mEyePosition += float3(distance, 0, 0);
+            break;
+        case MovementDirection::RIGHT:
+            mEyePosition += float3(-distance, 0, 0);
+            break;
+        case MovementDirection::BACKWARD:
+            mEyePosition += float3(0, 0, -distance);
+            break;
+    }
+    mTarget = mEyePosition + float3(0, 0, 1);
+    LookAt(GetEyePosition(), GetTarget());
+}
 
 void ProjApp::Config(ppx::ApplicationSettings& settings)
 {
@@ -710,11 +721,6 @@ void ProjApp::LoadNodes(
 
 void ProjApp::SetupCamera()
 {
-    UpdateCamera();
-}
-
-void ProjApp::UpdateCamera()
-{
     mCamera.LookAt(mCamera.GetEyePosition(), mCamera.GetTarget());
     mCamera.SetPerspective(60.f, GetWindowAspect());
 }
@@ -847,58 +853,33 @@ void ProjApp::Setup()
 
 void ProjApp::KeyDown(ppx::KeyCode key)
 {
-    mPressedKeys.insert(key);
+    mPressedKeys[key] = true;
 }
 
 void ProjApp::KeyUp(ppx::KeyCode key)
 {
-    mPressedKeys.erase(key);
-}
-
-void FreeCamera::Move(MovementDirection dir, float distance)
-{
-    switch (dir) {
-        case MovementDirection::FORWARD:
-            mEyePosition += float3(0, 0, distance);
-            break;
-        case MovementDirection::LEFT:
-            mEyePosition += float3(distance, 0, 0);
-            break;
-        case MovementDirection::RIGHT:
-            mEyePosition += float3(-distance, 0, 0);
-            break;
-        case MovementDirection::BACKWARD:
-            mEyePosition += float3(0, 0, -distance);
-            break;
-    }
-    mTarget = mEyePosition + float3(0, 0, 1);
+    mPressedKeys[key] = false;
 }
 
 void ProjApp::ProcessInput()
 {
-    if (mPressedKeys.empty()) {
-        return;
-    }
-
     float deltaTime = GetPrevFrameTime();
 
-    if (mPressedKeys.count(ppx::KEY_W) > 0) {
+    if (mPressedKeys[KEY_W]) {
         mCamera.Move(FreeCamera::MovementDirection::FORWARD, kCameraSpeed * deltaTime);
     }
 
-    if (mPressedKeys.count(ppx::KEY_A) > 0) {
+    if (mPressedKeys[KEY_A]) {
         mCamera.Move(FreeCamera::MovementDirection::LEFT, kCameraSpeed * deltaTime);
     }
 
-    if (mPressedKeys.count(ppx::KEY_S) > 0) {
+    if (mPressedKeys[KEY_S]) {
         mCamera.Move(FreeCamera::MovementDirection::BACKWARD, kCameraSpeed * deltaTime);
     }
 
-    if (mPressedKeys.count(ppx::KEY_D) > 0) {
+    if (mPressedKeys[KEY_D]) {
         mCamera.Move(FreeCamera::MovementDirection::RIGHT, kCameraSpeed * deltaTime);
     }
-
-    UpdateCamera();
 }
 
 void ProjApp::Render()
--- a/benchmarks/graphics_pipeline/main.cpp
+++ b/benchmarks/graphics_pipeline/main.cpp
@@ -45,11 +45,16 @@ public:
         BACKWARD
     };
 
+    FreeCamera()
+    {
+        mEyePosition = float3(0, 0, -5),
+        mTarget      = float3(0, 0, -4);
+    }
+
     FreeCamera(float3 eyePosition, float3 target)
     {
         mEyePosition = eyePosition;
         mTarget      = target;
-        LookAt(mEyePosition, mTarget);
     }
 
     // Moves the location of the camera in dir direction for distance units.
@@ -166,7 +171,7 @@ private:
     grfx::DescriptorSetLayoutPtr                                  mSetLayout;
     std::array<grfx::ShaderModulePtr, kAvailableVsShaders.size()> mVsShaders;
     std::array<grfx::ShaderModulePtr, kAvailablePsShaders.size()> mPsShaders;
-    std::unique_ptr<FreeCamera>                                   mCamera;
+    FreeCamera                                                    mCamera;
     float3                                                        mLightPosition = float3(10, 100, 10);
     std::array<Scene, kAvailableScenes.size()>                    mScenes;
     size_t                                                        mCurrentSceneIndex;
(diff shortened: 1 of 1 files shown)


benchmarks/graphics_pipeline/main.cpp
```diff
@@ -783,6 +850,60 @@ void ProjApp::Setup()
...
+        location += float3(distance, 0, 0);
+    }
+    else if (dir == MovementDirection::RIGHT) {
+        location += float3(-distance, 0, 0);
+    }
+    else if (dir == MovementDirection::BACKWARD) {
+        location += float3(0, 0, -distance);
+    }
+    else {
+        PPX_ASSERT_MSG(false, "unhandled direction code " << static_cast<int>(dir));
```
Keenuts, line 878: if you use a switch, the compiler will warn if it's not exhaustive, meaning no need for this assert.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -30,6 +31,50 @@
 
 using namespace ppx;
 
+class Person
```
Keenuts, line 34: Maybe this could be a `ControllableCamera` or `FlyCamera` or `FreeCamera`?
And inherit from the camera class, to which you add the move function?

And seems like rate of move is stored, but not uses, instead you do `Person.Move(direction, Person.GetRateOfMove())`.
Could this rate of move be a constant?

So you'd have `mCamera.Move(direction, kCameraSpeed * deltatime)` ?

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -783,6 +850,60 @@ void ProjApp::Setup()
...
+
+void ProjApp::KeyUp(ppx::KeyCode key)
+{
+    mPressedKeys.erase(key);
+}
+
+void Person::Move(MovementDirection dir, float distance)
+{
+    if (dir == MovementDirection::FORWARD) {
+        location += float3(0, 0, distance);
```
Keenuts, line 866: This camera will be controlled by a human being, meaning we want some movement regularity, even with a very high/very low framerate. To handle that, you need to multiply each movement vector by deltatime.
(https://medium.com/star-gazers/understanding-time-deltatime-6528a8c2b5c8)

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -30,6 +31,50 @@
...
+        RIGHT,
+        BACKWARD
+    };
+
+    Person()
+    {
+        Setup();
+    }
+
+    // Initialize the location of this person.
```
apazylbe, line 50: nit: Function comments should follow an implicit "This function...", could you update this one and the ones below. So for this one it would be "Initializes..."

benchmarks/graphics_pipeline/main.cpp
elviscapiaq: Done.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -30,6 +31,50 @@
...
+    const float3 GetLookAt() const { return location + float3(0, 0, 1); }
+
+    // Return the location of the person in world space.
+    const float3 GetLocation() const { return location; }
+
+    float GetRateOfMove() const { return rateOfMove; }
+
+private:
+    // Coordinate in world space where the person is standing.
+    float3 location;
```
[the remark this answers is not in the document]
elviscapiaq: Done.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -30,6 +31,50 @@
...
+
+    // Rate of motion (grid units).
+    float rateOfMove;
```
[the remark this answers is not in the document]
elviscapiaq: Done.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -30,6 +31,50 @@
...
+    }
+
+    // Initialize the location of this person.
+    void Setup()
+    {
+        location   = float3(0, 0, -5);
+        rateOfMove = 0.2f;
+    }
+
+    // Move the location of this person in @param dir direction for @param distance units.
```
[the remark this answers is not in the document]
elviscapiaq: Done.

benchmarks/graphics_pipeline/main.cpp
elviscapiaq: Done.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -30,6 +31,50 @@
...
+        BACKWARD
+    };
+
+    Person()
+    {
+        Setup();
+    }
+
+    // Initialize the location of this person.
+    void Setup()
```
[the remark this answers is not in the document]
elviscapiaq: Thank you for pointing that. Updated.

benchmarks/graphics_pipeline/main.cpp
elviscapiaq: Done.

benchmarks/graphics_pipeline/main.cpp
elviscapiaq: Thank you for the clarification, I added it.

elviscapiaq:
Ready for review.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -30,6 +31,37 @@
...
+
+    // Moves the location of the camera in dir direction for distance units.
+    void Move(MovementDirection dir, float distance);
+
+private:
+    // Initializes the eye position and target of the camera.
+    void Setup()
+    {
+        mEyePosition = float3(0, 0, -5);
+        mTarget      = float3(0, 0, -4);
```
Keenuts, line 61: this should move in the constructor. Setup is called once.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -663,11 +704,22 @@ void ProjApp::LoadNodes(
...
 
+void ProjApp::SetupCamera()
+{
+    UpdateCamera();
+}
+
+void ProjApp::UpdateCamera()
+{
+    mCamera.LookAt(mCamera.GetEyePosition(), mCamera.GetTarget());
+    mCamera.SetPerspective(60.f, GetWindowAspect());
```
Keenuts, line 715: Is that expected to reset the perspective value each frame?

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -663,11 +704,22 @@ void ProjApp::LoadNodes(
     }
 }
 
+void ProjApp::SetupCamera()
+{
+    UpdateCamera();
```
Keenuts, line 709: Why is this called? Aren't the constructor default values enough?

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -783,6 +835,62 @@ void ProjApp::Setup()
...
+{
+    mPressedKeys.insert(key);
+}
+
+void ProjApp::KeyUp(ppx::KeyCode key)
+{
+    mPressedKeys.erase(key);
+}
+
+void FreeCamera::Move(MovementDirection dir, float distance)
```
Keenuts, line 848: maybe move this above all the ProjApp function definitions? So the function implem order matches the definition order.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -137,10 +171,11 @@ class ProjApp
     grfx::DescriptorSetLayoutPtr                                  mSetLayout;
     std::array<grfx::ShaderModulePtr, kAvailableVsShaders.size()> mVsShaders;
     std::array<grfx::ShaderModulePtr, kAvailablePsShaders.size()> mPsShaders;
-    PerspCamera                                                   mCamera;
+    FreeCamera                                                    mCamera;
     float3                                                        mLightPosition = float3(10, 100, 10);
     std::array<Scene, kAvailableScenes.size()>                    mScenes;
     size_t                                                        mCurrentSceneIndex;
+    std::set<ppx::KeyCode>                                        mPressedKeys;
```
Keenuts, line 178: an std::set is an RB-tree, with log complexity for insert.
What you want is probably an unordered_set.

Also, since the we have like 30 keycodes, maybe this could be a simple std::array with a boolean in each cell (`array[KEY_SPACE] = true`) ?

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -783,6 +835,62 @@ void ProjApp::Setup()
...
+
+    if (mPressedKeys.count(ppx::KEY_S) > 0) {
+        mCamera.Move(FreeCamera::MovementDirection::BACKWARD, kCameraSpeed * deltaTime);
+    }
+
+    if (mPressedKeys.count(ppx::KEY_D) > 0) {
+        mCamera.Move(FreeCamera::MovementDirection::RIGHT, kCameraSpeed * deltaTime);
+    }
+
+    UpdateCamera();
```
Keenuts, line 891: I find it a bit weird that I could call camera.move() and not move the camera until I call mCamera.LookAt.
Maybe mCamera.LookAt() should be called from inside camera.move(), so it's a real "move" function, not a "prepareMove".

dnovillo (commented):
There exists mouse and WASD movement support in https://github.com/google/bigwheels/blob/main/projects/20_camera_motion/main.cpp#L109.  I wonder if the common parts may be lifted into ppx/?

If it makes sense.  I haven't actually looked at the new implementation.

elviscapiaq:
> There exists mouse and WASD movement support in https://github.com/google/bigwheels/blob/main/projects/20_camera_motion/main.cpp#L109. I wonder if the common parts may be lifted into ppx/?
> 
> If it makes sense. I haven't actually looked at the new implementation.

Thanks for the update. That was the reference I took to implement this PR but I added/modified some  parts for our needs, I am also in favor of adding the common parts into ppx. Also I noticed that the 20_camera_motion sample has the mouse movement capability incorporated, however, it moves only among the xz-plane (y = 0), so I am working on adding the 3D. Once I finish this part I think I can also add into ppx.

dnovillo:
> > There exists mouse and WASD movement support in https://github.com/google/bigwheels/blob/main/projects/20_camera_motion/main.cpp#L109. I wonder if the common parts may be lifted into ppx/?
> > If it makes sense. I haven't actually looked at the new implementation.
> 
> Thanks for the update. That was the reference I took to implement this PR but I added/modified some parts for our needs, I am also in favor of adding the common parts into ppx. Also I noticed that the 20_camera_motion sample has the mouse movement capability incorporated, however, it moves only among the xz-plane (y = 0), so I am working on adding the 3D version of it. Once I finish this part I think I can also add into ppx.

Thanks!  To clarify my initial question: I'm not looking for this re-factor in this PR.  I was curious as to whether there is a possible re-factoring here, but it needs to (a) make sense, and (b) it can be done as a future addition.  No point blocking this PR unnecessarily.

benchmarks/graphics_pipeline/main.cpp
elviscapiaq: Done.

benchmarks/graphics_pipeline/main.cpp
elviscapiaq: Actually we don't need this call. Updated.

benchmarks/graphics_pipeline/main.cpp
elviscapiaq: Sound good to me, added before ProjApp method implementations.

elviscapiaq:
Ready for review.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -30,6 +31,30 @@
...
+    {
+        FORWARD,
+        LEFT,
+        RIGHT,
+        BACKWARD
+    };
+
+    FreeCamera()
+    {
+        mEyePosition = float3(0, 0, -5);
```
apazylbe, line 50: Would it make more sense to accept these in the constructor instead?

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -663,11 +715,17 @@ void ProjApp::LoadNodes(
     }
 }
 
+void ProjApp::SetupCamera()
+{
+    mCamera.LookAt(mCamera.GetEyePosition(), mCamera.GetTarget());
```
apazylbe, line 720: Why not call these in the constructor? It sounds like these have to be called, otherwise the camera is in the invalid state?

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -783,6 +841,37 @@ void ProjApp::Setup()
...
+{
+    mPressedKeys[key] = false;
+}
+
+void ProjApp::ProcessInput()
+{
+    float deltaTime = GetPrevFrameTime();
+
+    if (mPressedKeys[KEY_W]) {
+        mCamera.Move(FreeCamera::MovementDirection::FORWARD, kCameraSpeed * deltaTime);
```
apazylbe, line 859: I think it's fine as the first iteration, but there might be some weird edge cases here. What if the application is really slow and both KeyDown and KeyUp are called in between input processing events? Also this doesn't actually take into account when the duration of the key press.

apazylbe (commented):
I am fine with the changes, there woud probably more work involved in adding it to ppx, as some behaviour seems weird.

elviscapiaq:
Ready for review and also I rebased it.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -137,10 +165,11 @@ class ProjApp
     grfx::DescriptorSetLayoutPtr                                  mSetLayout;
     std::array<grfx::ShaderModulePtr, kAvailableVsShaders.size()> mVsShaders;
     std::array<grfx::ShaderModulePtr, kAvailablePsShaders.size()> mPsShaders;
-    PerspCamera                                                   mCamera;
+    std::unique_ptr<FreeCamera>                                   mCamera;
```
Keenuts, line 168: Why is this a pointer? Isn't a simple member enough since the lifetime of the camera is tied to the lifetime of the app?
Seems like the constructor could be called in the app constructor.

Keenuts (approved):

Keenuts (approved):

apazylbe (approved):

PR #223, opened 2023-06-30
Migrate FishTornado onto new metrics, add application-level metrics management wrappers, smaller fixes, etc.

Slumpwuffle:
This PR includes the following changes:
1. Fixed boolean command line flags to allow for explicit --argument=true/--argument=false syntax.
2. Fixed help output of command line arguments to have a consistent 80 width spacing, alphabetical order, and minimal special casing for XR.
3. Added metrics command line arguments to control whether metrics are enabled/disabled.
4. Fixed smaller errors in some calling conventions inside metrics.
5. Added application-level metrics management wrappers to avoid direct pointer access exposure, instead using a map with IDs which can be invalidated.
6. Migrated fishtornado's metrics system onto using the new, application-level metrics. This includes changes to enable FT's metrics even when imgui is disabled.

### DIFF
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -235,6 +235,7 @@ struct ApplicationSettings
     bool        allowThirdPartyAssets = false;
     // Set to true to opt-in for metrics.
     bool enableMetrics = false;
+    std::string reportPath    = "";
 
     struct
     {
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -334,7 +334,13 @@ protected:
     virtual void DispatchScroll(float dx, float dy);
     virtual void DispatchRender();
     virtual void DispatchInitKnobs();
-    virtual void DrawGui(){}; // Draw additional project-related information to ImGui.
+    virtual void DispatchUpdateMetrics();
+    virtual void DrawGui() {} // Draw additional project-related information to ImGui.
+
+    // Override these methods in a derived class to change the default behavior of metrics.
+    virtual void SetupMetrics();
+    virtual void ShutdownMetrics();
+    virtual void UpdateMetrics() {}
 
     void TakeScreenshot();
 
@@ -343,6 +349,16 @@ protected:
     void DrawProfilerGrfxApiFunctions();
 
     KnobManager& GetKnobManager() { return mKnobManager; }
+
+public:
+    typedef uint32_t          MetricID;
+    static constexpr MetricID kInvalidMetricID = 0;
+
+    enum class MetricType
+    {
+        kGauge   = 1,
+        kCounter = 2,
+    };
 
 public:
     int  Run(int argc, char** argv);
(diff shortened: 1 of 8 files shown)


Slumpwuffle:
Fairly large PR; can be broken up if folks feel it's risky. Hopefully the commit breakdown makes it easier to review.

Keenuts:
> Fairly large PR; can be broken up if folks feel it's risky. Hopefully the commit breakdown makes it easier to review.

Quickly glanced (will take a proper look tomorrow). But as you said, if you split the PR, trivial changes can be merged quickly and we can focus on the other parts.
Some parts are clearly independant from the wrapper change.

include/ppx/application.h
```diff
@@ -526,6 +549,10 @@ class Application
         double   framerateRecordTimer   = 0.0;
         uint64_t framerateFrameCount    = 0;
         bool     resetFramerateTracking = true;
+
+        MetricID                                              nextMetricID = kInvalidMetricID + 1;
+        std::unordered_map<MetricID, metrics::MetricGauge*>   gauges;
```
apazylbe, line 554: Would it be cleaner if we move more of the record keeping into the metrics Run class and move `MetricID`, `MetricType`, `kInvalidMetricID` into `metrics.h`? Then Run's `AddMetric` can return a MetricID instead of a pointer and provide a separate function to get a metric by ID? It seems like the metrics run would be a more logical place to do the bookkeeping for the IDs since they are tied to the run lifetime anyway. Thoughts?

Keenuts: Yes. Using opaque handles solves the issues we have today, and I'm fine converting all APIs to opaque pointers. (IIRC Angela had this pattern the the initial knob API, I had the bad idea to suggest otherwise, sorry..).
But doing it at the application-level seems weird. As Aliya said, moving that at the metrics API level would be better.

Slumpwuffle: I'd considered this, and am fine with doing it, BUT it changes the semantics of the metrics API.

Right now, at the metrics level it's possible to have multiple simultaneous runs and different metrics associated with each. The application's specific USE of the metrics API, however, limits things to a single active run. This makes adding metrics trivial: you can call into the application to add a metrics to the active run; if there's no active run, no metric will be added.

Moving it into the metrics API means now the manager needs to take this into account. AddMetric will need to be in the manager to avoid the stale run scenario, which means AddMetric will either need to take a 'runId' argument, or the manager itself will be limited to having a single active run.

I don't personally see a reason for Manager to have more than a single active run, honestly, so I'm totally fine with locking things down at that level, and if we ultimately decide later that's not what we need, we can change things to suit.

Slumpwuffle:
Alright, let me split this up a bit.

Slumpwuffle:
PR has been split up and is going through review independently.

PR #225, opened 2023-06-30
Update Android build to use multi-project setup

apazylbe:
Use multi-project build setup instead of build flavours for samples. The root project `build.gradle` specifies how to build, and each subproject just creates a copyTask to copy assets that are used by the project.

Updates/benefits:
- In Android Studio, easier way to find and switch samples and no gradle sync when switching build flavours
- In CLI more straightforward way to discover projects and no need to remember syntax on how to build.
- Specifies assets per project, resulting in smaller APKs. Third party assets are only used in a few projects, and those are very large (> 300 MB). No need to add them to projects that don't use them.
- Instead of split APKS use abi filters to only build for arm64-v8a
- Build all samples in the github CI
- Modifies the default location of build and .cxx folders. The default would be in projects/<subproject> but
I don't like polluting it there, so modifying to a centralized location, so it's easier to manage/delete.

One drawback of the new setup is specifying assets that are used inside the project, so if the project is updated we have to remember to update the build.gradle file, and there is also no automated way to test right now.

### DIFF
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -16,8 +16,6 @@ jobs:
     runs-on: "ubuntu-20.04"
     strategy:
       fail-fast: true
-      matrix:
-        task: ["assembleMobileBasicMaterial", "assembleMobileCube", "assembleMobileFishtornado", "assembleMobileGbuffer", "assembleMobileTriangle" ]
     steps:
       - name: Checkout the code
         uses: actions/checkout@v3
@@ -35,6 +33,6 @@ jobs:
           mkdir "${HOME}/vulkan-sdk"
           tar -xf vulkansdk-linux-x86_64-1.3.216.0.tar.gz -C "${HOME}/vulkan-sdk"
           echo "VULKAN_SDK=${HOME}/vulkan-sdk/1.3.216.0/x86_64" >> $GITHUB_ENV
-      - name: Build the sample for Android
+      - name: Build all projects
         run: |
-          ./gradlew ${{ matrix.task }}
+          ./gradlew buildRelease
--- a/build.gradle
+++ b/build.gradle
@@ -77,8 +77,7 @@ subprojects {
                     }
                     else {
                         arguments '-DPPX_ANDROID=TRUE',
-                                  '-DPPX_BUILD_TESTS=FALSE',
-                                  '-DBUILD_TESTS=FALSE' // Required for OpenXR
+                                  '-DPPX_BUILD_TESTS=FALSE'
                     }
                     if (project.name.contains("_xr")) {
                         cppFlags '-std=c++20',
--- a/build.gradle
+++ b/build.gradle
@@ -1,159 +1,114 @@
-plugins {
-    id 'com.android.application' version '7.3.1'
-    id 'com.android.library' version '7.3.1' apply false
+ext {
+    assetsPath = '../../assets'
+    thirdPartyAssetsPath = '../../third_party/assets'
 }
 
-android {
-    namespace 'com.google.bigwheels'
-    compileSdk 32
+subprojects {
+    repositories {
+        mavenCentral()
+        google()
+    }
 
-    defaultConfig {
-        applicationId "com.google.bigwheels"
-        minSdk 29
-        targetSdk 32
-        versionCode 1
-        versionName "1.0"
+    apply plugin: "com.android.application"
 
-        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+    dependencies {
+        implementation 'androidx.appcompat:appcompat:1.5.1'
+        implementation 'com.google.android.material:material:1.7.0'
+        testImplementation 'junit:junit:4.13.2'
+        androidTestImplementation 'androidx.test.ext:junit:1.1.4'
+        androidTestImplementation 'androidx.test.espresso:espresso-core:3.5.0'
+    }
+
+    buildDir = "../../build/android/${project.name}"
+
+    android {
+        namespace 'com.google.bigwheels'
+        compileSdkVersion 32
+
+        compileOptions {
+            sourceCompatibility JavaVersion.VERSION_1_8
+            targetCompatibility JavaVersion.VERSION_1_8
+        }
+        buildFeatures {
+            prefab true
+        }
         externalNativeBuild {
             cmake {
-                arguments '-DPPX_ANDROID=TRUE',
-                          '-DPPX_BUILD_TESTS=FALSE',
-                          '-DBUILD_TESTS=FALSE' // Required for OpenXR
-                cppFlags '-std=c++20'
-                if (project.hasProperty('DXC_PATH')) {
-                  arguments "-DDXC_PATH=$DXC_PATH"
-                }
+                path file('../../CMakeLists.txt')
+                version '3.22.1+'
+                buildStagingDirectory "../../.cxx/${project.name}"
             }
         }
-        multiDexEnabled true
-    }
 
-    buildTypes {
-        release {
-            minifyEnabled false
-        }
-        debug {
-          applicationIdSuffix '.debug'
-          debuggable true
-        }
-    }
-    def defineFlavor = { sample_name ->
-      return {
-        dimension 'project'
-        // Package names cannot start with a number, adding prefixing.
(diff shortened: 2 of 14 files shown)


build.gradle
```diff
@@ -1,159 +1,114 @@
...
     }
 
-    buildTypes {
-        release {
-            minifyEnabled false
+    buildDir = "../../build/android/${project.name}"
+
+    android {
+        namespace 'com.google.bigwheels'
+        compileSdkVersion 32
```
Keenuts, line 26: nit: compileSdk? unless going back to the old naming is required)

build.gradle
```diff
@@ -1,159 +1,114 @@
...
-        applicationIdSuffix ".project_" + sample_name
-        manifestPlaceholders = [
-            sampleLibraryName : "vk_" + sample_name,
-            appLabel : "vk_" + sample_name
-        ]
-        it.buildConfigField "String", "sample_library_name", '"vk_' + sample_name + '"'
         externalNativeBuild {
             cmake {
-                targets "vk_" + sample_name
+                path file('../../CMakeLists.txt')
```
Keenuts, line 37: ${rootProject.projectDir}

build.gradle
```diff
@@ -1,159 +1,114 @@
...
+                    if (project.hasProperty('DXC_PATH')) {
+                      arguments "-DDXC_PATH=$DXC_PATH"
+                    }
+                    targets "vk_${project.name}"
+                }
+            }
+            multiDexEnabled true
+            it.buildConfigField "String", "sample_library_name", '"vk_' + project.name + '"'
+            sourceSets.main {
+                java.srcDirs = [ '../../src/android' ]
```
Keenuts, line 100: ${rootProject.projectDir}

build.gradle
```diff
@@ -1,159 +1,114 @@
...
+                assets.srcDirs = ['assets']
+                if (project.name.contains("_xr")) {
+                    manifest.srcFile '../../src/android/AndroidManifest.XR.xml'
```
Keenuts, line 103: ${rootProject.projectDir}

build.gradle
```diff
@@ -1,159 +1,114 @@
...
+                } else {
+                    manifest.srcFile '../../src/android/AndroidManifest.xml'
```
Keenuts, line 105: ${rootProject.projectDir}

projects/04_cube_xr/build.gradle
```diff
@@ -0,0 +1,11 @@
+task copyTask {
```
Keenuts, line 1: Also re-add the simple 04_cube (existed before)

projects/04_cube_xr/build.gradle
```diff
@@ -0,0 +1,11 @@
...
+    copy {
+        from rootProject.ext.assetsPath + '/basic/shaders/spv'
+        into 'assets/basic/shaders/spv'
+        include '*.spv'
+    }
+    copy {
+        from rootProject.ext.assetsPath + '/fishtornado'
+        into 'assets/fishtornado'
+    }
```
Keenuts, line 10: not needed for the cube sample

.github/workflows/android-build.yml
```diff
@@ -35,6 +33,6 @@ jobs:
           mkdir "${HOME}/vulkan-sdk"
           tar -xf vulkansdk-linux-x86_64-1.3.216.0.tar.gz -C "${HOME}/vulkan-sdk"
           echo "VULKAN_SDK=${HOME}/vulkan-sdk/1.3.216.0/x86_64" >> $GITHUB_ENV
-      - name: Build the sample for Android
+      - name: Build all projects
         run: |
-          ./gradlew ${{ matrix.task }}
+          ./gradlew buildRelease
```
Keenuts, line 38: What is the required storage for those? Are we close to the limit or not?

Keenuts:
The change is nice, thanks for doing that. Some fixes to remove weird `../..` paths, and some details, but overall looks good.

.github/workflows/android-build.yml
apazylbe: This seems to add about 9 GB, and adds to the <2 GB of the repository itself. The runners are guaranteed to have at least 14 GB, so its not close to the limit. FWIW, in my tests the available space was actually 24 GB on the runners, but I think only 14 is actually guaranteed.

build.gradle
```diff
@@ -1,159 +1,114 @@
-plugins {
-    id 'com.android.application' version '7.3.1'
-    id 'com.android.library' version '7.3.1' apply false
+ext {
+    assetsPath = '../../assets'
```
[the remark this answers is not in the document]
apazylbe: I used rootProject.projectDir directly, do you think there are any issues with that?

build.gradle
```diff
@@ -1,159 +1,114 @@
...
+        implementation 'com.google.android.material:material:1.7.0'
+        testImplementation 'junit:junit:4.13.2'
+        androidTestImplementation 'androidx.test.ext:junit:1.1.4'
+        androidTestImplementation 'androidx.test.espresso:espresso-core:3.5.0'
     }
 
-    buildTypes {
-        release {
-            minifyEnabled false
+    buildDir = "../../build/android/${project.name}"
```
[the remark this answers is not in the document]
apazylbe: Done. thanks!

build.gradle
apazylbe: Changed, thanks! not sure why I put compileSdkVersion before

build.gradle
apazylbe: Done, thanks!

build.gradle
```diff
@@ -1,159 +1,114 @@
...
-            sampleLibraryName : "vk_" + sample_name,
-            appLabel : "vk_" + sample_name
-        ]
-        it.buildConfigField "String", "sample_library_name", '"vk_' + sample_name + '"'
         externalNativeBuild {
             cmake {
-                targets "vk_" + sample_name
+                path file('../../CMakeLists.txt')
+                version '3.22.1+'
+                buildStagingDirectory "../../.cxx/${project.name}"
```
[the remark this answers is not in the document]
apazylbe: Without it, it puts .cxx directory in the projects/<project-name> and I kinda like having the centralized place for this, so I can delete it, exclude from grep etc.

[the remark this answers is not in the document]
apazylbe: I assume you meant buildStagingDirectory folder line

build.gradle
apazylbe: Done

build.gradle
apazylbe: Done.

build.gradle
apazylbe: done.

projects/04_cube_xr/build.gradle
apazylbe: oops, re-added, thanks!

projects/04_cube_xr/build.gradle
apazylbe: good catch, thanks!

build.gradle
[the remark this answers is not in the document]
Keenuts: Ah yes, I hadn't cleaned between my experiments and saw the old .cxx and missed the new ones in project. Makes sense! thanks

build.gradle
```diff
@@ -1,159 +1,113 @@
...
+        google()
+    }
 
-    defaultConfig {
-        applicationId "com.google.bigwheels"
-        minSdk 29
-        targetSdk 32
-        versionCode 1
-        versionName "1.0"
+    apply plugin: "com.android.application"
```
Keenuts, line 12: missed it before: no idea if that's important, but before we fixed the application plugin version to something. Seems like it would be a good thing to keep?

.github/workflows/android-build.yml
Keenuts: All right, so we do have some margin left 😊
Let's move back to 1 workflow for this then.

Keenuts (approved):
LGTM! Just a small nit, otherwise fine with this. I'll let you merge once fixed.

build.gradle
apazylbe: Hm, I thought the version came from `com.android.tools.build:gradle` in the `settings.gradle` but can't find a confirmation on that. I added a plugins block with version to `settings.gradle` for this.

PR #233, opened 2023-07-06
Adds push constants support for D3D12 and Vulkan

chaoticbob:
This change list adds push constants support for both D3D12 and Vulkan. The implementation currently limits the number of constants to 32 to keep it inline with what the majority of the Vulkan drivers out there can safely support. Apologies this is a big one but a lot of it has to come in all at once.

- Updated grfx::PipelineInterfaceCreateInfo to include a struct member for push constants parameters.
- Updated D3D12's pipeline interface object to create extra root parameter for push constants - using D3D12's root constants.
- Updated Vulkan's pipeline interface object to include a push constants range in pipeline layout creation.
- Added interface functions to grfx::CommandBuffer to set push constants for graphic and compute.
- Updated D3D12's and Vulkan's respective command buffer objects with implementations for new grfx::CommandBuffer interface functions.
- Added misc functions to D3D12 and Vulkan objects to help support implementations.
- Updated D3D12's command buffer to only set a graphics or compute root signature when it has to...otherwise it just skips it.
- Added sample to demonstrate how to use push constants.
- Removed unused code.
- Fixed some bad comments.


projects/CMakeLists.txt
```diff
@@ -41,6 +41,7 @@ add_subdirectory(22_image_filter)
 add_subdirectory(23_async_compute)
 add_subdirectory(27_mipmap_demo)
```
apazylbe, line 42: I don't know if projects 24-26 are reserved for something, but perhaps makes sense to fill existing missing numbers?

src/ppx/grfx/dx12/dx12_command.cpp
```diff
@@ -380,6 +384,24 @@ void CommandBuffer::SetScissors(
     mCommandList->RSSetScissorRects(static_cast<UINT>(scissorCount), rects);
 }
 
+void CommandBuffer::SetGraphicsPipelineInterface(const grfx::PipelineInterface* pInterface)
+{
+    // Only set oot signature if we have to
```
apazylbe, line 389: typo in root?

src/ppx/grfx/dx12/dx12_command.cpp
```diff
@@ -380,6 +384,24 @@ void CommandBuffer::SetScissors(
...
+    if (pInterface != mCurrentGraphicsInterface) {
+        mCurrentGraphicsInterface = pInterface;
+        mCommandList->SetGraphicsRootSignature(ToApi(mCurrentGraphicsInterface)->GetDxRootSignature().Get());
+    }
+}
+
+void CommandBuffer::SetComputePipelineInterface(const grfx::PipelineInterface* pInterface)
+{
+    // Only set oot signature if we have to
```
apazylbe, line 398: same typo in root?

src/ppx/grfx/dx12/dx12_pipeline.cpp
```diff
@@ -320,62 +320,21 @@ Result PipelineInterface::CreateApiObjects(const grfx::PipelineInterfaceCreateIn
...
+        D3D12_ROOT_PARAMETER1 parameter    = {};
+        parameter.ParameterType            = D3D12_ROOT_PARAMETER_TYPE_32BIT_CONSTANTS;
+        parameter.Constants.ShaderRegister = static_cast<UINT>(pCreateInfo->pushConstants.binding);
+        parameter.Constants.RegisterSpace  = static_cast<UINT>(pCreateInfo->pushConstants.set);
+        parameter.Constants.Num32BitValues = static_cast<UINT>(pCreateInfo->pushConstants.count);
+        parameter.ShaderVisibility         = ToD3D12ShaderVisibliity(pCreateInfo->pushConstants.shaderVisiblity);
+        // Store parameter
+        parameters.push_back(parameter);
+
+        // Store a specific paramter index for root constants to get to it without a search.
```
apazylbe, line 336: typo in parameter

assets/basic/shaders/PushConstantsTexture.hlsl
```diff
@@ -0,0 +1,50 @@
...
+#if defined(__spirv__)
+[[vk::push_constant]]
+#endif
+ConstantBuffer<DrawParams> Draw : register(b0);
+
+Texture2D    Textures[3]: register(t1);
+SamplerState Sampler0   : register(s4);
+
+struct VSOutput {
+	float4 Position : SV_POSITION;
```
apazylbe, line 35: nit: there is mixed use of tabs and spaces in the file, would be nice to be consistent, but up to you

apazylbe (commented):
Just some nits on my side, but otherwise lgtm

projects/29_push_constants/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+# Copyright 2022 Google LLC
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

projects/29_push_constants/PushConstants.cpp
```diff
@@ -0,0 +1,369 @@
+// Copyright 2022 Google LLC
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

projects/CMakeLists.txt
chaoticbob: I don't think so. I've changed the project to use 24.

src/ppx/grfx/dx12/dx12_command.cpp
chaoticbob: Yep. Fixed.

src/ppx/grfx/dx12/dx12_command.cpp
chaoticbob: Fixed.

src/ppx/grfx/dx12/dx12_pipeline.cpp
chaoticbob: Fixed.

assets/basic/shaders/PushConstantsTexture.hlsl
chaoticbob: Not sure how this happened. Fixed to all spaces.

include/ppx/grfx/grfx_command.h
```diff
@@ -300,13 +300,31 @@ class CommandBuffer
...
 
+    //
+    // Parameters count and offset are measured in DWORDs (uint32_t) aka 32-bit values.
+    // To set the first 4 32-bit values, use count = 4, offset = 0.
+    // To set the 16 DWORDs starting at offset 8, use count=16, offset = 8.
+    //
+    virtual void SetGraphicsPushConstants(
+        const grfx::PipelineInterface* pInterface,
+        uint32_t                       count,
+        const void*                    pValues,
```
Keenuts, line 311: Why a `void*`? the comment above says it's handled as 32-bit words, so why not `uint32_t*`? This way the type tells us how to handle count, as with a void, it's a bit unclear (from prototype, comment of course makes it clear).

include/ppx/grfx/grfx_command.h
```diff
@@ -300,13 +300,31 @@ class CommandBuffer
...
     virtual void BindComputeDescriptorSets(
         const grfx::PipelineInterface*    pInterface,
         uint32_t                          setCount,
         const grfx::DescriptorSet* const* ppSets) = 0;
 
+    // See comments at SetGraphicsPushConstants for explanation about count and offset.
+    virtual void SetComputePushConstants(
+        const grfx::PipelineInterface* pInterface,
+        uint32_t                       count,
+        const void*                    pValues,
```
Keenuts, line 325: same remark `void*`->`uint32_t*`

projects/24_push_constants/PushConstants.cpp
```diff
@@ -0,0 +1,369 @@
...
+const grfx::Api kApi = grfx::API_VK_1_1;
+#endif
+
+void PushConstantsApp::Config(ppx::ApplicationSettings& settings)
+{
+    settings.appName                    = "24_push_constants";
+    settings.enableImGui                = true;
+    settings.grfx.api                   = kApi;
+    settings.grfx.swapchain.depthFormat = grfx::FORMAT_D32_FLOAT;
+    settings.grfx.enableDebug           = true;
```
Keenuts, line 32: seems like we ship samples with debug = false by default.

projects/24_push_constants/PushConstants.cpp
```diff
@@ -0,0 +1,369 @@
...
+            {
+                // Calculate MVP
+                float4x4 P   = glm::perspective(glm::radians(60.0f), GetWindowAspect(), 0.001f, 10000.0f);
+                float4x4 V   = glm::lookAt(float3(0, 0, 3), float3(0, 0, 0), float3(0, 1, 0));
+                float4x4 T   = glm::translate(float3(0, 0, -10 * (1 + sin(t / 2))));
+                float4x4 R   = glm::rotate(t / 4, float3(0, 0, 1)) * glm::rotate(t / 4, float3(0, 1, 0)) * glm::rotate(t / 4, float3(1, 0, 0));
+                float4x4 M   = T * R;
+                float4x4 mat = P * V * M;
+                // Set MVP push constants
+                frame.cmd->SetGraphicsPushConstants(mPipelineInterface, 16, &mat);
```
Keenuts, line 307: nit: sizeof(float4x4)/sizeof(uint32_t)? up to you TBH

projects/24_push_constants/PushConstants.cpp
```diff
@@ -0,0 +1,369 @@
...
+        bytecode = LoadShader("basic/shaders", "PushConstantsTexture.ps");
+        PPX_ASSERT_MSG(!bytecode.empty(), "PS shader bytecode load failed");
+        shaderCreateInfo = {static_cast<uint32_t>(bytecode.size()), bytecode.data()};
+        PPX_CHECKED_CALL(GetDevice()->CreateShaderModule(&shaderCreateInfo, &mPS));
+
+        grfx::PipelineInterfaceCreateInfo piCreateInfo = {};
+        piCreateInfo.setCount                          = 1;
+        piCreateInfo.sets[0].set                       = 0;
+        piCreateInfo.sets[0].pLayout                   = mDescriptorSetLayout;
+        piCreateInfo.pushConstants.count               = 17; // float4x4 + uint
```
Keenuts, line 145: `(sizeof(float4x4) + sizeof(uint32_t)) / sizeof(uint32_t)` ?

projects/24_push_constants/PushConstants.cpp
```diff
@@ -0,0 +1,369 @@
...
+            // Bind descriptor set
+            frame.cmd->BindGraphicsDescriptorSets(mPipelineInterface, 1, &mDescriptorSet);
+
+            // Get elapsed seconds
+            float t = GetElapsedSeconds();
+
+            // Draw center cube
+            {
+                // Calculate MVP
+                float4x4 P   = glm::perspective(glm::radians(60.0f), GetWindowAspect(), 0.001f, 10000.0f);
```
Keenuts, line 300: move P and V as const in the scope above?

include/ppx/grfx/grfx_command.h
```diff
@@ -300,13 +300,31 @@ class CommandBuffer
...
+    //
+    // Parameters count and offset are measured in DWORDs (uint32_t) aka 32-bit values.
+    // To set the first 4 32-bit values, use count = 4, offset = 0.
+    // To set the 16 DWORDs starting at offset 8, use count=16, offset = 8.
+    //
+    virtual void SetGraphicsPushConstants(
+        const grfx::PipelineInterface* pInterface,
+        uint32_t                       count,
+        const void*                    pValues,
+        uint32_t                       offset = 0) = 0;
```
Keenuts, line 312: I find `offset` a bit unclear:
 - if it the offset to read from in the pValue vector? Or the push-constant offset to start updating from (like in vkCmdPushConstants).
 
 Either the param should be renamed (maybe `dstOffset` like similar cases in VK?) and/or the comment be changed to explain this bit?
 I find that `to set the 16 DWORD starting at offset` could mean we skip the first 16 words in both pValues and the push buffer.

src/ppx/grfx/vk/vk_pipeline.cpp
```diff
@@ -172,22 +173,22 @@ Result GraphicsPipeline::InitializeVertexInput(
...
-    //for (auto& binding : mInputBindings) {
-    //    for (auto& attribute : binding.attributes) {
-    //        VkVertexInputAttributeDescription vkAttribute = {};
-    //        vkAttribute.location                          = attribute.location;
-    //        vkAttribute.binding                           = attribute.binding;
-    //        vkAttribute.format                            = ToVkFormat(attribute.format);
-    //        vkAttribute.offset                            = attribute.offset;
-    //        vkAttributes.push_back(vkAttribute);
-    //    }
+    // for (auto& binding : mInputBindings) {
```
Keenuts, line 176: maybe remove this dead code? (since it's getting formatted)

include/ppx/grfx/grfx_pipeline.h
```diff
@@ -243,6 +227,24 @@ struct PipelineInterfaceCreateInfo
...
+    //
+    // The binding and set for push constants CAN NOT overlap with a binding
+    // AND set in sets (the struct immediately above this one). It's okay for
+    // push constants to be in an existing set at binding that is not used
+    // by an entry in the set layout.
+    //
+    struct
+    {
+        uint32_t              count           = 0;                 // Measured in DWORDs, must be less than or equal to PPX_MAX_PUSH_CONSTANTS
+        uint32_t              binding         = PPX_VALUE_IGNORED; // D3D12 only, ignored by Vulkan
```
Keenuts, line 244: Not very familiar with DXC register/binding relationship.
If I add a second ConstantBuffer in the shader, this time bound to the register `b9`,
DXC seems to say I have a constant buffers, bound to `cb0` and `cb9`.
So I'd have 2 bindings to do here, 1 with a binding to 0, 1 with a binding set to 9 no?

On the Vulkan side, seems like we can only have 1 push constant, so should BigWheels block this to 1 too on the DXC side? and maybe assert that binding is 0?
It might be restrictive, but at least is we compile a sample on windows+dx, we'd assert, and could learn "oh, vulkan doesn't support that, to be portable I need to do X?"

(Same reason we block the constant count to 32?)

Keenuts (commented):
Looks good! Some questions/remarks on API and VK/DX divergence.

include/ppx/grfx/grfx_command.h
chaoticbob: Same rationale as above. 

Also the distinction for Graphics and Compute is forced by DX since we don't know yet which type of pipeline is bound.

projects/24_push_constants/PushConstants.cpp
chaoticbob: Oops, my bad, I'll turn it off. Had it on to debug during dev.

projects/24_push_constants/PushConstants.cpp
chaoticbob: I tried this, but it looked a bit busy and more difficult to read compared to the usage in the create info. Lets just keep the constants for now. Maybe we can add a template function later to make it a bit more clear.

projects/24_push_constants/PushConstants.cpp
chaoticbob: Done.

src/ppx/grfx/vk/vk_pipeline.cpp
chaoticbob: Done.

include/ppx/grfx/grfx_pipeline.h
chaoticbob: > Not very familiar with DXC register/binding relationship. If I add a second ConstantBuffer in the shader, this time bound to the register `b9`, DXC seems to say I have a constant buffers, bound to `cb0` and `cb9`. So I'd have 2 bindings to do here, 1 with a binding to 0, 1 with a binding set to 9 no?

Do you mean DX here or is DXC on purpose? The example you gave is correct, there would be two bindings at 1 and 9. I think I'm misunderstanding something here. Can you add some more detail?

> 
> On the Vulkan side, seems like we can only have 1 push constant, so should BigWheels block this to 1 too on the DXC side? and maybe assert that binding is 0? It might be restrictive, but at least is we compile a sample on windows+dx, we'd assert, and could learn "oh, vulkan doesn't support that, to be portable I need to do X?"

I'm not able to follow, can you give an example?

include/ppx/grfx/grfx_command.h
chaoticbob: Changed to dstOffset.

include/ppx/grfx/grfx_pipeline.h
Keenuts: DX for DirectX, sorry, too used to say DXC..
After looking at the PipelineInterfaceCreateInfo, we do enforce 1 pushConstant per pipeline anyway, so my concern is not valid.
One could create 2 pushconstant in a shader, which is valid on DirectX, but won't be able to use them both in BigWheels, so that's fine.

Was worried we could do something in DirectX which was invalid in Vulkan. But no. Thanks!

include/ppx/grfx/grfx_command.h
Keenuts: Yeah, right, so uint32_t isn't quite right either. Fine for void* then!

include/ppx/grfx/vk/vk_command.h
```diff
@@ -157,6 +169,12 @@ class CommandBuffer
         uint32_t                          setCount,
         const grfx::DescriptorSet* const* ppSets);
 
+    void SetPushConstants(
+        const grfx::PipelineInterface* pInterface,
+        uint32_t                       count,
+        const void*                    pValues,
+        uint32_t                       offset);
```
Keenuts, line 176: nit: dstOffset too?

Keenuts (approved):
Thanks, one small nit, approval.
You have write access, so I'll let you push

include/ppx/grfx/vk/vk_command.h
```diff
@@ -70,13 +70,25 @@ class CommandBuffer
         uint32_t                          setCount,
         const grfx::DescriptorSet* const* ppSets) override;
 
+    virtual void SetGraphicsPushConstants(
+        const grfx::PipelineInterface* pInterface,
+        uint32_t                       count,
+        const void*                    pValues,
```
dnovillo, line 76: So, this means that the caller passes a contiguous memory region of 32-bit words, right?  Alignment and padding cannot be checked here, but perhaps this can be documented?

dnovillo (approved):
Thanks for this!

A newbie question mostly.  I'm going to use this to pass boolean flags, so I'm wondering if it's better to encode my flags in a single uint32_t word instead of having several words in my push constants.

include/ppx/grfx/vk/vk_command.h
chaoticbob: Thanks! Missed those and the related implementation. It's in now.

chaoticbob:
> Thanks for this!
> 
> A newbie question mostly. I'm going to use this to pass boolean flags, so I'm wondering if it's better to encode my flags in a single uint32_t word instead of having several words in my push constants.

I think that's best left up to the developer. If you're trying to save spaces, then it would be best to encode it. But if you're just wanting to get something done and don't want to mess with masking then just burn some push constants.

include/ppx/grfx/vk/vk_command.h
chaoticbob: For BW shaders, we compile with `-fvk-use-dx-layout` on. Which means allow tight packing on everything except for arrays. There may be edge cases.

I've added note to that function in `graphics_command.h` about Vulkan packing rules.

PR #234, opened 2023-07-06
Adds Mouse free-fly camera controls

elviscapiaq:
- Unlike "First Person Shooter" games where pressing W doesn't let you move off the ground (it stays at y=0), this implementation 
  let you move off the ground (y != 0).

### DIFF
--- a/benchmarks/graphics_pipeline/main.cpp
+++ b/benchmarks/graphics_pipeline/main.cpp
@@ -45,14 +45,28 @@ public:
         BACKWARD
     };
 
-    FreeCamera(float3 eyePosition, float3 target)
+    FreeCamera(float3 eyePosition, float3 target, float theta, float phi)
     {
         mEyePosition = eyePosition;
         mTarget      = target;
+        mTheta       = theta;
+        mPhi         = phi;
     }
 
     // Moves the location of the camera in dir direction for distance units.
     void Move(MovementDirection dir, float distance);
+
+    // Changes the location where the camera is looking at by turning `deltaTheta`
+    // radians and looking up `deltaPhi` radians. `deltaTheta` is an angle in
+    // the range [0, 2pi] and `deltaPhi` is an angle in the range [0, pi].
+    void Turn(float deltaTheta, float deltaPhi);
+
+private:
+    // Spherical coordinates in world space where the camera is looking at.
+    // `mTheta` is an angle in the range [0, 2pi].
+    // `mPhi` is an angle in the range [0, pi].
+    float mTheta;
+    float mPhi;
 };
 
 #if defined(USE_DX12)
@@ -95,13 +109,14 @@ class ProjApp
 {
 public:
     ProjApp()
-        : mCamera(float3(0, 0, -5), float3(0, 0, -4)) {}
+        : mCamera(float3(0, 0, -5), float3(0, 0, -4), pi<float>() / 2.0f, pi<float>() / 2.0f) {}
     virtual void InitKnobs() override;
     virtual void Config(ppx::ApplicationSettings& settings) override;
     virtual void Setup() override;
-    virtual void Render() override;
+    virtual void MouseMove(int32_t x, int32_t y, int32_t dx, int32_t dy, uint32_t buttons) override;
     virtual void KeyDown(ppx::KeyCode key) override;
     virtual void KeyUp(ppx::KeyCode key) override;
+    virtual void Render() override;
 
 private:
     struct PerFrame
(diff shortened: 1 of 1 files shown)


elviscapiaq:
This PR which adds Mouse free-fly camera controls is based on a previous PR (https://github.com/google/bigwheels/pull/222). Please take a look!

apazylbe:
https://github.com/google/bigwheels/pull/222 has been merged, feel free to rebase and mark as non-draft when this is ready for review

elviscapiaq:
I rebased it and it is ready for review now.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -240,22 +255,62 @@ class ProjApp
...
+    // Saturate mTheta values by making wrap around.
+    if (mTheta < 0) {
+        mTheta = 2 * pi<float>();
+    }
+    else if (mTheta > 2 * pi<float>()) {
+        mTheta = 0;
+    }
+
+    // mPhi is saturated by making it stop, so the world doesn't turn upside down.
+    if (mPhi < 0) {
```
Keenuts, line 305: You should leave some margin around the vertical axis:
right now, if you go down, or up, the camera reacts weirdly. Limiting the angle to some -85°+85° range should be enough for our usecase, and avoid those.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -240,22 +255,62 @@ class ProjApp
...
+            float3 unitVector = glm::normalize(SphericalToCartesian(mTheta, mPhi));
+            mEyePosition -= distance * unitVector;
             break;
+        }
+    }
+    mTarget = mEyePosition + SphericalToCartesian(mTheta, mPhi);
+    LookAt(mEyePosition, mTarget);
+}
+
+void FreeCamera::Turn(float deltaTheta, float deltaPhi)
```
Keenuts, line 291: You should add a comment saying if theta/phi is doing latitude or longitude (conventions seems to differ)

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -45,14 +45,28 @@ class FreeCamera
...
+        mTheta       = theta;
+        mPhi         = phi;
     }
 
     // Moves the location of the camera in dir direction for distance units.
     void Move(MovementDirection dir, float distance);
+
+    // Changes the location where the camera is looking at by turning `deltaTheta`
+    // radians and looking up `deltaPhi` radians. `deltaTheta` is an angle in
+    // the range [0, 2pi] and `deltaPhi` is an angle in the range [0, pi].
```
Keenuts, line 61: if you change the range, update this comment too.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -240,22 +255,62 @@ class ProjApp
...
+        mTheta = 0;
+    }
+
+    // mPhi is saturated by making it stop, so the world doesn't turn upside down.
+    if (mPhi < 0) {
+        mPhi = 0;
     }
-    mTarget = mEyePosition + float3(0, 0, 1);
-    LookAt(GetEyePosition(), GetTarget());
+    else if (mPhi > pi<float>()) {
```
Keenuts, line 308: Btw, you can use `std::clamp` for this.
https://en.cppreference.com/w/cpp/algorithm/clamp

Keenuts (commented):
Some details to make it functional.
Might need to look into setInputMode to capture the mouse, otherwise it's not usable on Linux

elviscapiaq:
Ready for review.

elviscapiaq:
Ready for review.

Keenuts (approved):

Keenuts:
Fine for me, I'll let @apazylbe handle the second approval & merging.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -45,14 +45,27 @@ class FreeCamera
         BACKWARD
     };
 
-    FreeCamera(float3 eyePosition, float3 target)
+    FreeCamera(float3 eyePosition, float3 target, float theta, float phi)
     {
         mEyePosition = eyePosition;
         mTarget      = target;
+        mTheta       = theta;
```
apazylbe, line 52: Should this assert that the values are within the range you have in the comment below? Or does it not matter?

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -45,14 +45,27 @@ class FreeCamera
         BACKWARD
     };
 
-    FreeCamera(float3 eyePosition, float3 target)
+    FreeCamera(float3 eyePosition, float3 target, float theta, float phi)
```
apazylbe, line 48: Could you add a comment of what theta and phi represent? Similar to below

apazylbe (commented):
Sorry for the delay, overall lgtm, just a couple of nits. Also just 2 comments on usability:
1) the camera in the current benchmark seems to move way too quickly when looking at existing objects
2) perhaps we need a key to stop/start tracking mouse movements, so that user can adjust debug settings or freeze what they are looking at, without it constantly changing
But both are not for this PR, just a note for future

elviscapiaq:
I added the two comments for usability as future work and also rebased this pull request with the main branch. 
It is now ready for review.

PR #241, opened 2023-07-17
Update Application to include automatic metrics

Slumpwuffle:
This change updates the application to automatically start metrics logging and produce a report if metrics are enabled (e.g. via the command line). This allows any application without specific metrics handling to get some baseline metrics information.

For more complicated use cases, classes derived from Application can override the SetupMetrics, ShutdownMetrics and UpdateMetrics methods to customize their behavior.

### DIFF
--- a/src/ppx/application.cpp
+++ b/src/ppx/application.cpp
@@ -684,24 +684,15 @@ void Application::DispatchConfig()
 
 void Application::DispatchSetup()
 {
+    SetupMetrics();
     Setup();
-}
-
-void Application::SaveMetricsReportToDisk()
-{
-    if (!mSettings.enableMetrics) {
-        return;
-    }
-
-    // Export the report from the metrics manager to the disk.
-    auto report = mMetrics.manager.CreateReport(mSettings.reportPath);
-    report.WriteToDisk(mSettings.overwriteMetricsFile);
 }
 
 void Application::DispatchShutdown()
 {
     Shutdown();
 
+    ShutdownMetrics();
     SaveMetricsReportToDisk();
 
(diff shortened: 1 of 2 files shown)


apazylbe (approved):

src/ppx/application.cpp
```diff
@@ -1305,8 +1340,9 @@ int Application::Run(int argc, char** argv)
             mAverageFrameTime = static_cast<float>(nowMs / mFrameCount);
         }
 
-        // Update the metrics.
-        UpdateMetrics();
+        // Update the metrics. This can be used for both recorded AND displayed metrics,
+        // and therefore should always be called.
+        DispatchUpdateMetrics();
```
Keenuts, line 1345: nit: why `dispatch`? I see this a LOT around BigWheels, but we just `UpdateSomething()`. Isn't dispatch used when there is some kind of scheduling involved? Or is there some English subtillity I miss?

Keenuts (approved):

src/ppx/application.cpp
Slumpwuffle: TBH I'm just copying the existing style; I agree, I think dispatch is pretty meaningless in a lot of what we're doing. I also don't know why these functions are all virtuals, when it appears they're used as the internal wrapper methods to CALL virtuals, and are never overridden by anything.

So basically I suspect this just needs some cleanup, but I wasn't willing to do that here. I figured using 'Dispatch' was enough of a flag to someone that it required the same cleanup consideration.

Keenuts: As long as we all know we do it "until we clean this up", I'm fine :)

PR #242, opened 2023-07-18
Fix build for clang 14

sjfricke:
was getting errors like

```
/include/ppx/knob.h:361:21: error: format string is not a string literal (potentially insecure) [-Werror,-Wformat-security]
        ImGui::Text(flagText.c_str());
```

on ubnutu 22.04

this fixes it and lets me build

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,6 +123,7 @@ else()
         -Wno-deprecated-copy \
         -Wno-deprecated-enum-enum-conversion \
         -Wno-pointer-bool-conversion \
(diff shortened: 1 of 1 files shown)


Keenuts:
Hello! Thanks for the contribution!
I'm not convinced this is the right way. The call-sites should be fixed instead.

sjfricke:
Fair enough, at least this is documented build issue, I just did this as a workaround locally for now

Keenuts:
Sent out https://github.com/google/bigwheels/pull/247 to fix the issue.

PR #244, opened 2023-07-18
Instantiate multiple spheres randomly

elviscapiaq:
- It was added a sphere counter to render a random number of sphere instances.
- To align with sphere and skydome indices, the nodes order in the altimeter GLTF file were updated.
- At the moment it only renders up to 500 spheres. A greater number of spheres is producing a crash (still analyzing its origin).


elviscapiaq:
I fixed an issue and now it is rendering up to 3000 spheres as expected.
Moving from draft PR. Ready for review.

elviscapiaq:
After Nathan's feedback, I decided to close this PR to divide this task in 5-6 subtasks and implement them easier.

PR #245, opened 2023-07-18
Adding more flexible specifications for commandline parsing

angela28chen:
* Changed CliOptions implementation to remove extra options. All options are stored as key (string) and value (list of strings) pairs in an unordered map.
* Some validation has been removed at the parsing stage (e.g. `--gpu -3` will parse without error, but run into a different error later)
* Added some `string_util` helpers for parsing

Expanded commandline flags syntax to allow:
* Resolution flags : `--flag-name NxM`
* no- prefix for boolean flags : `--no-flag-name`
* Other boolean values accepted
* = between flag and value : `--flag-name=parameter`

### DIFF
--- a/include/ppx/command_line_parser.h
+++ b/include/ppx/command_line_parser.h
@@ -35,7 +35,6 @@ struct StandardOptions
     bool deterministic          = false;
     bool enable_metrics         = false;
     bool headless               = false;
-    bool help                   = false;
     bool list_gpus              = false;
     bool overwrite_metrics_file = false;
     bool use_software_renderer  = false;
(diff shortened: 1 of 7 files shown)


include/ppx/command_line_parser.h
```diff
@@ -54,100 +53,114 @@ struct StandardOptions
...
-                ss.clear();
-                ss >> std::boolalpha >> val;
-                if (ss.fail()) {
-                    return defaultValue;
-                }
-            }
-            return val;
+    // Intended for list flags, specified on command line with comma-separated parameters and/or multiple instances of same flag
+    template <typename T1>
+    std::vector<T1> GetOptionValueOrDefault(const std::string& optionName, const std::vector<T1>& defaultValues) const
```
Keenuts, line 96: Those 2 functions seems weird:

For non-vector options, you still store all the values, even if only the last will be used. Why store those values since they are never used?

Feels like you should have 2 storages:
 - 1 for options that are specified once (and overridden if specified multiple times), and 1 for options that can be specified multiple times.
 
 For the GetOptionValueOrDefault, since we cannot overload just with the return value, I can imagine this 2 function to still exist. Not sure if having 2 different names is more usable or not. Maybe you and Aliya have another opinion?

Keenuts: Isn't that the current aim? I thought this was built to replace the legacy method 😀

Keenuts: I'd be in favor to drop the `GetExtraOptions()` and only rely on knobs. Silently fail when there is a typo in a flag is not a great user experience.
But yes, maybe that's too much for 1 PR, so up to you.

Do you plan to keep the unorderd_map in the long run?
Ignoring extra options, with this PR, if I do
`./binary --help=true --help=false --help=true`, you store 3 values for `help`, even if only the last one is required.
Probably not that important if fixing this makes the code less readable, just surprising.

Keenuts (commented):
A few small items and we'll be good.

include/ppx/command_line_parser.h
```diff
@@ -54,100 +54,115 @@ struct StandardOptions
...
-    const StandardOptions& GetStandardOptions() const
+    // (WILL BE DEPRECATED, USE KNOBS INSTEAD)
+    // Get the parameter value after converting it into the desired integral,
+    // floating-point, or boolean type. If the value fails to be converted,
+    // return the specified default value.
+    template <typename T>
+    T GetExtraOptionValueOrDefault(std::string_view optionName, const T& defaultValue) const
     {
-        return standardOptions;
+        static_assert(std::is_integral_v<T> || std::is_floating_point_v<T> || std::is_same_v<T, std::string>, "GetExtraOptionValueOrDefault must be called with an integral, floating-point, boolean, or std::string type");
```
Keenuts, line 125: Ah, seems like clang-format leaves this line as-is, surprising, maybe something in our config.

include/ppx/command_line_parser.h
```diff
@@ -54,100 +54,115 @@ struct StandardOptions
...
 private:
-    void AddExtraOption(const Option& opt)
-    {
-        extraOptions.insert(std::make_pair(opt.GetName(), opt));
-    }
-
-    std::unordered_map<std::string, Option> extraOptions;
-    StandardOptions                         standardOptions;
+    // All flag names (string) and parameters (vector of strings) specified on the command line
+    std::unordered_map<std::string_view, std::vector<std::string_view>> allOptions;
```
apazylbe, line 165: mAllOptions?

src/ppx/command_line_parser.cpp
```diff
@@ -37,157 +37,124 @@ bool IsOptionOrFlag(const std::string& s)
...
+    if (res == std::nullopt) {
+        PPX_LOG_ERROR("resolution flag must be in format <Width>x<Height>: " << valueStr);
+        return defaultValue;
+    }
+    int N = GetParsedOrDefault(res->first, defaultValue.first);
+    int M = GetParsedOrDefault(res->second, defaultValue.second);
+    return std::make_pair(N, M);
+}
+
+void CliOptions::AddOption(std::string_view key, std::string_view valueStr)
```
apazylbe, line 57: How about renaming key to something more descriptive to improve readability? Like optionName.

src/ppx/command_line_parser.cpp
```diff
@@ -37,157 +37,124 @@ bool IsOptionOrFlag(const std::string& s)
...
+        return true;
+    }
+    std::stringstream ss{std::string(valueStr)};
+    bool              valueAsBool;
+    ss >> valueAsBool;
+    if (ss.fail()) {
+        ss.clear();
+        ss >> std::boolalpha >> valueAsBool;
+        if (ss.fail()) {
+            return defaultValue;
```
apazylbe, line 80: Should this at least log an error instead of silently using the default?

include/ppx/string_util.h
```diff
@@ -16,14 +16,20 @@
...
 
 namespace ppx {
 namespace string_util {
 
 void TrimLeft(std::string& s);
 void TrimRight(std::string& s);
 
-std::string TrimCopy(const std::string& s);
+std::string      TrimCopy(const std::string& s);
+std::string_view TrimBothEnds(std::string_view s, std::string_view c = " \t");
```
apazylbe, line 28: Could you add a short comment for this function?

include/ppx/command_line_parser.h
```diff
@@ -54,100 +54,115 @@ struct StandardOptions
...
+public:
+    CliOptions() = default;
+
+    bool HasExtraOption(std::string_view option) const { return allOptions.contains(option); }
+
+    // Returns the number of unique options and flags that were specified on the commandline,
+    // not counting multiple appearances of the same flag such as: --assets-path a --assets-path b
+    size_t GetNumUniqueOptions() const { return allOptions.size(); }
+
+    // Try to parse option string into the type of the default value and return it.
```
apazylbe, line 82: nit: "Tries to parse", pretending the comment follows "This function..."

apazylbe (approved):
Just a few nits, otherwise lgtm! Thanks!

PR #246, opened 2023-07-19
Shutdown logging in the LogInitialized test

s-perron:
The `LogInitialized` does not use the `LogTest` class, so the fix in
https://github.com/google/bigwheels/pull/243 did not apply there. This
applies the same fix for `LogInitialized`.

### DIFF
--- a/src/test/log_console_test.cpp
+++ b/src/test/log_console_test.cpp
@@ -21,6 +21,24 @@
 
 namespace ppx {
 
+class LogStaticTest : public ::testing::Test
+{
+protected:
+    void SetUp() override
+    {
+        // Some other unrelated tests might have run before the logging tests and
+        // already initialized logging. Since we share global state and run all
+        // tests in a single process, we need to shut down any existing logging.
+        // If the logging was not initialized this operation is a no-op.
+        Log::Shutdown();
+    }
+
+    void TearDown() override
+    {
+        Log::Shutdown();
+    }
+};
+
 class LogTest : public ::testing::Test
 {
 protected:
@@ -44,7 +62,7 @@ protected:
     std::stringstream mOut;
 };
 
-TEST(LogStaticTest, LogInitialized)
+TEST_F(LogStaticTest, LogInitialized)
 {
     std::stringstream out;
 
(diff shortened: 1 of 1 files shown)


src/test/log_console_test.cpp
```diff
@@ -48,6 +48,8 @@ TEST(LogStaticTest, LogInitialized)
 {
     std::stringstream out;
 
+    // First shut down the logging in case another test has already initialized logging.
+    Log::Shutdown();
```
apazylbe, line 52: Isn't it a problem for the LogShutdown test case below as well? Perhaps it's better to create a LogStaticTest test fixture that would just call Log::Shutdown() in the SetUp?

s-perron: Right, I did not think of what would happen if the log was already initialized with a different stream. I'll try to fix it up tomorrow.

src/test/log_console_test.cpp
```diff
@@ -21,6 +21,24 @@
...
+        // Some other unrelated tests might have run before the logging tests and
+        // already initialized logging. Since we share global state and run all
+        // tests in a single process, we need to shut down any existing logging.
+        // If the logging was not initialized this operation is a no-op.
+        Log::Shutdown();
+    }
+
+    void TearDown() override
+    {
+        Log::Shutdown();
```
apazylbe, line 38: I think we don't actually need to call Log::Shutdown() since the tests do it themselves, and in one case it is even part of the test. Though I guess right now it is pretty harmless to call it again.

s-perron: I can remove it if you want. Even if it is not needed in the current test, I wanted to follow the pattern of `LogTest` where we guarantee that the log is shutdown after every test. I don't think it is a big deal either way. Tests are assuming the Log has not been shutdown, so it might be fair to not shut it down.

apazylbe: Yeah, I don't really have a strong preference either way, as you said there is an argument for either. Since it's currently this way, lgtm

apazylbe (approved):

s-perron:
@apazylbe I don't have write access. Could you merge when you get a chance? Thanks.

PR #247, opened 2023-07-20
fix warning on clang-14

Keenuts:


include/ppx/knob.h
```diff
@@ -358,7 +358,7 @@ class KnobFlag final
         std::stringstream ss;
         ss << mFlagName << ": " << mValue;
         std::string flagText = ss.str();
-        ImGui::Text(flagText.c_str());
+        ImGui::Text("%s", flagText.c_str());
```
apazylbe, line 361: Could we accomplish the same with `ImGui::TextUnformatted`?

apazylbe (approved):
thanks! lgtm, but one question

PR #252, opened 2023-07-25
Refactor the Geometry class

wangra-google:
- Add new VertexDataProcessor classes to replace the branches
- Add new GEOMETRY_VERTEX_ATTRIBUTE_LAYOUT_POSITION_PLANAR (position in binding 0, non-position data are interleaved in binding 1)
- Modify 08_basic_geometry to add the cubes for position_planar vertex attribute layout
- Add VertexLayoutTest shaders (modified version of VertexColor with additional stream of Normal)


assets/basic/shaders/VertexLayoutTest.hlsl
```diff
@@ -0,0 +1,45 @@
...
+    float3 Color : COLOR;
+};
+
+VSOutput vsmain(float4 Position
+                : POSITION,
+                  float3 Color
+                : COLOR,
+                  float3 Normal
+                : NORMAL)
+{
```
chaoticbob, line 34: Avoid running clang-format on HLSL files. We don't want the function signatures to look like this, it's jarring to read.

src/ppx/geometry.cpp
```diff
@@ -19,6 +19,334 @@
...
+        const uint32_t vertexBindingCount = GetVertexBindingCount(pGeom);
+        for (uint32_t i = 0; i < vertexBindingCount; ++i) {
+            if (GetVertexBindingAttributeCount(pGeom, i) != 1) {
+                PPX_ASSERT_MSG(false, "planar layout binding must have 1 attribute");
+                return false;
+            }
+        }
+        return true;
+    }
+    virtual Result UpdateVertexBuffer(Geometry* pGeom) override
```
chaoticbob, line 126: There should be an empty line between functions.

src/ppx/geometry.cpp
```diff
@@ -19,6 +19,334 @@
...
+    // returns false if the validation fails
+    virtual bool     Validate(Geometry* pGeom)                                        = 0;
+    virtual Result   UpdateVertexBuffer(Geometry* pGeom)                              = 0;
+    virtual uint32_t AppendVertexData(Geometry* pGeom, const TriMeshVertexData& vtx)  = 0;
+    virtual uint32_t AppendVertexData(Geometry* pGeom, const WireMeshVertexData& vtx) = 0;
+    virtual uint32_t GetVertexCount(const Geometry* pGeom)                            = 0;
+
+protected:
+    // Prevent from being deleted explicitly
+    ~IVertexDataProcessor() {}
```
chaoticbob, line 37: Missing `virtual`.

src/ppx/geometry.cpp
```diff
@@ -19,6 +19,334 @@
...
+
+    // Missing attributes will also result in NOOP.
+    template <typename T>
+    uint32_t AppendDataToVertexBuffer(Geometry* pGeom, uint32_t bufferIndex, const T& data)
+    {
+        if (bufferIndex != PPX_VALUE_IGNORED) {
+            pGeom->mVertexBuffers[bufferIndex].Append(data);
+            return pGeom->mVertexBuffers[bufferIndex].GetElementCount();
+        }
+        return PPX_VALUE_IGNORED;
```
chaoticbob, line 51: Does this match an existing behavior? `PPX_VALUE_IGNORED` is 0xFFFFFFFF. I don't see any checks against. Can you give some details on the intent of the return value if it doesn't match an existing behavior?

src/ppx/geometry.cpp
```diff
@@ -19,6 +19,334 @@
...
+            // clang-format off
+        switch (semantic) {
+            default: break;
+            case grfx::VERTEX_SEMANTIC_POSITION  : AppendDataToVertexBuffer(pGeom, kBufferIndex, vtx.position); break;
+            case grfx::VERTEX_SEMANTIC_NORMAL    : AppendDataToVertexBuffer(pGeom, kBufferIndex, vtx.normal); break;
+            case grfx::VERTEX_SEMANTIC_COLOR     : AppendDataToVertexBuffer(pGeom, kBufferIndex, vtx.color); break;
+            case grfx::VERTEX_SEMANTIC_TANGENT   : AppendDataToVertexBuffer(pGeom, kBufferIndex, vtx.tangent); break;
+            case grfx::VERTEX_SEMANTIC_BITANGENT : AppendDataToVertexBuffer(pGeom, kBufferIndex, vtx.bitangent); break;
+            case grfx::VERTEX_SEMANTIC_TEXCOORD  : AppendDataToVertexBuffer(pGeom, kBufferIndex, vtx.texCoord); break;
+        }
```
chaoticbob, line 201: Should be indented to align with the clang-format tags.

src/ppx/geometry.cpp
```diff
@@ -19,6 +19,334 @@
...
+        uint32_t       startSize = GetVertexBufferSize(pGeom, kBufferIndex);
+        const uint32_t attrCount = GetVertexBindingAttributeCount(pGeom, kBufferIndex);
+        for (uint32_t attrIndex = 0; attrIndex < attrCount; ++attrIndex) {
+            const grfx::VertexSemantic semantic = GetVertexBindingAttributeSematic(pGeom, kBufferIndex, attrIndex);
+
+            // clang-format off
+        switch (semantic) {
+            default: break;
+            case grfx::VERTEX_SEMANTIC_POSITION  : AppendDataToVertexBuffer(pGeom, kBufferIndex, vtx.position); break;
+            case grfx::VERTEX_SEMANTIC_COLOR     : AppendDataToVertexBuffer(pGeom, kBufferIndex, vtx.color); break;
```
chaoticbob, line 222: Needs indention to match clang-format tags.

src/ppx/geometry.cpp
```diff
@@ -19,6 +19,334 @@
...
+            case grfx::VERTEX_SEMANTIC_TANGENT   : AppendDataToVertexBuffer(pGeom, kBufferIndex, vtx.tangent); break;
+            case grfx::VERTEX_SEMANTIC_BITANGENT : AppendDataToVertexBuffer(pGeom, kBufferIndex, vtx.bitangent); break;
+            case grfx::VERTEX_SEMANTIC_TEXCOORD  : AppendDataToVertexBuffer(pGeom, kBufferIndex, vtx.texCoord); break;
+        }
+            // clang-format on
+        }
+        uint32_t endSize = GetVertexBufferSize(pGeom, kBufferIndex);
+
+        uint32_t bytesWritten = (endSize - startSize);
+        PPX_ASSERT_MSG((bytesWritten == GetVertexBufferElementSize(pGeom, kBufferIndex)), "size of vertex data written does not match buffer's element size");
```
chaoticbob, line 207: It would be useful to put the value of `GetVertexBufferElementSize()` in a const var and have it written out in the assert message along with `bytesWritten` to make debugging easier.

src/ppx/geometry.cpp
```diff
@@ -19,6 +19,334 @@
...
+        for (uint32_t i = 0; i < attributeCount; ++i) {
+            const grfx::VertexSemantic semantic = GetVertexBindingAttributeSematic(pGeom, kNonPositionBufferIndex, i);
+            // clang-format off
+            switch (semantic) {
+                default: {
+                    return Result::ERROR_GEOMETRY_INVALID_VERTEX_SEMANTIC;
+                }
+
+                case grfx::VERTEX_SEMANTIC_POSITION  : 
+                    PPX_ASSERT_MSG(false, "position should be in binding 0"); break;
```
chaoticbob, line 269: This should be on the same line as the case statement if the switch block is surrounded by clang-format tags.

chaoticbob (changes requested):
Some nits and questions.

src/ppx/geometry.cpp
wangra-google: This is trying to match the behavior of the original function Geometry::AppendPosition(), but since now it is a templated function, the ones for normal, color, tangent, bitangent and texcoord are also returning `ElementCount` (if attribute exist) or `PPX_VALUE_IGNORED` (if the vertex buffer doesn't have that attribute). I could add assert to the return value of position:
```
uint32_t n = AppendDataToVertexBuffer(pGeom, GetPositionBufferIndex(pGeom), vtx.position);
PPX_ASSERT_MSG(PPX_VALUE_IGNORED != n, "position should always available");
```
What do you think?

chaoticbob: Thanks for the clarification. Had my suspicion that was the case, but wanted to be sure. Neither `PPX_VALUE_IGNORED` or `0` are good return values for these. The intent of the return value is that if it's a valid value, then it could be used for something like calculating indices for a triangle. But both`PPX_VALUE_IGNORED` and `0` can result in incorrect indices. 

Can add a comment to both internal and external functions that `PPX_VALUE_IGNORED` is returned if the bufferIndex is invalid? Maybe also add a range check for `bufferIndex` instead of just against `PPX_VALUE_IGNORED`?

src/ppx/geometry.cpp
```diff
@@ -19,6 +19,334 @@
 
 namespace ppx {
 
+// -------------------------------------------------------------------------------------------------
+// VertexDataProcessor
+// -------------------------------------------------------------------------------------------------
+class IVertexDataProcessor
```
apazylbe, line 25: Are you adding "I" because it is an abstract class? I don't think any other classes follow this naming convention in BigWheels

src/ppx/geometry.cpp
```diff
@@ -19,6 +19,334 @@
 
 namespace ppx {
 
+// -------------------------------------------------------------------------------------------------
+// VertexDataProcessor
```
apazylbe, line 23: Could you add a comment for the class?

src/ppx/geometry.cpp
```diff
@@ -19,6 +19,334 @@
...
+// -------------------------------------------------------------------------------------------------
+class IVertexDataProcessor
+{
+public:
+    // returns false if the validation fails
+    virtual bool     Validate(Geometry* pGeom)                                        = 0;
+    virtual Result   UpdateVertexBuffer(Geometry* pGeom)                              = 0;
```
apazylbe, line 30: Could you add comments for these functions (including what they return)?

assets/basic/shaders/VertexLayoutTest.hlsl
wangra-google: fixed

src/ppx/geometry.cpp
wangra-google: fixed

src/ppx/geometry.cpp
wangra-google: fixed

src/ppx/geometry.cpp
wangra-google: change the return value to 0 for invalid case with comments, and added the range check as we discussed.

src/ppx/geometry.cpp
wangra-google: fixed

src/ppx/geometry.cpp
wangra-google: fixed

src/ppx/geometry.cpp
wangra-google: fixed

src/ppx/geometry.cpp
wangra-google: fixed

assets/basic/shaders/VertexLayoutTest.hlsl
```diff
@@ -0,0 +1,45 @@
+// Copyright 2022 Google LLC
```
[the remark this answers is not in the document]
wangra-google: fixed

projects/08_basic_geometry/README.md
```diff
@@ -15,4 +15,4 @@ Cube geometry data (vertices, indices and vertex colors) is generated at setup t
 
 Shader              | Purpose for this project
 ------------------- | ----------------------------------------------
-`VertexColors.hlsl` | Transform and draw a cube using vertex colors.
+`VertexLayoutTest.hlsl.hlsl` | Transform and draw a cube using vertex colors and normal as input.
```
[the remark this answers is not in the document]
wangra-google: oops, fixed

src/ppx/geometry.cpp
wangra-google: changed the class name to VertexDataProcessorBase

src/ppx/geometry.cpp
wangra-google: added comment as suggested

src/ppx/geometry.cpp
```diff
@@ -13,12 +13,392 @@
...
+        return GetVertexBufferElementCount(pGeom, kBufferIndex);
+    }
+
+    virtual uint32_t GetVertexCount(const Geometry* pGeom) override
+    {
+        return GetVertexBufferElementCount(pGeom, kBufferIndex);
+    }
+
+private:
+    const uint32_t kBufferIndex = 0;
```
chaoticbob, line 282: Does this refer to a specific buffer? Since the word buffer is used quite a lot in these classes can you add a comment on what this is used for?

src/ppx/geometry.cpp
```diff
@@ -13,12 +13,392 @@
...
+    }
+
+private:
+    const uint32_t kPositionBufferIndex    = 0;
+    const uint32_t kNonPositionBufferIndex = 1;
+};
+
+static VertexDataProcessorPlanar         sVDProcessorPlanar;
+static VertexDataProcessorInterleaved    sVDProcessorInterleaved;
+static VertexDataProcessorPositionPlanar sVDProcessorPositionPlanar;
```
chaoticbob, line 400: Not really an issue with the current design, but if data members ever got added to any of these classes or the base class - it could create problems for multithreaded cases of multiple Geometry objects. Might be useful to have a comment on the classes to warn of this for any future modifications.

src/ppx/geometry.cpp
chaoticbob: Oh I see, there's only one buffer in this class.

src/ppx/geometry.cpp
wangra-google: added some comments at the beginning of this file before the declaration of  `VertexDataProcessorBase`

chaoticbob (approved):
Looks good from my end.

src/ppx/geometry.cpp
```diff
@@ -13,12 +13,396 @@
...
+//     be careful when adding data members to any of these classes
+//     it could create problems for multithreaded cases for multiple geometry objects
+// -------------------------------------------------------------------------------------------------
+class VertexDataProcessorBase
+{
+public:
+    // validates the layout
+    // returns false if the validation fails
+    virtual bool Validate(Geometry* pGeom) = 0;
+    // update the vertex buffer and the vertex buffer index
```
apazylbe, line 37: Updates?
Assuming the function comment was followed by "This function..." 
and same for the next function

src/ppx/geometry.cpp
```diff
@@ -647,13 +1041,7 @@ uint32_t Geometry::GetIndexCount() const
...
 uint32_t Geometry::GetVertexCount() const
 {
-    uint32_t count = 0;
-    if (mCreateInfo.vertexAttributeLayout == ppx::GEOMETRY_VERTEX_ATTRIBUTE_LAYOUT_INTERLEAVED) {
-        count = mVertexBuffers[0].GetElementCount();
-    }
-    else if (mCreateInfo.vertexAttributeLayout == ppx::GEOMETRY_VERTEX_ATTRIBUTE_LAYOUT_PLANAR) {
-        count = mVertexBuffers[mPositionBufferIndex].GetElementCount();
-    }
+    uint32_t count = (mVDProcessor != nullptr) ? mVDProcessor->GetVertexCount(this) : 0;
```
apazylbe, line 1044: Just curious, is it valid for mVDProcessor to be nullptr here?

apazylbe (approved):
Just some nits and question. Otherwise lgtm!

PR #253, opened 2023-07-25
Change metrics to write to the external storage.

at-ninja:
External storage is easy to read and write to over `adb` commands, and doesn't require the APK to be built in `debug` configurations to utilize the metrics.

### DIFF
--- a/include/ppx/fs.h
+++ b/include/ppx/fs.h
@@ -140,7 +140,7 @@ bool path_exists(const std::filesystem::path& path);
 std::filesystem::path GetInternalDataPath();
 
 // Returns a path to the application's external data directory (can be used for output).
-std::filesystem::path GetInternalDataPath();
+std::filesystem::path GetExternalDataPath();
 #endif
 
 } // namespace ppx::fs
--- a/include/ppx/fs.h
+++ b/include/ppx/fs.h
@@ -138,6 +138,9 @@ bool path_exists(const std::filesystem::path& path);
 // Returns a path to the application's internal data directory (can be used for output).
 // NOTE: The internal data path on Android is extremely limited in terms of filesize!
 std::filesystem::path GetInternalDataPath();
+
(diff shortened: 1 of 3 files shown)


include/ppx/fs.h
```diff
@@ -138,6 +138,9 @@ bool path_exists(const std::filesystem::path& path);
 // Returns a path to the application's internal data directory (can be used for output).
 // NOTE: The internal data path on Android is extremely limited in terms of filesize!
 std::filesystem::path GetInternalDataPath();
+
+// Returns a path to the application's external data directory (can be used for output).
+std::filesystem::path GetExternalDataPath();
```
apazylbe, line 143: Do you know if we need to add external permission in the manifest file to do this?

at-ninja: No, new permissions are not needed, as long as access is restricted specifically to the scoped storage folder allocated by Android.

From [the documentation](https://developer.android.com/about/versions/10/privacy/changes#scoped-storage):

> By default, apps targeting Android 10 and higher are given [scoped access into external storage](https://developer.android.com/training/data-storage/files/external-scoped), or scoped storage. Such apps can see the following types of files within an external storage device without needing to request any storage-related user permissions:
> * Files in the app-specific directory, accessed using [getExternalFilesDir()](https://developer.android.com/reference/android/content/Context#getExternalFilesDir(java.lang.String)).
> * Photos, videos, and audio clips that the app created from the [media store](https://developer.android.com/training/data-storage/files/media).

apazylbe: I see, thanks!

apazylbe (approved):

apazylbe:
Adding @Slumpwuffle who (I think) authored the original reference to internal path, just in case there was a specific reason for using that

Keenuts (approved):

PR #254, opened 2023-07-25
Minor knob changes (Aligned usage messages and float slider knobs)

angela28chen:
* Flag descriptions in usage message are now aligned (Adding string_util for aligning)
* Adding float slider knobs

### DIFF
--- a/src/ppx/string_util.cpp
+++ b/src/ppx/string_util.cpp
@@ -79,19 +79,16 @@ std::string WrapText(const std::string& s, size_t width, size_t indent)
     size_t      textWidth   = width - indent;
     std::string wrappedText = "";
     while (remainingString != "") {
+        // Section off the next line from the remaining string, format it, and append it to wrappedText
         size_t lineLength = remainingString.find_last_of(" \t", textWidth);
         if (lineLength == std::string::npos) {
-            lineLength = textWidth;
+            lineLength = std::min(textWidth, remainingString.length());
         }
-
         std::string newLine = remainingString.substr(0, lineLength);
         TrimRight(newLine);
-
         wrappedText += std::string(indent, ' ') + newLine + "\n";
-        if (lineLength >= remainingString.length()) {
-            break;
-        }
 
+        // Update the remaining string
         remainingString = remainingString.substr(lineLength);
         TrimLeft(remainingString);
     }


dnovillo (approved):

src/ppx/knob.cpp
```diff
@@ -111,18 +112,25 @@ void KnobManager::DrawAllKnobs(bool inExistingWindow)
 
 std::string KnobManager::GetUsageMsg()
 {
-    std::string usageMsg = "\nApplication-Specific Flags:\n";
-    // Tends to have longer params, do not attempt to align description
-    // --flag_name <params> : description
+    std::string usageMsg    = "\nApplication-Specific Flags:\n";
+    size_t      indentWidth = 20;
+    size_t      totalWidth  = 80;
```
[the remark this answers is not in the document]
dnovillo: Yeah, you're absolutely right.   I suspect we'll never want to actually read from the tty to set this.  But, then again, we may 😄

src/test/string_util_test.cpp
```diff
@@ -93,4 +93,74 @@ TEST(StringUtilTest, SplitInTwoMultipleDelimiter)
...
+    std::string wantWrap = R"(The quick
+brown fox
+jumps over
+the
+extremely-
+long-word-
+here lazy
+dog.
+)";
+    std::string wrapped  = WrapText(toWrap, 10, 0);
```
apazylbe, line 164: Some tests with tabs?

src/ppx/string_util.cpp
```diff
@@ -68,5 +69,37 @@ std::optional<std::pair<std::string_view, std::string_view>> SplitInTwo(std::str
...
+        if (remainingString.length() > textWidth) {
+            nextChar = remainingString.at(textWidth);
+        }
+
+        auto lastSpace = newLine.find_last_of(" \t");
+
+        if ((nextChar != ' ') && (lastSpace != std::string::npos)) {
+            // The line break is interrupting a word that can go onto the next line
+            newLine = remainingString.substr(0, lastSpace);
+        }
```
apazylbe, line 94: Could we simplify the above to this? 
```
size_t lineLength = remainingString.find_last_of(" \t", textWidth);
if (lineLength == std::string::npos) {
    lineLength = textWidth;
}
std::string newLine = remainingString.substr(0, lineLength);
```

src/ppx/knob.cpp
```diff
@@ -13,11 +13,21 @@
...
 namespace ppx {
 
+// Spacing:
+//
+// --flag_name <params>    description...
+//                         continued description...
+// |usageMsgIndentWidth---|
+// |usageMsgTotalWidth-----------------------------|
+constexpr size_t usageMsgIndentWidth = 20;
+constexpr size_t usageMsgTotalWidth  = 80;
```
apazylbe, line 29: Could you add "k" prefix to the const? 
kUsageMsgIndentWidth and kUsageMsgTotalWidth

src/ppx/string_util.cpp
```diff
@@ -68,5 +69,34 @@ std::optional<std::pair<std::string_view, std::string_view>> SplitInTwo(std::str
     return std::make_pair(firstSubstring, secondSubstring);
 }
 
+std::string WrapText(const std::string& s, size_t width, size_t indent = 0)
```
apazylbe, line 72: Could you move the default value to the function declaration in the header file?

src/ppx/string_util.cpp
```diff
@@ -68,5 +69,34 @@ std::optional<std::pair<std::string_view, std::string_view>> SplitInTwo(std::str
...
+        size_t lineLength = remainingString.find_last_of(" \t", textWidth);
+        if (lineLength == std::string::npos) {
+            lineLength = textWidth;
+        }
+
+        std::string newLine = remainingString.substr(0, lineLength);
+        TrimRight(newLine);
+
+        wrappedText += std::string(indent, ' ') + newLine + "\n";
+        if (lineLength >= remainingString.length()) {
```
apazylbe, line 91: I am curious, do you need this check or can you let it break when it reaches the while condition above?

apazylbe (commented):
Just a couple of things, otherwise lgtm!

src/ppx/string_util.cpp
apazylbe: Ah right, I think line 84 should probably be
lineLength = std::max(textWidth, remainingString.length());
and then you wouldn't need this check and wouldn't rely on the substr above being fine with more characters requested than exist

apazylbe: yes, right! Thanks!

apazylbe (approved):

PR #256, opened 2023-07-27
Swapchain cleanup.

hysw:
Use common initialization for dx/vk/xr.
Remove some dead code.

### DIFF
--- a/include/ppx/grfx/grfx_swapchain.h
+++ b/include/ppx/grfx/grfx_swapchain.h
@@ -178,13 +178,16 @@ protected:
     friend class grfx::Device;
 
     // Make these protected since D3D12's swapchain resize will need to call them
+    Result CreateColorImages();
+    Result CreateColorImagesFromApiObjects(const std::vector<void*>& apiObjects);
     void   DestroyColorImages();
     Result CreateDepthImages();
+    Result CreateDepthImagesFromApiObjects(const std::vector<void*>& apiObjects);
     void   DestroyDepthImages();
     Result CreateRenderPasses();
     void   DestroyRenderPasses();
-    Result CreateRenderTargets();
-    void   DestroyRenderTargets();
+    Result CreateRenderTargetViews();
+    void   DestroyRenderTargetViews();
 
 private:
     virtual Result AcquireNextImageInternal(
@@ -211,8 +214,7 @@ private:
 
     std::vector<grfx::CommandBufferPtr> mHeadlessCommandBuffers;
 
-protected:
-    grfx::QueuePtr                         mQueue;
+private:
     std::vector<grfx::ImagePtr>            mDepthImages;
     std::vector<grfx::ImagePtr>            mColorImages;
     std::vector<grfx::RenderTargetViewPtr> mClearRenderTargets;
@@ -221,6 +223,7 @@ protected:
     std::vector<grfx::RenderPassPtr>       mClearRenderPasses;
     std::vector<grfx::RenderPassPtr>       mLoadRenderPasses;
 
+protected:
 #if defined(PPX_BUILD_XR)
     XrSwapchain mXrColorSwapchain = XR_NULL_HANDLE;
     XrSwapchain mXrDepthSwapchain = XR_NULL_HANDLE;
--- a/src/ppx/grfx/dx12/dx12_swapchain.cpp
+++ b/src/ppx/grfx/dx12/dx12_swapchain.cpp
@@ -133,9 +133,6 @@ Result Swapchain::CreateApiObjects(const grfx::SwapchainCreateInfo* pCreateInfo)
             return ERROR_GRFX_UNSUPPORTED_SWAPCHAIN_FORMAT;
         }
 
-        // Cache color format
-        mColorFormat = pCreateInfo->colorFormat;
-
         // Make swapchain waitable
         mFlags = DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT;
 
@@ -178,7 +175,7 @@ Result Swapchain::CreateApiObjects(const grfx::SwapchainCreateInfo* pCreateInfo)
         DXGI_SWAP_CHAIN_DESC1 dxDesc = {};
         dxDesc.Width                 = static_cast<UINT>(pCreateInfo->width);
         dxDesc.Height                = static_cast<UINT>(pCreateInfo->height);
-        dxDesc.Format                = dx::ToDxgiFormat(mColorFormat);
+        dxDesc.Format                = dxgiFormat;
         dxDesc.Stereo                = FALSE;
         dxDesc.SampleDesc.Count      = 1;
         dxDesc.SampleDesc.Quality    = 0;
@@ -246,27 +243,23 @@ Result Swapchain::CreateApiObjects(const grfx::SwapchainCreateInfo* pCreateInfo)
 
     // Create color images
     {
-        auto ppxres = CreateColorImages(pCreateInfo->width, pCreateInfo->height, pCreateInfo->colorFormat, colorImages);
+        std::vector<void*> apiObjects;
+        for (const auto& colorImage : colorImages) {
+            apiObjects.push_back(colorImage);
+        }
+        auto ppxres = CreateColorImagesFromApiObjects(apiObjects);
         if (Failed(ppxres)) {
             return ppxres;
         }
     }
 
 #if defined(PPX_BUILD_XR)
-    {
-        for (size_t i = 0; i < depthImages.size(); ++i) {
-            grfx::ImageCreateInfo imageCreateInfo = grfx::ImageCreateInfo::DepthStencilTarget(pCreateInfo->width, pCreateInfo->height, pCreateInfo->depthFormat, grfx::SAMPLE_COUNT_1);
-            imageCreateInfo.pApiObject            = depthImages[i];
-
-            grfx::ImagePtr image;
-            Result         ppxres = GetDevice()->CreateImage(&imageCreateInfo, &image);
-            if (Failed(ppxres)) {
-                PPX_ASSERT_MSG(false, "image create failed");
-                return ppxres;
-            }
-
-            mDepthImages.push_back(image);
+    if (!depthImages.empty()) {
+        std::vector<void*> apiObjects;
+        for (const auto& depthImage : depthImages) {
+            apiObjects.push_back(depthImage);
         }
+        CreateDepthImagesFromApiObjects(apiObjects);
     }
 #endif
 
@@ -287,38 +280,6 @@ void Swapchain::DestroyApiObjects()
     if (mQueue) {
         mQueue.Reset();
     }
-}
-
-Result Swapchain::CreateColorImages(uint32_t width, uint32_t height, grfx::Format format, const std::vector<ID3D12Resource*>& colorImages)
-{
-    for (size_t i = 0; i < colorImages.size(); ++i) {
-        grfx::ImageCreateInfo imageCreateInfo           = {};
-        imageCreateInfo.type                            = grfx::IMAGE_TYPE_2D;
-        imageCreateInfo.width                           = width;
-        imageCreateInfo.height                          = height;
-        imageCreateInfo.depth                           = 1;
-        imageCreateInfo.format                          = format;
-        imageCreateInfo.sampleCount                     = grfx::SAMPLE_COUNT_1;
-        imageCreateInfo.mipLevelCount                   = 1;
-        imageCreateInfo.arrayLayerCount                 = 1;
-        imageCreateInfo.usageFlags.bits.transferSrc     = true;
-        imageCreateInfo.usageFlags.bits.transferDst     = true;
-        imageCreateInfo.usageFlags.bits.sampled         = true;
-        imageCreateInfo.usageFlags.bits.storage         = true;
-        imageCreateInfo.usageFlags.bits.colorAttachment = true;
-        imageCreateInfo.pApiObject                      = colorImages[i];
-
-        grfx::ImagePtr image;
-        Result         ppxres = GetDevice()->CreateImage(&imageCreateInfo, &image);
-        if (Failed(ppxres)) {
-            PPX_ASSERT_MSG(false, "image create failed");
-            return ppxres;
-        }
-
-        mColorImages.push_back(image);
-    }
-
-    return ppx::SUCCESS;
 }
 
 Result Swapchain::AcquireNextImageInternal(
@@ -402,15 +363,9 @@ Result Swapchain::Resize(uint32_t width, uint32_t height)
     mCreateInfo.width  = width;
     mCreateInfo.height = height;
 
-    std::vector<ID3D12Resource*> colorImages;
-    for (auto& im : mColorImages) {
-        auto pRes = static_cast<dx12::Image*>(im.Get())->GetDxResource();
-        colorImages.push_back(pRes);
-    }
-
     // Destroy these to make sure there's no reference before resizing
     DestroyRenderPasses();
-    DestroyRenderTargets();
+    DestroyRenderTargetViews();
     DestroyDepthImages();
     DestroyColorImages();
 
@@ -436,30 +391,39 @@ Result Swapchain::Resize(uint32_t width, uint32_t height)
         }
 
         // Grab the texture resources created by the swapchain.
-        std::vector<ID3D12Resource*> colorImages;
+        std::vector<void*> apiObjects;
         for (UINT i = 0; i < desc.BufferCount; ++i) {
             D3D12ResourcePtr resource;
             HRESULT          hr = mSwapchain->GetBuffer(i, IID_PPV_ARGS(&resource));
             if (FAILED(hr)) {
                 return ppx::ERROR_API_FAILURE;
             }
-            colorImages.push_back(resource);
+            apiObjects.push_back(resource);
         }
 
-        auto ppxres = CreateColorImages(width, height, mColorFormat, colorImages);
+        Result ppxres = CreateColorImagesFromApiObjects(apiObjects);
         if (Failed(ppxres)) {
             return ppxres;
         }
     }
 
     // Create depth images
-    CreateDepthImages();
+    Result ppxres = CreateDepthImages();
+    if (Failed(ppxres)) {
+        return ppxres;
+    }
 
     // Create render targets
-    CreateRenderTargets();
+    ppxres = CreateRenderTargetViews();
+    if (Failed(ppxres)) {
+        return ppxres;
+    }
 
     // Create render passes
-    CreateRenderPasses();
+    ppxres = CreateRenderPasses();
+    if (Failed(ppxres)) {
+        return ppxres;
+    }
 
     return ppx::SUCCESS;
 }
--- a/src/ppx/grfx/grfx_swapchain.cpp
+++ b/src/ppx/grfx/grfx_swapchain.cpp
@@ -50,26 +50,9 @@ Result Swapchain::Create(const grfx::SwapchainCreateInfo* pCreateInfo)
 
     // Create color images if needed. This is only needed if we're creating
     // a headless swapchain.
-    if (mColorImages.empty()) {
-        for (uint32_t i = 0; i < mCreateInfo.imageCount; ++i) {
-            grfx::ImageCreateInfo rtCreateInfo = ImageCreateInfo::RenderTarget2D(mCreateInfo.width, mCreateInfo.height, mCreateInfo.colorFormat);
-            rtCreateInfo.ownership             = grfx::OWNERSHIP_RESTRICTED;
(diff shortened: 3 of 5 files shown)


hysw:
Mark as draft since I'm still need to verified XR builds.

src/ppx/grfx/grfx_swapchain.cpp
```diff
@@ -361,15 +509,38 @@ Result Swapchain::Present(
         return PresentHeadless(imageIndex, waitSemaphoreCount, ppWaitSemaphores);
     }
 
-    return PresentInternal(imageIndex, waitSemaphoreCount, ppWaitSemaphores);
+    if (IsIndirect()) {
+        RecordPreamble(imageIndex);
+    }
+    uint32_t                      nextWaitSemaphoreCount = waitSemaphoreCount;
+    const grfx::Semaphore* const* ppNextWaitSemaphores   = ppWaitSemaphores;
+    if (mIsRecording[imageIndex]) {
```
wangra-google, line 517: could you add some comments to explain what you want to achieve here? if `IsIndirect()` is false, `mIsRecording[imageIndex]` will always be false?

src/ppx/grfx/grfx_swapchain.cpp
```diff
@@ -402,5 +573,93 @@ Result Swapchain::PresentHeadless(uint32_t imageIndex, uint32_t waitSemaphoreCou
     return ppx::SUCCESS;
 }
 
+void Swapchain::RecordUI(uint32_t imageIndex, std::function<void(grfx::CommandBufferPtr)> f)
```
wangra-google, line 576: A bit confused here, why is RecordUI part of Swapchain?

include/ppx/grfx/grfx_swapchain.h
```diff
@@ -120,21 +121,26 @@ class Swapchain
     Swapchain() {}
     virtual ~Swapchain() {}
 
+    bool         IsIndirect() const { return mIsIndirect; }
```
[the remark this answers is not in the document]
hysw: It's not rendering directly to swapchain image provided by system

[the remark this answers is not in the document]
hysw: So the mode of operation
- Direct: app -> system swapchain
- Render at different (i.e. larger) resolution than system allows: app -> indirect -> system swapchain
- Render headless: app -> headless -> discard

include/ppx/grfx/grfx_swapchain.h
```diff
@@ -172,12 +180,27 @@ class Swapchain
...
+        void   DestroyColorImages(grfx::Device&);
+
+        Result WrapDepthImages(grfx::Device&, grfx::Format format, const std::vector<void*>& handles);
+        Result CreateDepthImages(grfx::Device&, grfx::Format format, uint32_t imageCount);
+        void   DestroyDepthImages(grfx::Device&);
+
+        Result CreateRenderPasses(grfx::Device&);
+        void   DestroyRenderPasses(grfx::Device&);
+
+        std::vector<grfx::ImagePtr>      depthImages;
```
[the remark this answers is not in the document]
hysw: Each one is corresponding to a color image.

src/ppx/grfx/grfx_swapchain.cpp
hysw: Not necessarily. It's intended to keep track if there's any post processing (e.g. imgui).

include/ppx/grfx/grfx_swapchain.h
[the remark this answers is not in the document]
wangra-google: I can understand multiple colors buffers so that when the GPU is writing to one, the other one can be used for present. But we never need to queue depth for present, so wondering why do we need multiple depth

src/ppx/grfx/grfx_swapchain.cpp
hysw: Currently ImGui is rendered in the middle of overall rendering process, which means any timing of command buffer will include the ImGui, and if we set scissor or render to a larger than screen buffer it will be truncated or have incorrect size.
Logically it should not be a part of Swapchain, and be part of application post render processing step, however it happens between app finish rendering (presenting) and send the image to display, which currently all resides in Swapchain. It's a way larger refactoring to make render not operate directly on a swapchain :(

include/ppx/grfx/grfx_swapchain.h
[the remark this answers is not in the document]
hysw: We do queue depth for present, see [src/ppx/xr_component.cpp:535] .

[src/ppx/xr_component.cpp:535]: https://github.com/google/bigwheels/tree/bc40a927267f83ee53399d69c8fe85f345b0f6d7/src/ppx/xr_component.cpp#535

src/ppx/grfx/grfx_swapchain.cpp
wangra-google: i couldn't find where is this function called, but i guess it is supposed to be called in the Swapchain::Present? I am wondering would it be cleaner if we keep a list of callback functions as member of the swapchain, and let the application to register callbacks (for example, UI callback), and we check the size of the registered callbacks, and only call this function (probably rename it to something else, something like `ExecuteCallbacksAfterBlit`) when the size of the callbacks are not 0?

hysw: It's not... yet, I'm rebasing a [giant PR](https://github.com/hysw/bigwheels/tree/tianc-vswap-offscreen) I had previously.
Set a callback by application is a better idea, except I need to re-architect how UI is drawn (which is currently in the middle of frame) and also make sure it does not break XR, which will take a bit more changes.

include/ppx/grfx/grfx_swapchain.h
```diff
@@ -202,14 +228,30 @@ class Swapchain
         uint32_t                      waitSemaphoreCount,
         const grfx::Semaphore* const* ppWaitSemaphores);
 
-    std::vector<grfx::CommandBufferPtr> mHeadlessCommandBuffers;
+    // RecordPreamble for copying intermediate frame buffer to device frame buffer.
+    void RecordPreamble(uint32_t imageIndex);
+
+    std::vector<grfx::SemaphorePtr>     mPostProcessSemaphores;
```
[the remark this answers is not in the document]
hysw: There's also copying the intermediate buffer to real device swapchain.

src/ppx/grfx/grfx_swapchain.cpp
```diff
@@ -402,5 +573,93 @@ Result Swapchain::PresentHeadless(uint32_t imageIndex, uint32_t waitSemaphoreCou
...
+        {
+            grfx::ImageToImageCopyInfo imcopy = {};
+
+            imcopy.extent.x = std::min(mCreateInfo.width, mIndirectTarget.width);
+            imcopy.extent.y = std::min(mCreateInfo.height, mIndirectTarget.height);
+
+            // Copy rendered image.
+            // Note(tianc): this should be a image blit instead of copy.
+            commandBuffer->TransitionImageLayout(srcImage, PPX_ALL_SUBRESOURCES, grfx::RESOURCE_STATE_PRESENT, grfx::RESOURCE_STATE_COPY_SRC);
+            commandBuffer->CopyImageToImage(&imcopy, srcImage, dstImage);
```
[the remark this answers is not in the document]
hysw: `vkCmdCopyImage` does not do scaling, but that's common one between vk and dx. `vkCmdBlitImage ` does not have a dx12 equivalent, so it will require more changes. This at least makes rendering at 4k possible on smaller screen without using headless, albeit cropped .

src/ppx/grfx/grfx_swapchain.cpp
```diff
@@ -402,5 +573,93 @@ Result Swapchain::PresentHeadless(uint32_t imageIndex, uint32_t waitSemaphoreCou
...
+        commandBuffer->TransitionImageLayout(dstImage, PPX_ALL_SUBRESOURCES, grfx::RESOURCE_STATE_PRESENT, grfx::RESOURCE_STATE_RENDER_TARGET);
+        {
+            // Clear screen.
+            grfx::RenderPassPtr renderPass = dstClear;
+
+            grfx::RenderPassBeginInfo beginInfo = {};
+            beginInfo.pRenderPass               = renderPass;
+            beginInfo.renderArea                = renderPass->GetRenderArea();
+            beginInfo.RTVClearCount             = 1;
+            beginInfo.RTVClearValues[0]         = {{0.5, 0.5, 0.5, 0}};
```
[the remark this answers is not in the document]
hysw: No specific reason, it makes it possible to distinguish which region is copied from intermediate (which will be black) and which part is not.

src/ppx/grfx/grfx_swapchain.cpp
```diff
@@ -402,5 +573,93 @@ Result Swapchain::PresentHeadless(uint32_t imageIndex, uint32_t waitSemaphoreCou
...
+    grfx::CommandBufferPtr commandBuffer = mCommandBuffers[mCurrentImageIndex];
+    grfx::ImagePtr         image         = mDeviceTarget.colorImages[imageIndex];
+    grfx::RenderPassPtr    renderPass    = mDeviceTarget.loadRenderPasses[imageIndex];
+
+    commandBuffer->TransitionImageLayout(image, PPX_ALL_SUBRESOURCES, grfx::RESOURCE_STATE_PRESENT, grfx::RESOURCE_STATE_RENDER_TARGET);
+    {
+        grfx::RenderPassBeginInfo beginInfo = {};
+        beginInfo.pRenderPass               = renderPass;
+        beginInfo.renderArea                = renderPass->GetRenderArea();
+        beginInfo.RTVClearCount             = 1;
```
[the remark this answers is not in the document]
hysw: This does not actually clear anything, the render pass is a `LOAD` pass, some part of BigWheels does not like 0 clear count previously, and I've yet to check if it's fixed on all apis and all platform.

src/ppx/grfx/grfx_swapchain.cpp
wangra-google: I am ok with submitting this temp solution if the real solution is coming soon. Also please add some comments to explain this with a link to the bug in the buganizer so that we don't forget about it. :)

include/ppx/grfx/grfx_swapchain.h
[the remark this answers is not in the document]
wangra-google: I think the `postprocess` in computer graphics is commonly referring to any full screen effects that are usually done to enhance/modify the visual of the rendered image. both UI and copying intermediate buffer do not belong to this category, so using `mPostProcessSemaphores` here might cause some misunderstanding.

src/ppx/grfx/grfx_swapchain.cpp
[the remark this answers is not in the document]
wangra-google: can you create a bug and put some comments with the link to the bug? So if we change the knob to adjust the resolution that are not matching the FB resolution, visually it would look wrong? if the srcImg has a higher res than the dstImg, would the copy cause any mem corruption?

src/ppx/grfx/grfx_swapchain.cpp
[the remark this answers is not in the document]
wangra-google: I see. can you create a bug for this? (RTVClearCount needs to be set to 1 even for LOAD pass) thanks :)

chaoticbob:
This is not the right direction to evolve the swapchain design. We need to move towards a path that make sit easier to evolve XR and non-XR independently. There is also a bunch of changes to idioms that are common to swapchains in general that make it confusing. It seems like the majority of these concepts belong in an app layer that sits over BigWheels.

The swapchain implementations need to keep simple without being convoluted by mixing app level idioms.

hysw:
> This is not the right direction to evolve the swapchain design. We need to move towards a path that make sit easier to evolve XR and non-XR independently. There is also a bunch of changes to idioms that are common to swapchains in general that make it confusing. It seems like the majority of these concepts belong in an app layer that sits over BigWheels.
> 
> The swapchain implementations need to keep simple without being convoluted by mixing app level idioms.

So making app not directly using swapchain? (e.g. https://github.com/hysw/bigwheels/commit/3931f6c8dbd7a3a002b3a062c9663127f50293dd ) Otherwise I don't see a way to insert command after present call.

chaoticbob:
> > This is not the right direction to evolve the swapchain design. We need to move towards a path that make sit easier to evolve XR and non-XR independently. There is also a bunch of changes to idioms that are common to swapchains in general that make it confusing. It seems like the majority of these concepts belong in an app layer that sits over BigWheels.
> > The swapchain implementations need to keep simple without being convoluted by mixing app level idioms.
> 
> So making app not directly using swapchain? (e.g. [hysw@3931f6c](https://github.com/hysw/bigwheels/commit/3931f6c8dbd7a3a002b3a062c9663127f50293dd) ) Otherwise I don't see a way to insert command after present call.

You can keep all of that at an app layer above BigWheels and customize the swapchain implementation by derivation. I don't want the swapchain interface in BW to deviate away from the APIs.

hysw:
Rebased and only kept whatever is still relevent, so marked comments as resolved as they no longer applies.

src/ppx/grfx/grfx_swapchain.cpp
```diff
@@ -50,26 +50,9 @@ Result Swapchain::Create(const grfx::SwapchainCreateInfo* pCreateInfo)
...
-
-            grfx::ImagePtr renderTarget;
-            ppxres = GetDevice()->CreateImage(&rtCreateInfo, &renderTarget);
-            if (Failed(ppxres)) {
-                return ppxres;
-            }
-
-            mColorImages.push_back(renderTarget);
-        }
+    ppxres = CreateColorImages();
```
apazylbe, line 53: nit: Maybe we should add a check to do this only if the swapchain is headless? Otherwise it has to happen after the API specific create, it probably always will, but maybe better to make that explicit.
Also wouldn't matter when we split the swapchain classes.

apazylbe (approved):

src/ppx/grfx/grfx_swapchain.cpp
hysw: If API specific create failed, then it won't get to this part, otherwise this behave the same as `CreateDepthImages`, create image if it's not exist.

src/ppx/grfx/grfx_swapchain.cpp
```diff
@@ -143,6 +129,67 @@ void Swapchain::Destroy()
...
+{
+    if (!mColorImages.empty()) {
+        return ppx::SUCCESS;
+    }
+    for (uint32_t i = 0; i < mCreateInfo.imageCount; ++i) {
+        grfx::ImageCreateInfo rtCreateInfo = ImageCreateInfo::RenderTarget2D(mCreateInfo.width, mCreateInfo.height, mCreateInfo.colorFormat);
+        rtCreateInfo.ownership             = grfx::OWNERSHIP_RESTRICTED;
+        rtCreateInfo.RTVClearValue         = {0.0f, 0.0f, 0.0f, 0.0f};
+        rtCreateInfo.initialState          = grfx::RESOURCE_STATE_PRESENT;
+        rtCreateInfo.usageFlags =
```
wangra-google, line 142: do we need IMAGE_USAGE_STORAGE to be consistent with CreateColorImagesFromApiObjects?

hysw: We can probably remove storage image usage flag.
`CreateColorImages` and `CreateColorImagesFromApiObjects` could share more of their parameters, however I'm not going to try to rewrite anything here (this change is mostly moving things around).

src/ppx/grfx/grfx_swapchain.cpp
```diff
@@ -155,20 +202,46 @@ void Swapchain::DestroyColorImages()
...
+    if (!mDepthImages.empty()) {
+        return ppx::SUCCESS;
+    }
+    if (mCreateInfo.depthFormat == grfx::FORMAT_UNDEFINED) {
+        return ppx::SUCCESS;
+    }
+    for (uint32_t i = 0; i < mCreateInfo.imageCount; ++i) {
+        grfx::ImageCreateInfo dpCreateInfo = ImageCreateInfo::DepthStencilTarget(mCreateInfo.width, mCreateInfo.height, mCreateInfo.depthFormat);
+        dpCreateInfo.ownership             = grfx::OWNERSHIP_RESTRICTED;
+        dpCreateInfo.DSVClearValue         = {1.0f, 0xFF};
```
Keenuts, line 214: This is already the default value. Can be removed.

src/ppx/grfx/grfx_swapchain.cpp
```diff
@@ -155,20 +202,46 @@ void Swapchain::DestroyColorImages()
...
+}
 
-            mDepthImages.push_back(depthStencilTarget);
+Result Swapchain::CreateDepthImagesFromApiObjects(const std::vector<void*>& apiObjects)
+{
+    PPX_ASSERT_MSG(mDepthImages.empty(), "Swapchain depth images already initialized.");
+    PPX_ASSERT_MSG(mCreateInfo.depthFormat != grfx::FORMAT_UNDEFINED, "Swapchain depth format not specified.");
+
+    for (size_t i = 0; i < apiObjects.size(); ++i) {
+        grfx::ImageCreateInfo imageCreateInfo = grfx::ImageCreateInfo::DepthStencilTarget(mCreateInfo.width, mCreateInfo.height, mCreateInfo.depthFormat, grfx::SAMPLE_COUNT_1);
```
Keenuts, line 234: SAMPLE_COUNT_1 is already the default value, can be removed.

src/ppx/grfx/grfx_swapchain.cpp
```diff
@@ -155,20 +202,46 @@ void Swapchain::DestroyColorImages()
...
+        mDepthImages.push_back(depthStencilTarget);
+    }
+
+    return ppx::SUCCESS;
+}
 
-            mDepthImages.push_back(depthStencilTarget);
+Result Swapchain::CreateDepthImagesFromApiObjects(const std::vector<void*>& apiObjects)
+{
+    PPX_ASSERT_MSG(mDepthImages.empty(), "Swapchain depth images already initialized.");
```
Keenuts, line 230: Feels slightly weird to error for that here, while above we return success with a no-op.

src/ppx/grfx/dx12/dx12_swapchain.cpp
```diff
@@ -246,27 +243,23 @@ Result Swapchain::CreateApiObjects(const grfx::SwapchainCreateInfo* pCreateInfo)
...
-                return ppxres;
-            }
-
-            mDepthImages.push_back(image);
+    if (!depthImages.empty()) {
+        std::vector<void*> apiObjects;
+        for (const auto& depthImage : depthImages) {
+            apiObjects.push_back(depthImage);
         }
+        CreateDepthImagesFromApiObjects(apiObjects);
```
Keenuts, line 262: nit: at least building for linux, seems like all those patterns can be shortened to either:
```cpp
std::vector<void*> apiObjects(depthImages.begin(), depthImages.end());
CreateDepthImagesFromApiObjects(apiObjects);
```
or
```cpp
CreateDepthImagesFromApiObjects({ depthImages.begin(), depthImages.end() });
```

src/ppx/grfx/grfx_swapchain.cpp
```diff
@@ -143,6 +129,67 @@ void Swapchain::Destroy()
...
+    if (!mColorImages.empty()) {
+        return ppx::SUCCESS;
+    }
+    for (uint32_t i = 0; i < mCreateInfo.imageCount; ++i) {
+        grfx::ImageCreateInfo rtCreateInfo = ImageCreateInfo::RenderTarget2D(mCreateInfo.width, mCreateInfo.height, mCreateInfo.colorFormat);
+        rtCreateInfo.ownership             = grfx::OWNERSHIP_RESTRICTED;
+        rtCreateInfo.RTVClearValue         = {0.0f, 0.0f, 0.0f, 0.0f};
+        rtCreateInfo.initialState          = grfx::RESOURCE_STATE_PRESENT;
+        rtCreateInfo.usageFlags =
+            grfx::IMAGE_USAGE_COLOR_ATTACHMENT |
```
Keenuts, line 143: nit: maybe this bit-list could use the same format as the other? (`imageCreateInfo.usageFlags.bits.transferSrc = true`, or the other way around?

src/ppx/grfx/grfx_swapchain.cpp
wangra-google: slightly different opinion here, only for discussion. :) personally, i feel it is better to assign the value here even it is using the default value so that it is clear what value is assigned to the member, otherwise, the user need to go to the definition of the struct and look for the default values.

src/ppx/grfx/grfx_swapchain.cpp
wangra-google: yeah, i found it is a bit weird as well. Also would it be better to have the default values defined in 1 place so that in the future if someone need to make a change, there is no need to modify in 2 functions?

PR #258, opened 2023-07-31
 3D grid for sphere instances generated randomly

elviscapiaq:
- It was added a knob to setup the number of spheres rendered in the scene. One sphere uses one `drawcall`.
- It renders up to 3000 spheres.


elviscapiaq:
This PR added a 3D grid of sphere instances generated randomly.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -455,8 +502,24 @@ void ProjApp::ProcessInput()
...
+    std::iota(indices.begin(), indices.end(), 0);
+    std::shuffle(indices.begin(), indices.end(), std::random_device{});
+    return {indices.begin(), indices.begin() + m};
+}
+
 void ProjApp::Render()
 {
+    mCurrentSphereCount = pSphereInstanceCount->GetValue();
+    if (mCurrentSphereCount != mPreviousSphereCount) {
+        mSphereIndices       = GenerateRandomIndices(mCurrentSphereCount, kMaxSphereInstanceCount);
```
wangra-google, line 519: i think we would need the spheres to be deterministic so that when we can compare the profiling result across different runs.

wangra-google (changes requested):

benchmarks/graphics_pipeline/main.cpp
elviscapiaq: Done.

elviscapiaq:
Ready for review.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -189,6 +295,195 @@ void ProjApp::Setup()
...
+    // Read/Generate sphere indices
+    {
+        const std::filesystem::path path = kSphereIndicesFilePath;
+        if (std::filesystem::exists(path)) {
+            mSphereIndices = ReadSphereIndicesFromJson(kSphereIndicesFilePath);
+            PPX_ASSERT_MSG(mSphereIndices.size() == kMaxSphereInstanceCount + 1, "Size of mSphereIndices must be " << kMaxSphereInstanceCount + 1);
+        }
+        else {
+            mSphereIndices.resize(kMaxSphereInstanceCount + 1);
+            for (uint32_t i = 0; i <= kMaxSphereInstanceCount; i++) {
```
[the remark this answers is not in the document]
elviscapiaq: You are right, I think I overkilled this part. Updated!

elviscapiaq:
Ready for review!

wangra-google (approved):

elviscapiaq:
I rebased this PR with the origin/main branch. I think it is ready to be merged.

assets/benchmarks/sphere_indices.json
hysw, line 1: Is there any reason to have a pre-generated list of indices? A pre-generated file will be invalid any time `kMaxSphereInstanceCount` is updated.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -285,6 +335,41 @@ void ProjApp::Setup()
...
+    {
+        const std::filesystem::path path = kSphereIndicesFilePath;
+        if (std::filesystem::exists(path)) {
+            mSphereIndices = ReadSphereIndicesFromJson(kSphereIndicesFilePath);
+            PPX_ASSERT_MSG(mSphereIndices.size() == kMaxSphereInstanceCount, "Size of mSphereIndices must be " << kMaxSphereInstanceCount);
+        }
+        else {
+            mSphereIndices.resize(kMaxSphereInstanceCount);
+            std::iota(mSphereIndices.begin(), mSphereIndices.end(), 0);
+            std::shuffle(mSphereIndices.begin(), mSphereIndices.end(), std::random_device{});
```
hysw, line 356: Have you consider using deterministic pseudo random number generator? You can reproduce a sequence with a seed instead of storing the entire sequence.

hysw (commented):
Just some personal opinion on the index sequence.

assets/benchmarks/sphere_indices.json
elviscapiaq: Yes. We want investigators play some runtime options like increase the "number of spheres" and when they detect a big drop in performance we want to replay this setup including the position of the spheres. So, this is the reason why we want to have a pre-defined list of positions for sphere instances. Here the indices are used to determine the sphere positions in the scene.

benchmarks/graphics_pipeline/main.cpp
elviscapiaq: Thanks. Done.

elviscapiaq:
Ready for review.

elviscapiaq:
I declined to include a deterministic random number generator. This is because the generated random numbers can vary based on how the shuffle method is implemented. To simplify things, I am reverting to the original approach of using JSON files to generate deterministic random numbers for the sphere indices.

Keenuts:
> I declined to include a deterministic random number generator. This is because the generated random numbers can vary based on how the shuffle method is implemented. To simplify things, I am reverting to the original approach of using JSON files to generate deterministic random numbers for the sphere indices.

Which PRNG implementation can vary?
Isn't https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine guaranteed to behave the same?

```cpp
std::mt19937 gen(my_seed);
std::shuffle(my_vector.begin(), me_vector.end(), gen);
```

Keenuts:
> It's `std::shuffle`, which can vary between different standard libraries.

😢.... thanks!

> Although a deterministic shuffle takes 3 lines to implement.

Yes, since this would be used to shuffle not only drawcalls, but also attributes etc, we might want to have a proper shuffle function than, and store the seed in the config (instead of storing the generated sequence in a json file)

hysw:
> > Although a deterministic shuffle takes 3 lines to implement.
> 
> Yes, since this would be used to shuffle not only drawcalls, but also attributes etc, we might want to have a proper shuffle function than, and store the seed in the config (instead of storing the generated sequence in a json file)

MS's implementation is basically 
```
template <typename Iter, typename F>
void vs_shuffle(Iter begin, Iter end, F&& f) {
  size_t count = end - begin;
  for (size_t i = 1; i < count ; i++) {
    using std::swap;
    swap(begin[i], begin[f() % (i + 1)]);
  }
}
```

glibc's shuffle have a specialization for `count < f() / count`

elviscapiaq:
Thanks for the feedback. I added an own implementation for the `Shuffle` function and I used a deterministic random number generator as suggested. Ready for review.

Keenuts (approved):

PR #260, opened 2023-08-01
Change StandardOptions into knobs

angela28chen:
* Changing `StandardOptions` struct to hold knobs
* Standard options flag usage message is now generated
* Reducing some Application variables that are the same as `StandardOptions` (`mMaxFrames`, all metrics-related ones inside `ApplicationSettings`)
* Simplifying some of the `Application::Run` logic since knob ranges and validators will cover those checks
* Adding some test cases for complex knobs
* Added `ToString()` util for drawing values of complex knobs
* Fixing broken `--assets-path` flag which would not register when used multiple times
* Fixing bug with `WrapText()` which would break the line too early

### DIFF
--- a/src/ppx/application.cpp
+++ b/src/ppx/application.cpp
@@ -326,7 +326,7 @@ Result Application::InitializeGrfxDevice()
         ci.enableSwapchain          = true;
         ci.applicationName          = mSettings.appName;
         ci.engineName               = mSettings.appName;
-        ci.useSoftwareRenderer      = mStandardOptions.use_software_renderer;
+        ci.useSoftwareRenderer      = mStandardOpts.pUseSoftwareRenderer->GetValue();
 #if defined(PPX_BUILD_XR)
         ci.pXrComponent = mSettings.xr.enable ? &mXrComponent : nullptr;
         // Disable original swapchain when XR is enabled as the XR swapchain will be coming from OpenXR.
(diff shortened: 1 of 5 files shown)


src/ppx/application.cpp
```diff
@@ -674,9 +670,10 @@ void Application::DispatchConfig()
         AddAssetDir(path / "third_party/assets");
     }
 
-    if (!mStandardOptions.assets_paths.empty()) {
+    auto assetPaths = mStandardOpts.pAssetsPaths->GetValue();
+    if (!assetPaths.empty()) {
         // Insert at front, in reverse order, so we respect the command line ordering for priority.
-        for (auto it = mStandardOptions.assets_paths.rbegin(); it < mStandardOptions.assets_paths.rend(); it++) {
+        for (auto it = assetPaths.rbegin(); it < assetPaths.rend(); it++) {
```
Keenuts, line 676: options are parsed in the correct order no?
So `rbegin` should be `begin`.
(was probably wrong before..)

Keenuts (approved):

Keenuts:
> * I believe Aliya suggested previously that it's better to have the knobs defined and initialized inside `command_line_parser` files to avoid bloating `application`, but I couldn't figure out how to do this without running into circular dependency issues since `knob.h` includes the parser header.

What is the issue?


> * Currently the knobs are ordered in the usage message in the same order as they are defined (standard knobs, and then application-specific knobs). I think this ordering makes sense but it would be nice to have a delineating line between standard knobs and application-specific ones, and maybe automatic alphabetization rather than relying on manual? I'm not sure how to approach this though, maybe a follow-up CL adding a specific property to `Knob` to indicate if it is "standard" or not?

my 2 cents: I did a command-line parsing utility for SPIRV-Tools not so long ago. Same thing, automatic parsing and definition.
I tried a *LOT* to have a flexible-enough help text, with separators, groups, having indentation match what we could want (things like notes going to a new line?).
Conclusion was: have the help message be a constant global I just print.
The code complexity was not worth it.

So maybe the simplest option is to have a global text message for common flags, and have each application provide a second bloc to append, with their own help.

PR #261, opened 2023-08-01
Bug fixes, helper functions and better logging for string_util and knobs

angela28chen:
Preparing knobs and utilities for implementing `StandardOptions` as knobs

To support vector knobs
* `GetExtraOptionValueOrDefault` in knob code replaced with `GetOptionValueOrDefault`
* Adding some unit test cases for complex knobs

Logging fixes/changes
* Changing `Application-Specific Flags` title into `Flags`
* Fix bug for `string_util::WrapText()` which would line break too early, added unit test case to catch this
* Added `Knob::ValueString()` and logging for initial knob values inside `UpdateFromFlags`, for easier debugging 
* Added `string_util::ToString()` util for drawing values of knobs

### DIFF
--- a/src/ppx/knob.cpp
+++ b/src/ppx/knob.cpp
@@ -60,9 +60,14 @@ void KnobCheckbox::Draw()
     RaiseUpdatedFlag();
 }
 
+std::string KnobCheckbox::ValueString()
+{
+    return ppx::string_util::ToString(mValue);
+}
+
 void KnobCheckbox::UpdateFromFlags(const CliOptions& opts)
 {
-    SetDefaultAndValue(opts.GetExtraOptionValueOrDefault(mFlagName, mValue));
+    SetDefaultAndValue(opts.GetOptionValueOrDefault(mFlagName, mValue));
 }
 
 void KnobCheckbox::SetValue(bool newValue)
@@ -121,7 +126,7 @@ void KnobManager::DrawAllKnobs(bool inExistingWindow)
 
 std::string KnobManager::GetUsageMsg()
 {
-    std::string usageMsg = "\nApplication-Specific Flags:\n";
+    std::string usageMsg = "\nFlags:\n";
     for (const auto& knobPtr : mKnobs) {
         std::string knobMsg = "--" + knobPtr->mFlagName;
         if (knobPtr->mFlagParameters != "") {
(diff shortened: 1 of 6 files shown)


src/ppx/knob.cpp
```diff
@@ -140,6 +145,7 @@ void KnobManager::UpdateFromFlags(const CliOptions& opts)
 {
     for (auto& knobPtr : mKnobs) {
         knobPtr->UpdateFromFlags(opts);
+        PPX_LOG_INFO("KNOB: " << knobPtr->mFlagName << " : (" << knobPtr->ValueString() << ")");
```
apazylbe, line 148: Is this only used for debugging or is it useful in general?

src/ppx/string_util.cpp
```diff
@@ -80,9 +80,19 @@ std::string WrapText(const std::string& s, size_t width, size_t indent)
     std::string wrappedText = "";
     while (remainingString != "") {
         // Section off the next line from the remaining string, format it, and append it to wrappedText
-        size_t lineLength = remainingString.find_last_of(" \t", textWidth);
-        if (lineLength == std::string::npos) {
-            lineLength = std::min(textWidth, remainingString.length());
+        size_t lineLength = remainingString.length();
+        if (lineLength > textWidth) {
+            // Break the line at textWidth
+            lineLength = textWidth;
```
apazylbe, line 86: Ah, I think the case not covered before was if it was the last line, which your if above is taking care of. You could still place the code from 83-85 inside the if statement, and I think wouldn't need the std::min. It might be simpler, up to you though.

src/test/string_util_test.cpp
```diff
@@ -179,4 +179,99 @@ dog.
...
+                    Pellentesque vel urna ac tellus euismod finibus quis ac
+                    magna. Cras sit amet sapien id neque lobortis aliquam.
+                    Vivamus porttitor neque eu eros mollis imperdiet. Vivamus
+                    blandit neque sed nisl pretium, quis volutpat dui pharetra.
+)";
+    std::string wrapped  = WrapText(toWrap, 80, 20);
+    EXPECT_EQ(wrapped, wantWrap);
+}
+
+TEST(StringUtilTest, ToString_Bool)
```
apazylbe, line 210: I think these (bool, int and float) should be split into separate test cases to "test one thing"

Keenuts (approved):

PR #262, opened 2023-08-03
Move C++20 feature out of the header file.

at-ninja:
[`std::unordered_map::contains`](https://en.cppreference.com/w/cpp/container/unordered_map/contains) was added in C++20.

This keeps the header file C++17 compliant.


Keenuts:
The project builds with C++20 by default (https://github.com/google/bigwheels/blob/main/CMakeLists.txt#L116).
This shouldn't be something we handle here.

PR #265, opened 2023-08-10
Add --config-json-path to accept flags specified in a JSON file.

angela28chen:
Add flag --config-json-path to accept commandline flags that are specified in a JSON file. This flag can be used multiple times. The priority of any conflicting flags is as follows:

later flag on commandline > earlier flag on commandline > flag in later JSON file > flag in earlier JSON file

For flags that accept multiple values, the same priority is used and only flags specified in the same JSON file or flags all on the commandline will be combined.

Detailed changes:
* Changed map `mAllOptions` to hold `strings` instead of `string_views` since there are now temporary strings originating from the JSON structure, changed some input parameters for `CliOptions` member functions to match.
* Created some `CommandLineParser` methods to aid in parsing options (these are public for unit testing purposes)
  * `ParseJson`
  * `ParseOption`
* Created more methods for `CliOptions`
  * `AddOption` overload that accepts one key and a vector of values to reduce the time needed to insert arrays. This function is only used by the JSON parsing flow. On the command line the flag must still be specified multiple times (`--flag-name 1,2,3` is still invalid)
  * `OverwriteOptions`
* Updated `CommandLineParser::Parse`
  * Added a second initial pass of argv during commandline parsing to identify any provided JSON config files
  * Implemented overwriting
* Added `mJsonConfigFlagName` to `CommandLineParser` so that the string `config-json-path` isn't specified in multiple places

### DIFF
--- a/src/ppx/application.cpp
+++ b/src/ppx/application.cpp
@@ -806,11 +806,16 @@ void Application::SaveMetricsReportToDisk()
 void Application::InitStandardKnobs()
 {
     // Flag names in alphabetical order
-    std::vector<std::string> defaultAssetsPaths = {};
+    std::vector<std::string> defaultEmptyList = {};
     mStandardOpts.pAssetsPaths =
-        mKnobManager.CreateKnob<KnobFlag<std::vector<std::string>>>("assets-path", defaultAssetsPaths);
+        mKnobManager.CreateKnob<KnobFlag<std::vector<std::string>>>("assets-path", defaultEmptyList);
     mStandardOpts.pAssetsPaths->SetFlagDescription(
         "Add a path in front of the assets search path list.");
+    mStandardOpts.pAssetsPaths->SetFlagParameters("<path>");
+
+    mStandardOpts.pConfigJsonPath = mKnobManager.CreateKnob<KnobFlag<std::vector<std::string>>>("config-json-path", defaultEmptyList);
+    mStandardOpts.pAssetsPaths->SetFlagDescription(
+        "Additional commandline flags specified in a JSON file.");
     mStandardOpts.pAssetsPaths->SetFlagParameters("<path>");
 
     mStandardOpts.pDeterministic =
--- a/src/ppx/application.cpp
+++ b/src/ppx/application.cpp
@@ -813,13 +813,13 @@ void Application::InitStandardKnobs()
         "Add a path in front of the assets search path list.");
     mStandardOpts.pAssetsPaths->SetFlagParameters("<path>");
 
-    mStandardOpts.pConfigJsonPath = mKnobManager.CreateKnob<KnobFlag<std::vector<std::string>>>(mCommandLineParser.GetJsonConfigFlagName(), defaultEmptyList);
-    mStandardOpts.pConfigJsonPath->SetFlagDescription(
+    mStandardOpts.pConfigJsonPaths = mKnobManager.CreateKnob<KnobFlag<std::vector<std::string>>>(mCommandLineParser.GetJsonConfigFlagName(), defaultEmptyList);
+    mStandardOpts.pConfigJsonPaths->SetFlagDescription(
         "Additional commandline flags specified in a JSON file. Values specified in JSON files are "
         "always lower priority than those specified on the command line. Between different files, the "
         "later ones take priority. For lists, the JSON values will come earlier in the array than the "
         "command-line values");
-    mStandardOpts.pConfigJsonPath->SetFlagParameters("<path>");
+    mStandardOpts.pConfigJsonPaths->SetFlagParameters("<path>");
 
     mStandardOpts.pDeterministic =
         mKnobManager.CreateKnob<KnobFlag<bool>>("deterministic", false);
--- a/src/ppx/application.cpp
+++ b/src/ppx/application.cpp
@@ -672,8 +672,8 @@ void Application::DispatchConfig()
 
     auto assetPaths = mStandardOpts.pAssetsPaths->GetValue();
     if (!assetPaths.empty()) {
-        // Insert at front, in reverse order, so we respect the command line ordering for priority.
-        for (auto it = assetPaths.rbegin(); it < assetPaths.rend(); it++) {
+        // Asset directories specified later have higher priority
+        for (auto it = assetPaths.begin(); it < assetPaths.end(); it++) {
             AddAssetDir(*it, /* insert_at_front= */ true);
         }
     }
--- a/src/ppx/application.cpp
+++ b/src/ppx/application.cpp
@@ -813,10 +813,13 @@ void Application::InitStandardKnobs()
         "Add a path in front of the assets search path list.");
     mStandardOpts.pAssetsPaths->SetFlagParameters("<path>");
 
-    mStandardOpts.pConfigJsonPath = mKnobManager.CreateKnob<KnobFlag<std::vector<std::string>>>("config-json-path", defaultEmptyList);
-    mStandardOpts.pAssetsPaths->SetFlagDescription(
-        "Additional commandline flags specified in a JSON file.");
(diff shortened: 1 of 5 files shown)


Keenuts:
> Note: I tried to keep the input parameters for `CliOptions` as `std::string_view` in some cases, but I was running into issues with heterogeneous lookup for unordered containers (could get it to work on Windows but not on the current versions of the Android and Linux CI builds)

Arf, that's sad...
Seeing this https://groups.google.com/a/isocpp.org/g/std-proposals/c/oXBQ6WVXLCU
meaning map does supports it since 14(https://en.cppreference.com/w/cpp/container/map/find)

What's sadder is it wasn't done to the unordered_map until c++20 😥
(https://en.cppreference.com/w/cpp/container/unordered_map/find)

So this explains the lack of support, at least from the Android NDK.
Surprising to see clang also refuses it...
Thanks for looking into that!

src/ppx/command_line_parser.cpp
```diff
@@ -118,25 +133,73 @@ std::optional<CommandLineParser::ParsingError> CommandLineParser::Parse(int argc
...
+
+std::optional<CommandLineParser::ParsingError> CommandLineParser::AddJsonOptions(const nlohmann::json& jsonConfig)
+{
+    std::stringstream ss;
+    for (auto it = jsonConfig.cbegin(); it != jsonConfig.cend(); ++it) {
+        ss << it.value();
+        std::string value = ss.str();
+        ss.str("");
+
+        if (value.length() > 0 && value.substr(0, 1) == "[") {
```
Keenuts, line 175: you probably shouldn't parse the data again. What about using functions like `it->is_array()`.

src/ppx/command_line_parser.cpp
```diff
@@ -118,25 +133,73 @@ std::optional<CommandLineParser::ParsingError> CommandLineParser::Parse(int argc
...
+            // Special case, arrays specified in JSON are added directly to mOpts to avoid inserting element by element
+            std::vector<std::string> jsonStringArray;
+            for (const auto& elem : it.value()) {
+                ss << elem;
+                jsonStringArray.push_back(ss.str());
+                ss.str("");
+            }
+            mOpts.AddOption(it.key(), jsonStringArray);
```
Keenuts, line 183: You assumes the passed config is an object. You should validate that.

src/ppx/command_line_parser.cpp
```diff
@@ -118,25 +133,73 @@ std::optional<CommandLineParser::ParsingError> CommandLineParser::Parse(int argc
...
+        nlohmann::json data = nlohmann::json::parse(f);
+        if (auto error = AddJsonOptions(data)) {
+            return error;
+        }
+    }
+
+    return std::nullopt;
+}
+
+std::optional<CommandLineParser::ParsingError> CommandLineParser::AddJsonOptions(const nlohmann::json& jsonConfig)
```
Keenuts, line 167: Overall: don't assume anything regarding the json config layout.
The top-level object could be an array, flag values could be arrays when not expected, or even objects, etc.
Please use the `is_array()`, `is_object()`, etc to validate the structure when loading it.
Reading `it->value()` is fine as we delegate the parsing to the command_line utility, but the rest should be validated.

src/ppx/application.cpp
```diff
@@ -800,13 +800,18 @@ void Application::SaveMetricsReportToDisk()
...
     mStandardOpts.pAssetsPaths =
-        mKnobManager.CreateKnob<KnobFlag<std::vector<std::string>>>("assets-path", defaultAssetsPaths);
+        mKnobManager.CreateKnob<KnobFlag<std::vector<std::string>>>("assets-path", defaultEmptyList);
     mStandardOpts.pAssetsPaths->SetFlagDescription(
         "Add a path in front of the assets search path list.");
     mStandardOpts.pAssetsPaths->SetFlagParameters("<path>");
 
+    mStandardOpts.pConfigJsonPath = mKnobManager.CreateKnob<KnobFlag<std::vector<std::string>>>("config-json-path", defaultEmptyList);
+    mStandardOpts.pAssetsPaths->SetFlagDescription(
+        "Additional commandline flags specified in a JSON file.");
```
Keenuts, line 812: You should describe the complete behavior:
`-flag=1 -json=toto.json -json=tata.json`
-> tata.json values override values from toto.json

`-flag=1 -json=toto.json -flag=2`
-> toto.json overrides any values, even if `flag` is specified after.

Keenuts (commented):
Thanks for this!
A few things to fix, but overall OK.

The the JSON overriding flags, even when specified after, this is fine IMO. Complexity to manage the ordering is not yet justified. But it should be documented.

apazylbe (commented):
some optional suggestions, otherwise lgtm

apazylbe (approved):

src/ppx/command_line_parser.cpp
```diff
@@ -131,14 +121,24 @@ std::optional<CommandLineParser::ParsingError> CommandLineParser::Parse(int argc
...
-            continue;
-        }
         args.emplace_back(res->first);
         args.emplace_back(res->second);
     }
 
+    // Another pass to identify JSON config files, add that option, and remove it from the argument list
+    std::vector<std::string_view> argsFiltered;
+    for (size_t i = 0; i < args.size(); ++i) {
+        bool nextArgumentIsParameter = (i + 1 < args.size()) && !StartsWithDoubleDash(args[i + 1]);
```
Keenuts, line 131: One thing that bugs me: it seems this is the only flag which allows `--flag value`.
Seems like all the other flags have the format `--flag=value`.
Seems a bit inconsistent.

I'm fine sending this for now, but once you work on the error handling in flags, I believe this should be sanitized.
There is one ambiguous case if you allow `--flag value`: the booleans.
For SPIRV-Tools, `--bool-flag` cannot have a second value.
so it's either `--bool-flag`, `--no-bool-flag` or `--bool-flag=false`.
But for all other flags, it can be `--flag=value` and `--flag value`

Keenuts (commented):
Forgot to send the review, sorry!

src/ppx/application.cpp
```diff
@@ -800,13 +800,21 @@ void Application::SaveMetricsReportToDisk()
...
     mStandardOpts.pAssetsPaths->SetFlagDescription(
         "Add a path in front of the assets search path list.");
     mStandardOpts.pAssetsPaths->SetFlagParameters("<path>");
 
+    mStandardOpts.pConfigJsonPath = mKnobManager.CreateKnob<KnobFlag<std::vector<std::string>>>(mCommandLineParser.GetJsonConfigFlagName(), defaultEmptyList);
+    mStandardOpts.pConfigJsonPath->SetFlagDescription(
+        "Additional commandline flags specified in a JSON file. Values specified in JSON files are "
+        "always lower priority than those specified on the command line. Between different files, the "
+        "later ones take priority. For lists, the JSON values will come earlier in the array than the "
+        "command-line values");
```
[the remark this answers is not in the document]
Keenuts: Just cloned this PR and tried:

`file.json`
```json
{
  "assets-path": [ "/tmp/c", "/tmp/d" ]
}
```

`./build/bin/vk_01_triangle --assets-path /tmp/a --assets-path /tmp/b --config-json-path file.json`

The lookup order is:
```
"/tmp/b"
"/tmp/a"
"/tmp/d"
"/tmp/c"
```

So the JSON paths are indeed in the back, the the insertion order is flipped.

I believe the logical order should be `A, B, C D` (flag order, and then, json order)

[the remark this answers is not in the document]
Keenuts: Oh you pushed 8mn ago, let me clone again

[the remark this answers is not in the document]
Keenuts: Yep, I observe `B, A, D, C`, and not `A, B, C, D`

apazylbe (approved):

PR #266, opened 2023-08-10
merging patch from internal upstream.

phoenixxxx:


s-perron:
This change moves C++20 specific code to the cpp file. This way the header file can be included in a C++17 project.

Keenuts:
This was addressed a couple days ago internally, so we shouldn't need that.
(Feel free to re-open if there is a different need I miss)

PR #269, opened 2023-08-15
Add back the `--metrics-filename` knob.

at-ninja:
This knob happened to be removed as part of https://github.com/google/bigwheels/pull/260. Adding it back, so that we can fully use the metrics feature again.

### DIFF
--- a/src/ppx/application.cpp
+++ b/src/ppx/application.cpp
@@ -842,6 +842,16 @@ void Application::InitStandardKnobs()
         "Prints a list of the available GPUs on the current system with their "
         "index and exits (see --gpu).");
 
+    mStandardOpts.pMetricsFilename =
+        mKnobManager.CreateKnob<KnobFlag<std::string>>("metrics-filename", "report_@");
+    mStandardOpts.pMetricsFilename->SetFlagDescription(
+        "If metrics are enabled, save the metrics report to the "
+        "provided filename (including path). If used, any `@` "
+        "symbols in the filename (not the path) will be replaced "
+        "with the current timestamp. If the filename does not end in "
+        "`.json`, it will be appended. Default: `report_@`. See also: "
+        "`--enable-metrics` and `--overwrite-metrics-file`.");
+
     mStandardOpts.pOverwriteMetricsFile =
         mKnobManager.CreateKnob<KnobFlag<bool>>("overwrite-metrics-file", false);
     mStandardOpts.pOverwriteMetricsFile->SetFlagDescription(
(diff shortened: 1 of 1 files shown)


src/ppx/application.cpp
```diff
@@ -842,6 +842,16 @@ void Application::InitStandardKnobs()
         "Prints a list of the available GPUs on the current system with their "
         "index and exits (see --gpu).");
 
+    mStandardOpts.pMetricsFilename =
```
Slumpwuffle, line 845: Can you please also add proper handling if this value is never set, as discussed?

at-ninja: Done, I've added checks to `Application::SaveMetricsReportToDisk` to prevent `nullptr`s from crashing the application.

src/ppx/application.cpp
```diff
@@ -788,13 +788,18 @@ void Application::ShutdownMetrics()
 
 void Application::SaveMetricsReportToDisk()
 {
-    if (!mStandardOpts.pEnableMetrics->GetValue()) {
+    if (mStandardOpts.pEnableMetrics == nullptr || !mStandardOpts.pEnableMetrics->GetValue()) {
```
apazylbe, line 791: Did you actually run into the case of this being nullptr? 
I don't see how that can happen, but I might be missing something. I think it should be asserted that it isn't nullptr instead.

src/ppx/application.cpp
```diff
@@ -788,13 +788,18 @@ void Application::ShutdownMetrics()
...
         return;
     }
 
+    std::string metricsFilename;
+    if (mStandardOpts.pMetricsFilename != nullptr) {
```
apazylbe, line 796: Same question of whether this can be nullptr. Also this has a default value of report_@, the one error I ran into is in Report::WriteToDisk when there parent_path is empty with the default value, and it throws an error. I think that's something we need to fix as well.

src/ppx/application.cpp
at-ninja: I've changed these to assertions. There's currently a crash at `HEAD` from a null pointer dereference in this function, and the assertions should improve the error messages if it happens again in the future.

src/ppx/application.cpp
at-ninja: I'd like to restore the functionality that was removed during the Knob transition. Can the `WriteToDisk` issue be addressed in a follow-up PR?

The assertions should catch if `nullptr` finds its way back into this function.

apazylbe (approved):

src/ppx/application.cpp
apazylbe: Sure!

Keenuts (approved):

PR #270, opened 2023-08-15
 Adjust the vertex shader input signature to match the vertex binding

elviscapiaq:
- The type of POSITION (float4) in the vertex shader input signature has
  been modified, as it didn't correspond with the type of POSITION
  (float3) in the vertex binding.

### DIFF
--- a/assets/benchmarks/shaders/Benchmark_VsAluBound.hlsl
+++ b/assets/benchmarks/shaders/Benchmark_VsAluBound.hlsl
@@ -15,7 +15,7 @@
 #include "Benchmark.hlsli"
 
 VSOutput vsmain(
-    float4 position : POSITION,
+    float3 position : POSITION,
     float2 uv       : TEXCOORD,
     float3 normal   : NORMAL,
     float3 tangent   : TANGENT)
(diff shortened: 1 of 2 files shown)


elviscapiaq:
I have confirmed that both the "VS simple" and "VS ALU bound" are operating as intended. Currently, for skinning we are contemplating the use of fake skinning matrices for both the joint matrix and weight matrix. Ready for your.

wangra-google (approved):

Keenuts (commented):
LGTM. Waiting for you to either rebase on main, or to have parent commits submitted.

Keenuts:
One question: How did you find this out?
Did a layer complained? Or something?
Renderdoc does show the float3 for the vertex input, and vulkan layers don't seem to complain, so curious if this was just by reading the code, or something else.

elviscapiaq:
> One question: How did you find this out?
> Did a layer complained? Or something?
> Renderdoc does show the float3 for the vertex input, and vulkan layers don't seem to complain, so curious if this was just by reading the code, or something else.
Mainly, it was through reading the code, but I was also trying to figure out why no layer was complaining about it. In the end, I realized that for most GPUs, when the types between vertex input and vertex shader input signature don't match, we have two scenarios: 
i) If we expect to read more data in the vertex shader from the vertex input, then the missing data is usually filled with `ones`. This was the reason why it had been working fine so far. (This is not safe and the main reason for this PR)
ii) When we expect to read less data in the vertex shader from the vertex input, the extra data to be read from the vertex input is ignored. (This is safe)

apazylbe:
The parent changes seem to be submitted, I will wait for you to rebase before reviewing and merging

elviscapiaq:
I rebased the PR. Ready for review.

elviscapiaq:
I rebased this PR. Ready for review.

apazylbe (approved):

PR #273, opened 2023-08-17
Fix gradle isssue with packaging generated shaders

apazylbe:
Previously copyTask would run before shaders are actually generated, so they wouldn't be included correctly on the first try after a clean build.

This changes a few things:
- copyTask must now run after the native build that generates shaders, and before generate assets task
- subprojects specify a list of assets, and a task defined at the root build.gradle copies them
- the shaders are copied into a build folder, instead of a new assets folder in a subproject

### DIFF
--- a/projects/15_basic_material/build.gradle
+++ b/projects/15_basic_material/build.gradle
@@ -1,23 +1,13 @@
-task copyTask {
-    copy {
-        from rootProject.ext.assetsPath + '/basic/models'
-        into 'assets/basic/models'
-    }
-    copy {
-        from rootProject.ext.assetsPath + '/basic/shaders/spv'
-        into 'assets/basic/shaders/spv'
-        include '*.spv'
-    }
-    copy {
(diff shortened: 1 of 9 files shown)


projects/15_basic_material/build.gradle
```diff
@@ -1,23 +1,13 @@
...
-    copy {
-        from rootProject.ext.thirdPartyAssetsPath + '/poly_haven'
-        into 'assets/poly_haven'
-    }
-}
+ArrayList projectAssets = [
+    "basic/models",
+    "basic/shaders/spv",
+    "common/textures",
+    "materials/shaders",
```
Keenuts, line 5: `/spv` at the end? (seems like we just copy additional HLSL files)

Keenuts (approved):

projects/15_basic_material/build.gradle
apazylbe: thanks! fixed

PR #275, opened 2023-08-21
Moving string parsing logic from `CliOptions` into `string_util`

angela28chen:
For a more logical structure, and also because this can be re-used for setting knob values with textbox inputs in the UI.


apazylbe:
Thanks for the changes! Just wondering, was the change to include KnobDemo intentional? Since it wasn't there at first.

include/ppx/command_line_parser.h
```diff
@@ -64,11 +66,17 @@ class CliOptions
...
+        if (error != std::nullopt) {
+            PPX_LOG_ERROR(error->errorMsg);
+            return defaultValue;
+        }
+        return parsedValue;
     }
 
     // Same as above, but intended for list flags that are specified on the command line
-    // with multiple instances of the same flag
+    // with multiple instances of the same flag, or with comma-separated values
```
apazylbe, line 79: I vaguely remember @Keenuts suggesting we stick to one flag one value for simplicity, and since "," can be a valid name on Linux, and a similar change was removed in a previous PR. Is there any reason to add it now? 
Also based on brief testing: 
./program --assets-path=A,B --assets-path=C    => (A,B,C)
./program --assets-path=A,B --assets-path=C,D => (A,B)        and C,D are ignored
So only the first one can be comma separated?

include/ppx/string_util.h
```diff
@@ -31,10 +35,18 @@ std::string TrimCopy(const std::string& s);
 // Trims all characters specified in c from both the left and right sides of s
 std::string_view TrimBothEnds(std::string_view s, std::string_view c = " \t");
 
+// Splits s at every instance of delimeter and returns a vector of substrings
+// Returns std::nullopt if s contains: leading/trailing/consecutive delimiters
+std::optional<std::vector<std::string_view>> Split(std::string_view s, char delimiter);
```
apazylbe, line 40: Personally, I think it would be cleaner for these two to just return the value (std::vector or std::pair) without the optional, and for the caller to know and deal with the empty values, etc. But up to you!

include/ppx/command_line_parser.h
apazylbe: I see. I think it is fine to allow comma separated values, but I would at least want to consistent and predictable. Seems like right now if the first value is comma separated, then subsequent comma separated values are ignored? 
Also fwiw, I don't think it would be very meaningful for assets paths to be modified at runtime at the moment, since most are loaded at startup, but I guess there would be other uses of list of values.

include/ppx/string_util.h
Keenuts: yes. Also, this is a basic string util, so I'd expect it's behavior to be straightforward:
split at the separator, and return the values in between.
No delimiter -> return the whole string
Multiple delimiters -> returns en empty tokens between each delimiter.

include/ppx/string_util.h
```diff
@@ -31,9 +37,17 @@ std::string TrimCopy(const std::string& s);
...
 std::string_view TrimBothEnds(std::string_view s, std::string_view c = " \t");
 
+// Splits s at every instance of delimeter and returns a vector of substrings
+// Return value will contain empty substrings if s has leading/trailing/consecutive delimiters
+std::vector<std::string_view> Split(std::string_view s, char delimiter);
+
 // Splits s at the first instance of delimeter and returns two substrings
-// Returns std::nullopt if s does not contain the delimeter
-std::optional<std::pair<std::string_view, std::string_view>> SplitInTwo(std::string_view s, char delimiter);
+// Returns an empty pair if s does not have exactly one delimiter, or if either of the substrings are empty
```
apazylbe, line 45: Since this is a string utility that can be used for other things, I would actually go one step further, and split on the first instance of delimeter, and return everything pre and everything post without making any assumptions.

If there is no delimiter, the second part will be empty, if there is more than one delimeter the second value will have those delimeters, if the strings are empty, return empty strings, then allow the caller to decide what to do with those return values.
With the current implementation it is acceptable to pass:
`--list-gpus ""`
but the following would produce an error: 
`--list-gpus=""`
which sounds inconsistent

apazylbe (commented):
Just two comments, otherwise lgtm. Thanks Angela! 
I guess we don't need Split into the vector anymore since we can implement SplitInTwo without that, but if you think it is useful, sounds good to me.

include/ppx/string_util.h
```diff
@@ -74,6 +88,47 @@ std::string ToString(std::pair<T, T> values)
     return ss.str();
 }
 
+// -------------------------------------------------------------------------------------------------
+// Parsing Strings
+// -------------------------------------------------------------------------------------------------
+
+// Parse() attempts to parse valueStr into the specified type
+// If successful, overwrites parsedValue and returns std::nullopt
+// If unsucessful, does not overwrite parsedValuel, logs error and returns ERROR_FAILED instead
```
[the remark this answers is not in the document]
angela28chen: Thanks, fixed

include/ppx/string_util.h
angela28chen: I agree, made those changes.
I changed a bit of `Commandline::Parse` logic so that empty value is allowed and empty option name is caught by other checks, added appropriate unit tests.

angela28chen:
Thanks! I removed `Split()`, changed logic for `SplitInTwo()` and made the callers validate the result more, and updated the PR description. I think this is ready for merging

src/test/command_line_parser_test.cpp
```diff
@@ -169,45 +149,55 @@ TEST(CommandLineParserTest, Parse_EqualSignsMultipleFail)
...
+}
+
+TEST(CommandLineParserTest, Parse_EqualSignsEmptyValuePass)
 {
     CommandLineParser parser;
     const char*       args[] = {"/path/to/executable", "--a", "--b=", "--c", "--d", "11"};
-    auto              error  = parser.Parse(sizeof(args) / sizeof(args[0]), args);
-    EXPECT_TRUE(error);
-    EXPECT_THAT(error->errorMsg, HasSubstr("Malformed flag with '='"));
+    EXPECT_TRUE(Success(parser.Parse(sizeof(args) / sizeof(args[0]), args)));
```
Keenuts, line 173: `--b=` is allowed a valid flag value?
Seems like a new behavior?

Keenuts (commented):
LGTM overall, just one behavior change which seems surprising

src/test/command_line_parser_test.cpp
Keenuts: Ahh, I see, it's to handle --something="" as valid value, app should validate.

Keenuts (approved):

PR #283, opened 2023-09-01
Add knob for Vertex Buffer Format

elviscapiaq:
- It was added a new Vertex Buffer that stores 16-bit floating point format for the vertex input including position, texCoord, normal and tangent.

### DIFF
--- a/benchmarks/graphics_pipeline/main.cpp
+++ b/benchmarks/graphics_pipeline/main.cpp
@@ -93,7 +93,13 @@ static constexpr std::array<const char*, 3> kAvailablePsShaders = {
     "Benchmark_PsAluBound",
     "Benchmark_PsMemBound"};
 
-static constexpr uint32_t kPipelineCount = kAvailablePsShaders.size() * kAvailableVsShaders.size();
+static constexpr std::array<const char*, 2> kAvailableVbFormats = {
+    "Low_Precision",
+    "High_Precision"};
+
+static constexpr uint32_t kPipelineCount = kAvailablePsShaders.size() * kAvailableVsShaders.size() * kAvailableVbFormats.size();
+
+static constexpr uint32_t kMeshCount = kAvailableVbFormats.size();
 
 class ProjApp
     : public ppx::Application
@@ -173,11 +179,13 @@ private:
     std::array<grfx::GraphicsPipelinePtr, kPipelineCount>         mPipelines;
     std::array<grfx::ShaderModulePtr, kAvailableVsShaders.size()> mVsShaders;
     std::array<grfx::ShaderModulePtr, kAvailablePsShaders.size()> mPsShaders;
+    std::array<grfx::MeshPtr, kMeshCount>                         mSphereMeshes;
     uint32_t                                                      mSphereIndexCount;
 
 private:
     std::shared_ptr<KnobDropdown<std::string>> pKnobVs;
     std::shared_ptr<KnobDropdown<std::string>> pKnobPs;
+    std::shared_ptr<KnobDropdown<std::string>> pKnobVbFormat;
     std::shared_ptr<KnobSlider<int>>           pSphereInstanceCount;
     std::shared_ptr<KnobSlider<int>>           pDrawCallCount;
     std::shared_ptr<KnobSlider<int>>           pNoiseQuadsCount;
@@ -267,6 +275,10 @@ void ProjApp::InitKnobs()
     pKnobPs->SetDisplayName("Pixel Shader");
     pKnobPs->SetFlagDescription("Select the pixel shader for the graphics pipeline.");
 
+    pKnobVbFormat = GetKnobManager().CreateKnob<ppx::KnobDropdown<std::string>>("vb_format", 0, kAvailableVbFormats);
+    pKnobVbFormat->SetDisplayName("Vertex Buffer Format");
+    pKnobVbFormat->SetFlagDescription("Select the format for the vertex buffer.");
+
     pSphereInstanceCount = GetKnobManager().CreateKnob<ppx::KnobSlider<int>>("sphere-count", /* defaultValue = */ 50, /* minValue = */ 1, kMaxSphereInstanceCount);
     pSphereInstanceCount->SetDisplayName("Sphere Count");
     pSphereInstanceCount->SetFlagDescription("Select the number of spheres to draw on the screen.");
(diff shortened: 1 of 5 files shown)


benchmarks/graphics_pipeline/main.cpp
```diff
@@ -379,8 +404,13 @@ void ProjApp::Setup()
         const uint32_t sphereVertexCount = mesh.GetCountPositions();
         const uint32_t sphereTriCount    = mesh.GetCountTriangles();
 
-        Geometry geo;
-        PPX_CHECKED_CALL(Geometry::Create(GeometryOptions::InterleavedU32().AddTexCoord().AddNormal().AddTangent(), &geo));
+        Geometry        lowPrecision;
+        GeometryOptions options;
+        options.IndexTypeU32().AddPosition(grfx::FORMAT_R16G16B16_FLOAT).AddTexCoord(grfx::FORMAT_R16G16_FLOAT).AddNormal(grfx::FORMAT_R16G16B16_FLOAT).AddTangent(grfx::FORMAT_R16G16B16A16_FLOAT);
```
wangra-google, line 409: i think for normal and binormal, FORMAT_R8G8B8A8_SNORM should be good enough

include/ppx/tri_mesh.h
```diff
@@ -60,6 +60,19 @@ struct TriMeshVertexData
     float3 bitangent;
 };
 
+//! @struct TriMesh16VertexData16
+//!
+//!
+struct TriMeshVertexData16
```
wangra-google, line 66: Probably better with the name `TriMeshVertexDataCompressed`

include/ppx/tri_mesh.h
```diff
@@ -60,6 +60,19 @@ struct TriMeshVertexData
...
+{
+    glm::u16vec3 position;
```
wangra-google, line 68: i think to be consistent with the other vector types, it would be better to do something like `using half3 = glm::u16vec3`

src/ppx/geometry.cpp
```diff
@@ -247,6 +260,39 @@ class VertexDataProcessorInterleaved : public VertexDataProcessorBase
         return GetVertexBufferElementCount(pGeom, kBufferIndex);
     }
 
+    virtual uint32_t AppendVertexData(Geometry* pGeom, const TriMeshVertexData16& vtx) override
+    {
+        uint32_t       startSize = GetVertexBufferSize(pGeom, kBufferIndex);
```
wangra-google, line 265: can this code block within the function being templated instead of a copy of existing code?

src/ppx/geometry.cpp
```diff
@@ -363,6 +409,36 @@ class VertexDataProcessorPositionPlanar : public VertexDataProcessorBase
         return n;
     }
 
+    virtual uint32_t AppendVertexData(Geometry* pGeom, const TriMeshVertexData16& vtx) override
+    {
+        const uint32_t n = AppendDataToVertexBuffer(pGeom, GetPositionBufferIndex(pGeom), vtx.position);
```
wangra-google, line 414: can this code block within the function being templated instead of a copy of existing code?

src/ppx/geometry.cpp
```diff
@@ -177,6 +178,18 @@ class VertexDataProcessorPlanar : public VertexDataProcessorBase
...
```
wangra-google, line 183: can this code block within the function being templated instead of a copy of existing code?

wangra-google (changes requested):

include/ppx/tri_mesh.h
elviscapiaq: Done.

include/ppx/tri_mesh.h
elviscapiaq: Done.

src/ppx/geometry.cpp
elviscapiaq: Thanks for the feedback. Updated.

src/ppx/geometry.cpp
elviscapiaq: Updated.

src/ppx/geometry.cpp
elviscapiaq: Updated.

elviscapiaq:
I updated the code such that the function `AppendVertexData `is being templated instead of copying code.
I had an issue here since the class templated function `AppendVertexData ` implements a pure virtual function in the derived classes. I found a solution to address this issue. Let me know if this make sense. Ready for review.

benchmarks/graphics_pipeline/main.cpp
elviscapiaq: Updated.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -379,8 +404,14 @@ void ProjApp::Setup()
         const uint32_t sphereVertexCount = mesh.GetCountPositions();
         const uint32_t sphereTriCount    = mesh.GetCountTriangles();
 
-        Geometry geo;
-        PPX_CHECKED_CALL(Geometry::Create(GeometryOptions::InterleavedU32().AddTexCoord().AddNormal().AddTangent(), &geo));
+        Geometry        lowPrecision;
+        GeometryOptions options;
+        options.isVertexDataCompressed = true;
+        options.IndexTypeU32().AddPosition(grfx::FORMAT_R16G16B16_FLOAT).AddTexCoord(grfx::FORMAT_R16G16_FLOAT).AddNormal(grfx::FORMAT_R8G8B8_SNORM).AddTangent(grfx::FORMAT_R8G8B8A8_SNORM);
```
wangra-google, line 410: is there a reason why you use rgb8 for normal but rgba8 for tangent? normally rgba8 should be fine. better to be aligned to 4 bytes

elviscapiaq: Updated.

elviscapiaq:
I fixed the code such that the floats are mapped to a int8. It seems when we use the `FORMAT_R8G8B8A8_SNORM` , the shader maps values (`int8` type) from `[-128, 127]` to `[-1, 1]`, so I added a function to map a float between [-1, 1] to [-128, 127]. It should work because normal and tangents have all component values between [-1, 1]. Ready for review.

wangra-google (approved):

elviscapiaq:
I rebased this PR. I think it is ready to be merged.

Keenuts:
> I rebased this PR. I think it is ready to be merged.

This PR doesn't compile for me (and the CI do break, so reassuring 😊 )

elviscapiaq:
I fixed the issue producing the tests failed and replaced the function to convert a float to a 16-bits IEEE-754 with the glm version to avoid license issues. Ready to be merged.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -844,10 +882,11 @@ void ProjApp::Render()
             frame.cmd->DrawIndexed(mSkyBox.mesh->GetIndexCount());
 
             // Draw sphere instances
-            uint32_t pipeline_index = pKnobVs->GetIndex() * kAvailablePsShaders.size() + pKnobPs->GetIndex();
-            frame.cmd->BindGraphicsPipeline(mPipelines[pipeline_index]);
-            frame.cmd->BindIndexBuffer(mSphere.mesh);
-            frame.cmd->BindVertexBuffers(mSphere.mesh);
+            const size_t vbFormatIndex = pKnobVbFormat->GetIndex();
+            const size_t pipelineIndex = pKnobVs->GetIndex() * kAvailablePsShaders.size() * kAvailableVbFormats.size() + pKnobPs->GetIndex() * kAvailableVbFormats.size() + vbFormatIndex;
```
Keenuts, line 886: "This is getting out of hand!" 🙃
We might want to revisit this in another PR. Keeping adding new indices and pipelines in the lest and computing the index depending on what we need is getting messy.

Keenuts (approved):

PR #287, opened 2023-09-05
Split rendering in graphics_pipeline into separate renderpasses 

angela28chen:
* First renderpass for sphere scene
* [optional] N renderpasses for N noise quads
* [optional] Last renderpass for` ImGui`
* Also moved end timestamp so that it does not include imgui rendering

### DIFF
--- a/benchmarks/graphics_pipeline/main.cpp
+++ b/benchmarks/graphics_pipeline/main.cpp
@@ -816,14 +816,14 @@ void ProjApp::Render()
         // Write start timestamp
         frame.cmd->WriteTimestamp(frame.timestampQuery, grfx::PIPELINE_STAGE_TOP_OF_PIPE_BIT, /* queryIndex = */ 0);
 
-        grfx::RenderPassPtr renderPass = swapchain->GetRenderPass(imageIndex);
-        PPX_ASSERT_MSG(!renderPass.IsNull(), "render pass object is null");
+        // =====================================================================
+        // Scene renderpass
+        // =====================================================================
+        grfx::RenderPassPtr currentRenderPass = swapchain->GetRenderPass(imageIndex);
+        PPX_ASSERT_MSG(!currentRenderPass.IsNull(), "render pass object is null");
 
-        // =====================================================================
-        //  Render scene
-        // =====================================================================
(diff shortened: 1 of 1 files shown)


angela28chen:
Thanks Peter for helping with load op part!

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -844,28 +844,54 @@ void ProjApp::Render()
...
                     frame.cmd->PushGraphicsConstants(mNoiseQuads.pipelineInterface, 1, &noiseQuadRandomSeed);
                     frame.cmd->Draw(4, 1, 0, 0);
                 }
-            }
+                frame.cmd->EndRenderPass();
 
-            DrawImGui(frame.cmd);
+                // Force resolve by transitioning image layout
+                frame.cmd->TransitionImageLayout(currentRenderPass->GetRenderTargetImage(0), PPX_ALL_SUBRESOURCES, grfx::RESOURCE_STATE_RENDER_TARGET, grfx::RESOURCE_STATE_SHADER_RESOURCE);
+                frame.cmd->TransitionImageLayout(currentRenderPass->GetRenderTargetImage(0), PPX_ALL_SUBRESOURCES, grfx::RESOURCE_STATE_SHADER_RESOURCE, grfx::RESOURCE_STATE_RENDER_TARGET);
```
hysw, line 872: Not sure if transition back to  RENDER_TARGET should be included in the timestamp range, since the first one should resolve the image already.

hysw (approved):

benchmarks/graphics_pipeline/main.cpp
angela28chen: Hmm, are you suggesting we should add more timestamps and do some addition/subtraction to specifically exclude all the noise quads transitions from `shader_resource -> render_target`?

hysw: I'm not suggest anything, the changes seems fine.
I'm not sure if there's any additional cost associated with the second transition, and would like someone more knowledgeable to clarify.

angela28chen:
REVIEWER NOTE:
I'm casting a wide net with reviewers since I don't know everyone's availability. I am OOO beginning on Friday so if this isn't merged before then please feel free to cherry pick / edit as necessary. Thanks!

wangra-google (approved):

PR #291, opened 2023-09-07
Add virtual to metrics related methods.

at-ninja:
This is useful for mocking during unit tests, and verifying that metrics are added as requested.

### DIFF
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -374,11 +374,13 @@ protected:
     virtual void DrawGui(){}; // Draw additional project-related information to ImGui.
 
     // Override these methods in a derived class to change the default behavior of metrics.
+    // Virtual for unit testing purposes.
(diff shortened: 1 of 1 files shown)


apazylbe (approved):

include/ppx/application.h
```diff
@@ -464,15 +464,15 @@ class Application
 
     // Returns true when a run is active, otherwise returns false.
     // See StartMetricsRun for why this wrapper is necessary.
-    bool HasActiveMetricsRun() const;
+    virtual bool HasActiveMetricsRun() const;
```
Keenuts, line 467: You should add a comment to say "it's virtual for unit testing, even if it's not visible here"
Otherwise we'll probably end up cleaning this out thinking it's virtual for no reason.

at-ninja: Done.

PR #292, opened 2023-09-08
Add the knob for Vertex Attribute Layout

elviscapiaq:
- It was added the VERTEX_ATTRIBUTE_LAYOUT_POSITION_PLANAR to the Vertex Attribute Layout.

### DIFF
--- a/benchmarks/graphics_pipeline/main.cpp
+++ b/benchmarks/graphics_pipeline/main.cpp
@@ -96,9 +96,13 @@ static constexpr std::array<const char*, 2> kAvailableVbFormats = {
     "Low_Precision",
     "High_Precision"};
 
-static constexpr uint32_t kPipelineCount = kAvailablePsShaders.size() * kAvailableVsShaders.size() * kAvailableVbFormats.size();
+static constexpr std::array<const char*, 2> kAvailableVertexAttrLayouts = {
+    "Interleaved",
+    "Position_Planar"};
 
-static constexpr uint32_t kMeshCount = kAvailableVbFormats.size();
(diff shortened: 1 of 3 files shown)


wangra-google (approved):

elviscapiaq:
I have rebased this PR and also adding new indices and pipelines, as well as various vertex buffers depending on what we need is getting messy as it was pointed in a previous PR review, so I am considering refactoring this part for the next PR where I will be adding the knob for Level of Details (LOD). However, for now, it's probably fine. Ready for review.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -648,6 +670,12 @@ void ProjApp::CreateSpherePipelines()
                 gpCreateInfo.outputState.depthStencilFormat     = GetSwapchain()->GetDepthFormat();
                 gpCreateInfo.pPipelineInterface                 = mSphere.pipelineInterface;
                 PPX_CHECKED_CALL(GetDevice()->CreateGraphicsPipeline(&gpCreateInfo, &mPipelines[pipelineIndex++]));
+
+                // Position Planar Pipeline
+                gpCreateInfo.vertexInputState.bindingCount = 2;
```
Keenuts, line 675: Something surprising, but maybe it's by design (@wangra-google for this?)
- The interleaved representation makes sense to me.
- The linear representation seems half-linear, half-interleaved:
 position & rest have their own linear spaces.
 tangent/uv/normal are interleaved in the second space.

Is that expected?

Keenuts (approved):

PR #294, opened 2023-09-12
Benchmark: Add a MultiDimensionalIndexer class

elviscapiaq:
- This class makes much easier to get the index corresponding to a given dimension indices which is the case for our graphics pipelines and meshes.


wangra-google (approved):

Keenuts (approved):
This container is a bit weird, but it is a net improvement over what we had. Thanks! 😊

PR #295, opened 2023-09-12
 Add the Level of Detail (LOD) for sphere mesh

elviscapiaq:


wangra-google (approved):

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -434,71 +489,83 @@ void ProjApp::Setup()
...
-                uint32_t v0 = PPX_VALUE_IGNORED;
-                uint32_t v1 = PPX_VALUE_IGNORED;
-                uint32_t v2 = PPX_VALUE_IGNORED;
-                mesh.GetTriangle(triIndex, v0, v1, v2);
-                lowPrecisionInterleaved.AppendIndicesTriangle(v0 + i * sphereVertexCount, v1 + i * sphereVertexCount, v2 + i * sphereVertexCount);
-                lowPrecisionPositionPlanar.AppendIndicesTriangle(v0 + i * sphereVertexCount, v1 + i * sphereVertexCount, v2 + i * sphereVertexCount);
-                highPrecisionInterleaved.AppendIndicesTriangle(v0 + i * sphereVertexCount, v1 + i * sphereVertexCount, v2 + i * sphereVertexCount);
-                highPrecisionPositionPlanar.AppendIndicesTriangle(v0 + i * sphereVertexCount, v1 + i * sphereVertexCount, v2 + i * sphereVertexCount);
+        // LODs for spheres
+        mSphereLODs.push_back(LOD{/* longitudeSegments = */ 10, /* latitudeSegments = */ 10, kAvailableLODs[0]});
```
Keenuts, line 493: Instead of having fixed sized LODs, could the segment count be a slider?
Every time you update the slider you'll have to recreate the meshes, but you'd be more flexible in choosing the subdivision level.

wangra-google: The goal for this knob here is to make the benchmark either bound by vertex input/shading (high vertex count) or bound by other stages (low vertex count), so I think it would be fine to have just fixed sized LOD for now.  We can probably push this version first, and add the segment count in the future if we find that it is required for some other tests. :)

Keenuts: Sure, okok

Keenuts (approved):

elviscapiaq:
I rebased the PR and it is ready to be merged.

PR #297, opened 2023-09-16
Graphics clear funtions

chaoticbob:
This CL adds missing clear functions for render targets and depth stencils that clear images within a render pass.

- Added clear functions for render targets and depth stencil.
- Added depth/stencil clear flag enum bits.
- Added functions to expose mip level and array layer in RTV and DSV.
- Added logic to determine if grfx::RenderPass contains any attachments with clear ops.
- Changed clear count check in grfx::CommandBuffer::BeingRenderPass to only check if there's a clear attachment.
- Added function to grfx::RenderPass to return index of an image.
- Added clear function implementation to Vulkan and D3D12.
- Modified 04_cube to use clear functions.

### DIFF
--- a/include/ppx/grfx/grfx_command.h
+++ b/include/ppx/grfx/grfx_command.h
@@ -245,7 +245,7 @@ public:
     CommandBuffer() {}
     virtual ~CommandBuffer() {}
 
-    grfx::CommandType GetCommandType() { return mCreateInfo.pPool->GetCommandType(); }
+    grfx::CommandType GetCommandType() const { return mCreateInfo.pPool->GetCommandType(); }
 
     virtual Result Begin() = 0;
     virtual Result End()   = 0;
--- a/include/ppx/grfx/grfx_command.h
+++ b/include/ppx/grfx/grfx_command.h
@@ -245,13 +245,27 @@ public:
     CommandBuffer() {}
     virtual ~CommandBuffer() {}
 
+    grfx::CommandType GetCommandType() { return mCreateInfo.pPool->GetCommandType(); }
+
     virtual Result Begin() = 0;
     virtual Result End()   = 0;
 
     void BeginRenderPass(const grfx::RenderPassBeginInfo* pBeginInfo);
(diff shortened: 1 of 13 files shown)


include/ppx/grfx/grfx_command.h
```diff
@@ -245,13 +245,27 @@ class CommandBuffer
     CommandBuffer() {}
     virtual ~CommandBuffer() {}
 
+    grfx::CommandType GetCommandType() { return mCreateInfo.pPool->GetCommandType(); }
```
wangra-google, line 248: const?

projects/04_cube/Cube.cpp
```diff
@@ -220,19 +244,22 @@ void CubeApp::Render()
...
-        beginInfo.DSVClearValue             = {1.0f, 0xFF};
+
+        // Clear RTV to greyish blue
+        grfx::RenderTargetClearValue rtvClearValue = {0.23f, 0.23f, 0.33f, 0};
+        grfx::DepthStencilClearValue dsvClearValue = {1.0f, 0xFF};
 
         frame.cmd->TransitionImageLayout(renderPass->GetRenderTargetImage(0), PPX_ALL_SUBRESOURCES, grfx::RESOURCE_STATE_PRESENT, grfx::RESOURCE_STATE_RENDER_TARGET);
         frame.cmd->BeginRenderPass(&beginInfo);
         {
+            frame.cmd->ClearRenderTarget(renderPass->GetRenderTargetImage(0), rtvClearValue);
```
wangra-google, line 261: why do we move to explicit clear instead of doing the clear in the beginrenderpass?

chaoticbob: The commit message says that this change to the sample was to test the implementation of the clear functions.

include/ppx/grfx/grfx_command.h
chaoticbob: Yes. Fixed.

wangra-google (approved):

PR #300, opened 2023-09-22
Re-factor simulation instance creation

### LINKED ISSUE #299: Allow fluid simulation class to be embedded in other apps
One of the main goals of this refactoring is to allow the simulation to use any given command buffer, device and resolution.

- Implement a static `Create` function to initialize an instance of the simulator with error handling for resource creation.
- Use an app-provided command buffer instead of having a fixed one.
- Allow access to intermediate textures.

dnovillo:
This is the first step to fix #299.  It moves the creation of the simulation instance into a static method and starts disentangling it from the application.

Further cleanups will remove the ppx::Application link and the command buffer creation.

### DIFF
--- a/projects/fluid_simulation/shaders.h
+++ b/projects/fluid_simulation/shaders.h
@@ -8,6 +8,7 @@
 #ifndef FLUID_SIMULATION_SHADERS_H
 #define FLUID_SIMULATION_SHADERS_H
 
+#include "ppx/application.h"
 #include "ppx/graphics_util.h"
 #include "ppx/grfx/grfx_buffer.h"
 #include "ppx/grfx/grfx_config.h"
@@ -19,7 +20,6 @@
 
 namespace FluidSim {
 
-class ProjApp;
 class FluidSimulation;
 
 // Pipeline interface, descriptor layout and sampler used by compute shaders.
@@ -89,7 +89,7 @@ public:
     }
 
 private:
-    ProjApp* GetApp() const;
+    ppx::Application* GetApp() const;
 
     FluidSimulation*               mSim;
     ppx::grfx::ImagePtr            mTexture;
@@ -158,7 +158,7 @@ public:
 
     const std::string& GetName() const { return mShaderFile; }
     FluidSimulation*   GetSim() const { return mSim; }
-    ProjApp*           GetApp() const;
+    ppx::Application*  GetApp() const;
     GraphicsResources* GetGraphicsResources() const;
     ComputeResources*  GetComputeResources() const;
 
--- a/projects/fluid_simulation/sim.cpp
+++ b/projects/fluid_simulation/sim.cpp
@@ -28,13 +28,13 @@ const ppx::grfx::Format kR    = ppx::grfx::FORMAT_R16_FLOAT;
 const ppx::grfx::Format kRG   = ppx::grfx::FORMAT_R16G16_FLOAT;
 const ppx::grfx::Format kRGBA = ppx::grfx::FORMAT_R16G16B16A16_FLOAT;
 
-const SimulationConfig& FluidSimulation::GetConfig() const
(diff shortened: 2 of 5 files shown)


projects/fluid_simulation/shaders.h
```diff
@@ -89,7 +89,7 @@ class Texture
     }
 
 private:
-    ProjApp* GetApp() const;
+    ppx::Application* GetApp() const;
```
chaoticbob, line 92: Is it necessary for the FLuidSim to know about app?

Also, you might be aware already, there's static method `Application::Get()` that returns the current app instance. You can use this to get both the Application and Device.

projects/fluid_simulation/sim.cpp
```diff
@@ -28,32 +28,32 @@ const ppx::grfx::Format kR    = ppx::grfx::FORMAT_R16_FLOAT;
...
+ppx::Result FluidSimulation::Initialize()
 {
     // Create descriptor pool shared by all pipelines.
     ppx::grfx::DescriptorPoolCreateInfo dpci = {};
     dpci.sampler                             = 1024;
     dpci.sampledImage                        = 1024;
     dpci.uniformBuffer                       = 1024;
     dpci.storageImage                        = 1024;
-    PPX_CHECKED_CALL(GetApp()->GetDevice()->CreateDescriptorPool(&dpci, &mDescriptorPool));
+    PPX_CHECKED_CALL(GetDevice()->CreateDescriptorPool(&dpci, &mDescriptorPool));
```
chaoticbob, line 45: This seems like it can be leaner since we know the number of simulation frames and the number of resources we'll use.

projects/fluid_simulation/sim.cpp
```diff
@@ -238,7 +238,7 @@ void FluidSimulation::InitBloomTextures()
     ppx::int2 res = GetResolution(GetConfig().pBloomResolution->GetValue());
     mBloomTexture = std::make_unique<Texture>(this, "bloom", res.x, res.y, kRGBA);
     PPX_ASSERT_MSG(mBloomTextures.empty(), "Bloom textures already initialized");
-    for (uint32_t i = 0; i < GetConfig().pBloomIterations->GetValue(); i++) {
+    for (int i = 0; i < GetConfig().pBloomIterations->GetValue(); i++) {
```
chaoticbob, line 241: Curious why the change from `uint32_t` to `int`?

projects/fluid_simulation/sim.h
```diff
@@ -67,13 +67,34 @@ class ProjApp;
 class FluidSimulation
 {
 public:
-    FluidSimulation(ProjApp* app);
-    ProjApp*                     GetApp() const { return mApp; }
-    const SimulationConfig&      GetConfig() const;
+    FluidSimulation(ppx::Application* app, ppx::grfx::DevicePtr device, ppx::uint2 resolution, const SimulationConfig& config)
```
chaoticbob, line 70: IHMO: I don't think it's necessary to store a pointer to the application. The device I can understand.

projects/fluid_simulation/shaders.h
dnovillo: No it is not, but that requires more refactoring. I wanted to make it in incremental steps to facilitate reviews.

projects/fluid_simulation/sim.cpp
dnovillo: Warning about different signedness. The knob returns int.

projects/fluid_simulation/sim.h
dnovillo: Yeah. I'll be removing application ties next.

projects/fluid_simulation/sim.cpp
dnovillo: Yup. You want this change in this PR or a subsequent one? This is the first of several cleanups.

chaoticbob (approved):

PR #301, opened 2023-09-23
Utility class updates and fixes

chaoticbob:
**NOTE**: The enums that start with `ERROR_SCENE` are not final yet, so they're not in the `ToString` method.

- Added LoadFromMemory method to Bitmap to allow bitmap creation where file data is a chunk of memory.
- Corrected missing word in ppx::Result enums and updated places that uses it.
- Added single index append for ppx::Geometry for when topology is not know ahead of time.
- Added some more objects to ScopeDestroyer.
- Added additiona quaternion header to math_config.h

### DIFF
--- a/src/ppx/geometry.cpp
+++ b/src/ppx/geometry.cpp
@@ -1073,39 +1073,39 @@ uint32_t Geometry::GetLargestBufferSize() const
     return size;
 }
 
-void Geometry::AppendIndex(uint32_t vtx)
+void Geometry::AppendIndex(uint32_t idx)
 {
     if (mCreateInfo.indexType == grfx::INDEX_TYPE_UINT16) {
-        mIndexBuffer.Append(static_cast<uint16_t>(vtx));
+        mIndexBuffer.Append(static_cast<uint16_t>(idx));
     }
     else if (mCreateInfo.indexType == grfx::INDEX_TYPE_UINT32) {
-        mIndexBuffer.Append(vtx);
+        mIndexBuffer.Append(idx);
     }
 }
 
-void Geometry::AppendIndicesTriangle(uint32_t vtx0, uint32_t vtx1, uint32_t vtx2)
+void Geometry::AppendIndicesTriangle(uint32_t idx0, uint32_t idx1, uint32_t idx2)
 {
     if (mCreateInfo.indexType == grfx::INDEX_TYPE_UINT16) {
-        mIndexBuffer.Append(static_cast<uint16_t>(vtx0));
-        mIndexBuffer.Append(static_cast<uint16_t>(vtx1));
-        mIndexBuffer.Append(static_cast<uint16_t>(vtx2));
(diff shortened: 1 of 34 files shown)


src/ppx/grfx/grfx_scope.cpp
```diff
@@ -140,5 +184,16 @@ Result ScopeDestroyer::AddObject(grfx::Queue* pParent, grfx::CommandBuffer* pObj
     return ppx::SUCCESS;
 }
 
+void ScopeDestroyer::ReleaseAll()
+{
```
wangra-google, line 188: do you need to destroy the elems before clearing?

wangra-google (approved):

src/ppx/grfx/grfx_scope.cpp
chaoticbob: This function is to release all the objects without destroying them. For cases where you're using scope destroyer if there's an error but otherwise you'd keep the objects.  There's a comment in the header file that explains it.

wangra-google: if this is the case, maybe better name it `ReleaseAllWithoutDestroy()`?

src/ppx/geometry.cpp
```diff
@@ -1073,6 +1073,16 @@ uint32_t Geometry::GetLargestBufferSize() const
     return size;
 }
 
+void Geometry::AppendIndex(uint32_t vtx)
```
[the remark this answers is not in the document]
chaoticbob: Good catch. Looks like the AppendIndice* functions use `vtx`. I've changed them all to `idx` to make it more clear.

src/ppx/grfx/grfx_scope.cpp
chaoticbob: The idiom comes from `unique_ptr` and a few other pointer classes that use release vs reset. I think in general release means to release ownership of and don't do anything else. 

https://en.cppreference.com/w/cpp/memory/unique_ptr/release

But if you really think the longer function name will help, I'll change it.

wangra-google: maybe because i am coming from game engine world, when i see Release, my first impression is releasing/destroying the resource (or at least decr ref count which might trigger a destroy), if i don't see this, I would worry: "hmm, there might be a a mem leak?", and check the code implementations to see if there is a bug :p But if the function is named *WithoutDestroy, at least I would know it is intentional to not destroy the elems here :)

But I see your point, and did a quick search in BigWheels, it seems that this is the only place to have a Release function for now, so I think as long as future Release function in BigWheels follows the same rule, it would be fine. 

So I am ok with both solutions :)

wangra-google (approved):

src/ppx/grfx/grfx_scope.cpp
chaoticbob: Understood. Thanks. If it gets too confusing later we'll change it to the more explicit name.

PR #302, opened 2023-09-28
Fixes and speed improvement for graphics_pipeline vertex buffer setup

angela28chen:
graphics_pipeline vertex buffers setup
* Changed LOD labels so that now `LOD0` has the highest detail and `LOD2` has the lowest.
* All vertex buffers used for the 4 different settings (high/low precision X interleaved/planar) are resized to their final sizes at the beginning to avoid the need for constant resizing at every append. 
* Added more logging for vertex buffer setup
* Overall setup time is cut in half, from ~70sec to ~35sec

geometry
* Added "Initial resize mode" to allow applications to have the speedier option to set a known size for the vertex buffer before appending/writing values into it, rather than resizing every time a new element is appended.

### DIFF
--- a/include/ppx/geometry.h
+++ b/include/ppx/geometry.h
@@ -59,8 +59,8 @@ struct GeometryOptions
     uint32_t                      vertexBindingCount                      = 0;
     grfx::VertexBinding           vertexBindings[PPX_MAX_VERTEX_BINDINGS] = {};
     grfx::PrimitiveTopology       primtiveTopology                        = grfx::PRIMITIVE_TOPOLOGY_TRIANGLE_LIST;
-    uint32_t                      finalIndexCount                         = 0;
-    uint32_t                      finalVertexCount                        = 0;
+    uint32_t                      maxIndexCount                           = 0;
+    uint32_t                      maxVertexCount                          = 0;
 
     // Creates a create info objects with a UINT16 or UINT32 index
     // type and position vertex attribute.
@@ -104,10 +104,10 @@ struct GeometryOptions
     GeometryOptions& AddTangent(grfx::Format format = grfx::FORMAT_R32G32B32A32_FLOAT);
     GeometryOptions& AddBitangent(grfx::Format format = grfx::FORMAT_R32G32B32_FLOAT);
 
-    // NOTE: Setting final index/vertex counts will set up index/vertex
+    // NOTE: Setting max index/vertex counts will set up index/vertex
     // buffers for "Use #2"
-    GeometryOptions& FinalIndexCount(uint32_t count);
-    GeometryOptions& FinalVertexCount(uint32_t count);
+    GeometryOptions& MaxIndexCount(uint32_t count);
+    GeometryOptions& MaxVertexCount(uint32_t count);
 
 private:
     GeometryOptions& AddAttribute(grfx::VertexSemantic semantic, grfx::Format format);
@@ -154,11 +154,11 @@ public:
     public:
         Buffer() {}
         Buffer(BufferType type, uint32_t elementSize, uint32_t knownFinalElementCount = 0)
-            : mType(type), mElementSize(elementSize), mKnownFinalElementCount(knownFinalElementCount)
+            : mType(type), mElementSize(elementSize), mKnownElementCount(knownFinalElementCount)
         {
-            if (mKnownFinalElementCount > 0) {
+            if (mKnownElementCount > 0) {
                 // Pre-determined final size of buffer
-                mData.resize(mKnownFinalElementCount * mElementSize);
+                mData.resize(mKnownElementCount * mElementSize);
             }
         }
         ~Buffer() {}
@@ -170,6 +170,7 @@ public:
         char*       GetData() { return DataPtr(mData); }
         const char* GetData() const { return DataPtr(mData); }
         uint32_t    GetElementCount() const;
+        void        SetOffset(uint32_t offset) { mOffset = offset; }
 
         // Trusts that calling code is well behaved :)
         //
@@ -178,7 +179,7 @@ public:
         {
             uint32_t sizeOfValue = static_cast<uint32_t>(sizeof(T));
 
-            if (mKnownFinalElementCount == 0) {
+            if (mKnownElementCount == 0) {
                 // Allocate storage for incoming data
                 SetSize(mOffset + sizeOfValue);
             }
@@ -196,7 +197,7 @@ public:
         {
             uint32_t sizeOfValues = count * static_cast<uint32_t>(sizeof(T));
 
-            if (mKnownFinalElementCount == 0) {
+            if (mKnownElementCount == 0) {
                 // Allocate storage for incoming data
                 SetSize(mOffset + sizeOfValues);
             }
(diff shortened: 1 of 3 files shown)


src/ppx/geometry.cpp
wangra-google, line 1: there are several places in this file that we are doing memcpy elem by elem which could be really slow especially when we have lots of data, anyway to somehow copy the entire data just once?

wangra-google:
> graphics_pipeline vertex buffers setup
> 
> * Changed LOD labels so that now `LOD0` has the highest detail and `LOD2` has the lowest.
> * All vertex buffers used for the 4 different settings (high/low precision X interleaved/planar) are resized to their final sizes at the beginning to avoid the need for constant resizing at every append.
> * Added more logging for vertex buffer setup
> * Overall setup time is cut in half, from ~70sec to ~35sec
> 
> geometry
> 
> * Added "Initial resize mode" to allow applications to have the speedier option to set a known size for the vertex buffer before appending/writing values into it, rather than resizing every time a new element is appended.

Thanks both for looking at this, and it is great that we cut the cost to half, good work! :) 
However, I feel 35sec is still a bit slow for a simple scene like this (without profiling, i feel less than 10sec would be a good loading time, but i could be wrong). Do you know which part is taking the most of the time?

include/ppx/geometry.h
```diff
@@ -261,6 +263,18 @@ class Geometry
...
 
+    // Use the following resize functions before using the append functions for speed over convenience,
+    // corresponding to Buffer "Use #1" described above.
+    // Caller is responsible for resizing Buffers to the appropriate size.
+    //
+    // Enabling the initial resize mode will indicate that initial resizing is expected,
+    // and it will change the behaviour of all the Append functions used afterwards to use
+    // the Buffer DataPtrs.
+    //
+    void EnableInitialResizeMode();
```
chaoticbob, line 274: Is there actually a use case where an application wants to switch the resize mode of a geometry object?

If there isn't, this should come in via the create info or options. It's not advisable to have operational states be changeable if they cause objects to switch personalities.

chaoticbob (changes requested):

include/ppx/geometry.h
```diff
@@ -270,11 +284,15 @@ class Geometry
...
-    uint32_t                                              mNormaBufferIndex     = PPX_VALUE_IGNORED;
+    uint32_t                                              mNormalBufferIndex    = PPX_VALUE_IGNORED;
     uint32_t                                              mColorBufferIndex     = PPX_VALUE_IGNORED;
     uint32_t                                              mTexCoordBufferIndex  = PPX_VALUE_IGNORED;
     uint32_t                                              mTangentBufferIndex   = PPX_VALUE_IGNORED;
     uint32_t                                              mBitangentBufferIndex = PPX_VALUE_IGNORED;
+
+    bool               mInitialResizeMode  = false;
+    char*              mIndexBufferDataPtr = nullptr;
+    std::vector<char*> mVertexBuffersDataPtrs;
```
chaoticbob, line 295: I think it would be better to modify the behavior of `Geometry::Buffer` to do what you're wanting instead of adding new member vars that require additional tracking of the buffers separately.

chaoticbob:
Some high level thoughts on changes to the `Geometry` class.

As mentioned in the previous commets, I think the extra member variables bring addtional tracking information that adds complexity which may not be desirable. It might be worth taking a look at modifying `Geometry::Buffer` to reserve a specific size with `std::vector::reserve`. Calls to `std::vector::resize` shouldn't reallocate if the new size is under the vector's capacity and since the `value_type` is `char` no default constructors are called. This may remove the need to modify the current `Append*` functions.

Also as @wangra-google suggested, if there are cases where one can exploit a chunk copy rather tha individual appends, that's a potential performance win. `AppendIndicesU32()` was added recently to handle cases the application knows that it's going to copy a large number of U32 indices in. Might be worth considering this for chunked vertex data as well. 

On the `Resize` functions, I think it would be safer to use element counts instead of size in bytes. The class already has information about both the index element type and the vertex element type/stride. Any `Resize` function should be able to figure out the required number of bytes be it index or vertex data. This reduces the possiblity of having a buffer over run due to accidental under sizing.

include/ppx/geometry.h
```diff
@@ -59,6 +59,8 @@ struct GeometryOptions
     uint32_t                      vertexBindingCount                      = 0;
     grfx::VertexBinding           vertexBindings[PPX_MAX_VERTEX_BINDINGS] = {};
     grfx::PrimitiveTopology       primtiveTopology                        = grfx::PRIMITIVE_TOPOLOGY_TRIANGLE_LIST;
+    uint32_t                      finalIndexCount                         = 0;
+    uint32_t                      finalVertexCount                        = 0;
```
chaoticbob, line 64: Can you help me understand what is meant by final here? Is the intention that buffer can't grow beyond the final counts?

angela28chen: Yes, the intention is that there are mainly two use cases for Geometry:
1. We know the final count, so we set these `finalIndexCount` and `finalVertexCount` values to resize the vector once at the beginning
2. We don't know the final count so buffers will have to be resized for every vertex/index append

wangra-google: if this is the case, would `maxIndexCount` and `maxVertexCount` be better?

angela28chen: Done

angela28chen:
Tracking current progress:

Debug
```
SETUP TIMING [sec]:
18.4373: Total setup
  1.33764: loading textures
  0.0009698: skybox
  0.0011379: sphere: grid, shuffle, LOD meshes
  2.64182: sphere: creating geometry
  0.0075528: sphere: single sphere
  1.17748: sphere: duplicate N times
  7.54823: sphere: write unique vertex data
  1.48535: sphere: write indices
  3.01297: sphere: create mesh from geometry
  0.0089801: meshes indexer, uniform buffers, descriptor set layout
  0.0046155: shaders
  0.0105147: create skybox and sphere pipelines
  0.0028234: fullscreen quads
1.19722: Total untimed setup
```

Release
```
SETUP TIMING [sec]: 
7.16688: Total setup
  0.733653: loading textures
  0.00061: skybox
  6.53e-05: sphere: grid, shuffle, LOD meshes
  1.56052: sphere: creating geometry
  0.0011736: sphere: single sphere
  0.771871: sphere: duplicate N times
  0.408202: sphere: write unique vertex data
  0.3108: sphere: write indices
  2.67962: sphere: create mesh from geometry
  0.0022533: meshes indexer, uniform buffers, descriptor set layout
  0.0057939: shaders
  0.0092911: create skybox and sphere pipelines
  0.0029229: fullscreen quads
0.680101: Total untimed setup
```

chaoticbob:
> Tracking current progress:
> 
> Debug
> 
> ```
> SETUP TIMING [sec]:
> 18.4373: Total setup
>   1.33764: loading textures
>   0.0009698: skybox
>   0.0011379: sphere: grid, shuffle, LOD meshes
>   2.64182: sphere: creating geometry
>   0.0075528: sphere: single sphere
>   1.17748: sphere: duplicate N times
>   7.54823: sphere: write unique vertex data
>   1.48535: sphere: write indices
>   3.01297: sphere: create mesh from geometry
>   0.0089801: meshes indexer, uniform buffers, descriptor set layout
>   0.0046155: shaders
>   0.0105147: create skybox and sphere pipelines
>   0.0028234: fullscreen quads
> 1.19722: Total untimed setup
> ```
> 
> Release
> 
> ```
> SETUP TIMING [sec]: 
> 7.16688: Total setup
>   0.733653: loading textures
>   0.00061: skybox
>   6.53e-05: sphere: grid, shuffle, LOD meshes
>   1.56052: sphere: creating geometry
>   0.0011736: sphere: single sphere
>   0.771871: sphere: duplicate N times
>   0.408202: sphere: write unique vertex data
>   0.3108: sphere: write indices
>   2.67962: sphere: create mesh from geometry
>   0.0022533: meshes indexer, uniform buffers, descriptor set layout
>   0.0057939: shaders
>   0.0092911: create skybox and sphere pipelines
>   0.0029229: fullscreen quads
> 0.680101: Total untimed setup
> ```

Is this down from the 70 originally?

angela28chen:
> Is this down from the 70 originally?

Yes, though the 70sec (debug) was timed manually. I plan to separate the timing logic from the other changes later on and I will see if I can get a before/after detailed timing report at the end.

Keenuts:
> > Is this down from the 70 originally?
> 
> Yes, though the 70sec (debug) was timed manually. I plan to separate the timing logic from the other changes later on and I will see if I can get a before/after detailed timing report at the end.

The initial code made something like 18M allocations to create those spheres. Do you know how many we do now?

angela28chen:
> The initial code made something like 18M allocations to create those spheres. Do you know how many we do now?

Since I changed it to one resize at the beginning per vector, it should be `(4 index buffers + 6 vertex buffers (interleaved/planar high/low precision buffers) * 2 (single sphere and large combined buffer) )* 3 (LODs) = 48 times resize is called`

angela28chen:
I changed this PR back to a draft since I'm trying to split the changes into
1. Geometry https://github.com/google/bigwheels/pull/309
2. Timers and timing for graphics_pipeline
3. Speed improvements for graphics_pipeline setup (current)

angela28chen:
Note: Closing, going with solution proposed in #318 instead to make fewer changes to geometry code

PR #304, opened 2023-09-29
Scene Graph + Renderer Part 1

chaoticbob:
***NOTE: ERROR_SCENE_* enums aren't final yet**

This change adds an initial implmentation of a scene graph. Each object has documentation and usage guidelines. GLTF loader and examples will follow in next PR.

Primary TL;DR;s:
  - Implementation uses std::shared_ptr for shared object referencing - for better or for worse.
  - Some scene graph objects (images, samplers, textures, materials, meshes, and nodes) can be used in standalone mode and doesn't have to be apart of a scene. This is useful for when someone wants to use GLTFs as an asset library instead of a scene.

I would like to avoid any discussions about the theoretical aspects of scene graphs. I'm aware there's  different approaches to scene graphs. This implementation favors simplicity and flexibility instead of absolute dead reckoning performance.


src/ppx/scene/scene_resource_manager.cpp
```diff
@@ -0,0 +1,131 @@
...
+ppx::Result ResourceManager::Cache(uint64_t objectId, const scene::MeshRef& object)
+{
+    return CacheObject<scene::Mesh>(objectId, object, mMeshes);
+}
+
+void ResourceManager::DestroyAll()
+{
+    for (auto& it : mImages) {
+        auto objectRef = it.second;
+        objectRef->~Image();
```
Keenuts, line 95: Why are destructors called manually?

- objectRef is a shared pointer
- if no other reference is called, the vector will call the destructors when `.clear()` is called.

I imagine you call it explicitly because you want it to be killed even is some reference exits, but is that something we want?
Isn't it even illegal to double-call a destructor? (but working by change)

include/ppx/scene/scene_scene.h
```diff
@@ -0,0 +1,111 @@
...
+    // Returns a camera node that matches name
+    scene::CameraNode* FindCameraNode(const std::string& name) const;
+    // Returns a light node that matches name
+    scene::LightNode* FindLightNode(const std::string& name) const;
+
+    ppx::Result AddNode(const scene::NodeRef& node);
+
+private:
+    template <typename NodeT>
+    NodeT* FindNodeByName(const std::string& name, const std::vector<NodeT*>& container) const
```
Keenuts, line 85: Do you plan to handle name conflicts when adding children, or is it just "find first and your problem if 2 nodes share the same name"?

include/ppx/scene/scene_resource_manager.h
```diff
@@ -0,0 +1,126 @@
...
+            return false;
+        }
+
+        outObject = (*it).second;
+        return true;
+    }
+
+    template <
+        typename ObjectT,
+        typename ObjectRefT = std::shared_ptr<ObjectT>>
```
Keenuts, line 93: I'm not sure I follow what's happening here.
It seems this class should own the objects they store.
But we pass a reference to a `shared_ptr` when storing a value. Which strongly implies "I'm sharing ownership with you, not releasing to you".

But what's stored is a shared_ptr, meaning we do share ownership.
However, this doesn't handle allocation manually, which makes me wonder why the dtor is manually called in the cleanup function down.

Keenuts (commented):
Thanks for this!

I'd like to have this scene graph abstraction in BigWheels.
I like a few thinks, like having materials implementation matching GLTF's default materials, and allowing materials to expose attributes requirements. So we can emit proper diagnostics if a mesh is lacking some when used with a specific material.

I would also like to see some resource manager using opaque handles, as BigWheels is has a fixed initialize-live-die lifecycle. So we could never recycle handles, and emit good diagnostic when accessing a previously deleted handle, or the wrong type for a handle.

Just some things that needs clarification

include/ppx/scene/scene_material.h
```diff
@@ -0,0 +1,302 @@
...
+//
+// scene::Image objects can be shared between different scene::Texture objects.
+//
+// Corresponds to GLTF's image object.
+//
+class Image
+    : public grfx::NamedObjectTrait
+{
+public:
+    Image(
```
apazylbe, line 41: Should we document that these objects "own" the pImage and pImageView and will destroy them in the destructor? Same for sampler

include/ppx/scene/scene_resource_manager.h
apazylbe: +1, is there any specific reason to use shared_ptr here rather than weak_ptr or raw pointer? since the resource manager is actually deleting the objects, rather than relying on shared_ptr

include/ppx/scene/scene_config.h
```diff
@@ -0,0 +1,193 @@
...
+    union
+    {
+        struct
+        {
+            bool texCoords : 1;
+            bool normals   : 1;
+            bool tangents  : 1;
+            bool colors    : 1;
+        } bits;
+        uint32_t mask = 0;
```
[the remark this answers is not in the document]
chaoticbob: This came up during the early days of BigWheels. The concensus was that all the compilers that BigWheels uses does the right thing even if the spec says it's undefined.

include/ppx/scene/scene_config.h
```diff
@@ -0,0 +1,193 @@
...
+        return *this;
+    }
+
+    grfx::VertexBinding GetVertexBinding() const
+    {
+        grfx::VertexBinding binding = grfx::VertexBinding(1, grfx::VERTEX_INPUT_RATE_VERTEX);
+
+        uint32_t offset = 0;
+        if (this->bits.texCoords) {
+            binding.AppendAttribute(grfx::VertexAttribute{"TEXCOORD", kVertexAttributeTexCoordLocation, grfx::FORMAT_R32G32_FLOAT, kVertexAttributeBinding, offset, grfx::VERTEX_INPUT_RATE_VERTEX});
```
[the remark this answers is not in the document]
chaoticbob: We'll add support for half floats and compressed vertex data later. It was a bit too much to add all of that on the first go. I also don't have any test data other than 32-bit floats :(

include/ppx/scene/scene_resource_manager.h
```diff
@@ -0,0 +1,126 @@
...
+//   - materials can be shared among primitive batches
+//   - mesh data can be shared among meshes
+//   - meshes can be shared among nodes
+//
+// Both scene::Scene and scene::Mesh will call ResourceManager::DestroyAll()
+// in their destructors to destroy all its references to shared objects.
+// If afterwards a shared object has an external reference, the code holding
+// the reference is responsible for the shared objct.
+//
+class ResourceManager
```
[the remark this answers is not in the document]
chaoticbob: I went back and forth on this a few times and I'm still somewhat on the fence about it. The shared_ptrs definitely make for an annoying case of having explicitly destroy app level reference in Shutdown() so the objects get destroyed before the device.

The most recent thinking for shared ownership was to allow for some objects (images, textures, materials, meshes) to live outside the scene if an application just wanted to load only it explicitly. Or if an application creates an object and wants to add it to the scene.

The other difficulty with non-shared ownership was determining if an object was used by anything at all. Like if all meshes using a material got removed. Without searching through all the meshes, there wasn't an easier way to determine if said material was uniquely owned or not.

I need to think about scoping a bit more on this.

include/ppx/scene/scene_scene.h
chaoticbob: The latter. Most 3D programs won't allow the user to name two nodes the same thing. But if someone is creating a GLTF by hand all bets are off. I can add a warning if two nodes have the same name.

include/ppx/scene/scene_resource_manager.h
chaoticbob: What I really want here is a non-owning reference count but I couldn't find a C++ pattern for it other than maybe using std::weak_ptr. The reference count is useful if an application wants to sweep delete objects that aren't being used by other objects.

My reading on std::weak_ptr was that it doesn't affect the reference count unless it was converted to std::shared_ptr for usage. 

Open to other approaches though.

include/ppx/scene/scene_material.h
```diff
@@ -0,0 +1,302 @@
...
+    virtual std::string GetIdentString() const override { return PPX_MATERIAL_IDENT_ERROR; }
+
+    virtual scene::VertexAttributeFlags GetRequiredVertexAttributes() const override;
+};
+
+// -------------------------------------------------------------------------------------------------
+
+// Unlit Material
+//
+// Implemtations should render a texture without any lighting. Base color factor
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

include/ppx/scene/scene_mesh.h
```diff
@@ -0,0 +1,131 @@
...
+    uint32_t           vertexCount  = 0;
+    ppx::AABB          boundingBox  = {};
+};
+
+// -------------------------------------------------------------------------------------------------
+
+// Mesh
+//
+// Contains all the information necessary to render a model:
+//   - geomtry data reference
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

include/ppx/scene/scene_node.h
```diff
@@ -0,0 +1,202 @@
...
+
+// Scene Graph Node
+//
+// This is the base class for scene graph nodes. It contains transform, parent, children,
+// and visbility properties. scene::Node is instantiable and can be used as a locator/empty/group
+// node that just contains children nodes.
+//
+// This node objects can also be used as standalone objects outside of a scene. Standalone
+// nodes will have neither a parent or children. Loader implementations must not populate a
+// stadalone node's parent or children if loading a standalone node.
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

include/ppx/scene/scene_node.h
```diff
@@ -0,0 +1,202 @@
...
+
+private:
+    scene::MeshRef mMesh = nullptr;
+};
+
+// -------------------------------------------------------------------------------------------------
+
+// Scene Graph Camera Node
+//
+// Scene graph nodes that contains reference contains an instannce of
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

include/ppx/scene/scene_material.h
```diff
@@ -0,0 +1,302 @@
...
+private:
+    scene::ImageRef   mImage   = nullptr;
+    scene::SamplerRef mSampler = nullptr;
+};
+
+// -------------------------------------------------------------------------------------------------
+
+// Texture View
+//
+// This struct contains a reference a texture object and the transform
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

include/ppx/scene/scene_material.h
```diff
@@ -0,0 +1,302 @@
...
+    scene::TextureView mOcclusionTextureView         = {};
+    scene::TextureView mEmissiveTextureView          = {};
+};
+
+// -------------------------------------------------------------------------------------------------
+
+// Material Factory
+//
+// Customizable factory that provides implementations of materials. An application
+// can inherit this class to provide implemtnations of materials as it sees fit.
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

include/ppx/scene/scene_mesh.h
```diff
@@ -0,0 +1,131 @@
...
+    grfx::VertexBufferView      mPositionBufferView     = {};
+    grfx::VertexBufferView      mAttributeBufferView    = {};
+};
+
+// -------------------------------------------------------------------------------------------------
+
+// Primitive Batch
+//
+// Contains all information necessary for a draw call. The material
+// reference will determine whch pipeline gets used. The offsets and
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

include/ppx/scene/scene_node.h
```diff
@@ -0,0 +1,202 @@
...
+};
+
+// -------------------------------------------------------------------------------------------------
+
+// Scene Graph Mesh Node
+//
+// Scene graph nodes that contains reference to a scene::Mesh object.
+//
+// When used as a standalone node, scene::Mesh stores a mesh and its required objects
+// as populated as by a loader.
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

include/ppx/scene/scene_scene.h
```diff
@@ -0,0 +1,111 @@
...
+namespace ppx {
+namespace scene {
+
+// Scene Graph
+//
+// Basic scene graph designed to feed into a renderer. Manages nodes and all
+// their required objects: meshes, mesh geometry data, materials, textures,
+// images, and samplers.
+//
+// See scene::Resource manager for details on object sharing among the various
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

src/ppx/scene/scene_node.cpp
```diff
@@ -0,0 +1,206 @@
...
+}
+
+void Node::SetParent(scene::Node* pNewParent)
+{
+    auto pCurrentParent = GetParent();
+    if (!IsNull(pCurrentParent)) {
+        pCurrentParent->RemoveChild(this);
+    }
+
+    mParent = pNewParent;
```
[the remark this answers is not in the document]
chaoticbob: Yep! Thanks! Fixed.

include/ppx/scene/scene_mesh.h
```diff
@@ -0,0 +1,131 @@
...
+
+#include "ppx/scene/scene_config.h"
+#include "ppx/scene/scene_resource_manager.h"
+
+namespace ppx {
+namespace scene {
+
+// Mesh Data
+//
+// Container for geomtry data and the buffer views required by
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

include/ppx/scene/scene_resource_manager.h
```diff
@@ -0,0 +1,126 @@
...
+
+// Resource Manager
+//
+// This class stores required objects for scenes and standalone meshes.
+// It also acts as a cache during scene loading to prevent loading of
+// duplicate objects.
+//
+// The resource manager acts as the external owner of all shared resources
+// for scenes and meshes. Required objects can be shared in a variety of
+// difference cases:
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

include/ppx/scene/scene_resource_manager.h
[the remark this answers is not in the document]
chaoticbob: Looked at this again and removed an accessor in scene::Scene that allows outside access to an instance of scene::ResourceManager.

The loader(s) use this class for caching. There's only a GLTF loader right now, but we'll probably have an OBJ loader at some point soon after. We could just friend these loaders, but the intent was to make it easy for other loaders in the future to work with it.

src/ppx/scene/scene_resource_manager.cpp
chaoticbob: This was left over from one of the refactors. It's been removed.

include/ppx/scene/scene_resource_manager.h
chaoticbob: dtor was left over from one of the refactors. It's been removed.

src/ppx/scene/scene_node.cpp
```diff
@@ -0,0 +1,206 @@
...
+    SetEvaluatedDirty();
+}
+
+void Node::SetRotationOrder(Transform::RotationOrder value)
+{
+    Transform::SetRotationOrder(value);
+    SetEvaluatedDirty();
+}
+
+void Node::SetParent(scene::Node* pNewParent)
```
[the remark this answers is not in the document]
chaoticbob: I looked at the logic of having an either/or case of SetParent/AddChild and it got a bit too much to handle. I simplified it so that the calling code has more responsibility of of parenting:
 - A node can only add a child node if that child node doesn't have an existing parent and node isn't in the child's subtree
 - Calling code is responsible for asking parent to remove child if it wants to reparent

src/ppx/scene/scene_node.cpp
```diff
@@ -0,0 +1,206 @@
...
+    // Set node's parent, this will remove it from the current parent
+    pNode->SetParent(this);
+
+    // Add child to current node
+    mChildren.push_back(pNode);
+
+    return ppx::SUCCESS;
+}
+
+void Node::RemoveChild(const scene::Node* pChild)
```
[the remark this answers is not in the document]
chaoticbob: Node parenting doesn't affect ownership or referene count. Scene owns all nodes, parenting is just a relationship. Standalone nodes can't have children.

include/ppx/scene/scene_material.h
chaoticbob: Done.

include/ppx/scene/scene_mesh.h
```diff
@@ -0,0 +1,131 @@
...
+    : public grfx::NamedObjectTrait
+{
+public:
+    MeshData(
+        const scene::VertexAttributeFlags& activeVertexAttributes,
+        grfx::Mesh*                        pGpuMesh);
+    virtual ~MeshData();
+
+    grfx::Mesh*                        GetGpuMesh() const { return mGpuMesh.Get(); }
+    const scene::VertexAttributeFlags& GetActiveVertexAttributes() const { return mActiveVertexAttributes; }
```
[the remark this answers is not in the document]
chaoticbob: That's makes a lot more sense, changed!

src/ppx/scene/scene_mesh.cpp
```diff
@@ -0,0 +1,124 @@
...
+    : mResourceManager(std::move(resourceManager)),
+      mMeshData(meshData),
+      mBatches(std::move(batches))
+{
+    UpdateBoundingBox();
+}
+
+Mesh::~Mesh()
+{
+    if (mResourceManager) {
```
wangra-google, line 85: a bit lost here, why `mResourceManager->DestroyAll()` in destructor of Mesh class? isn't `ResourceManager` contains resources of all meshes? Does this mesh class contains all meshes in the scene?

chaoticbob: It's explained in the comments that when a mesh is load standalone it uses a local resource manager to store the geometry data that's shared across the primitives.

src/ppx/scene/scene_mesh.cpp
```diff
@@ -0,0 +1,147 @@
...
+        mBoundingBox.Expand(batch.GetBoundingBox().GetMin());
+        mBoundingBox.Expand(batch.GetBoundingBox().GetMax());
+    }
+}
+
+std::vector<const scene::Material*> Mesh::GetMaterials() const
+{
+    std::vector<const scene::Material*> materials;
+    for (auto& batch : mBatches) {
+        if (!IsNull(batch.GetMaterial())) {
```
apazylbe, line 137: Should this be `if (IsNull(batch.GetMaterial))` since it is missing material case? Or perhaps I am confused

src/ppx/scene/scene_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+void Node::SetRotationOrder(Transform::RotationOrder value)
+{
+    Transform::SetRotationOrder(value);
+    SetEvaluatedDirty();
+}
+
+scene::Node* Node::GetChild(uint32_t index) const
+{
+    scene::Node* pChild = nullptr;
+    if (ppx::GetElement(index, mChildren, &pChild)) {
```
apazylbe, line 91: Should this negate the condition? `if (ppx::GetElement(index, mChildren, &pChild)`
`GetElement` returns true if the element was found?

apazylbe: Also do we even need to check? Or just return pChild after calling GetElement?

src/ppx/scene/scene_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+        return ppx::ERROR_UNEXPECTED_NULL_ARGUMENT;
+    }
+
+    // Cannot add self as a child
+    if (pNewChild == this) {
+        return ppx::ERROR_SCENE_INVALID_NODE_HIERARCHY;
+    }
+
+    // Cannot add child if current node is in child's subtree
+    if (IsInSubTree(this)) {
```
apazylbe, line 129: Should this be `pNewChild->IsInSubTree(this)`?

src/ppx/scene/scene_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+{
+}
+
+Node::~Node()
+{
+}
+
+void Node::SetVisible(bool visible)
+{
+    mVisible = visible;
```
apazylbe, line 34: Does this need to set visibility of the children or no?

src/ppx/scene/scene_node.cpp
```diff
@@ -0,0 +1,250 @@
...
+        auto& concatenatedMatrix = GetConcatenatedMatrix();
+        mEvaluatedMatrix         = parentEvaluatedMatrix * concatenatedMatrix;
+        mEvaluatedDirty          = false;
+    }
+    return mEvaluatedMatrix;
+}
+
+void Node::SetParent(scene::Node* pNewParent)
+{
+    mParent = pNewParent;
```
apazylbe, line 53: should this update mEvaluatedDirty?

chaoticbob: Thanks. One day I'll remember this function exists. Fixed.

src/ppx/scene/scene_node.cpp
chaoticbob: Yep. Fixed.

src/ppx/scene/scene_node.cpp
chaoticbob: We don't. Check removed.

src/ppx/scene/scene_mesh.cpp
chaoticbob: You're right. Fixed.

include/ppx/scene/scene_node.h
```diff
@@ -0,0 +1,208 @@
...
+};
+
+// -------------------------------------------------------------------------------------------------
+
+// Scene Graph Light Node
+//
+// Scene graph node that contains all properties to represent different light
+// types.
+//
+// When used as a standalone node, scene::Camera stores light information as populated
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

include/ppx/scene/scene_node.h
```diff
@@ -0,0 +1,208 @@
...
+
+private:
+    scene::MeshRef mMesh = nullptr;
+};
+
+// -------------------------------------------------------------------------------------------------
+
+// Scene Graph Camera Node
+//
+// Scene graph nodes that contains reference contains an instance of
```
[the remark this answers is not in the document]
chaoticbob: Updated comment to make sense.

src/ppx/scene/scene_node.cpp
chaoticbob: Good question. Generally it's left up to the app on what to do. Added parameter to recurse if desired.

include/ppx/scene/scene_material.h
```diff
@@ -0,0 +1,306 @@
...
+};
+
+// -------------------------------------------------------------------------------------------------
+
+// Texture View
+//
+// This struct contains a reference to a texture object and the transform
+// data that must be applied by the shader before sampling a pixel.
+//
+// scene::Texture view objects are used directly by scene::Matreial objets.
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

include/ppx/scene/scene_resource_manager.h
Keenuts: I'd be fine with the weak_ptr, as it shows there is no ownership.

src/ppx/scene/scene_material.cpp
```diff
@@ -0,0 +1,219 @@
...
+      mImageView(pImageView)
+{
+}
+
+Image::~Image()
+{
+    if (mImageView) {
+        auto pDevice = mImageView->GetDevice();
+        pDevice->DestroySampledImageView(mImageView);
+        mImageView.Reset();
```
Keenuts, line 37: this is not needed right? Since the pointer destructor will be called at the end.

include/ppx/scene/scene_resource_manager.h
chaoticbob: weak_ptr doesn't do reference counting. It's not in there yet, but eventually the scene::Scene will have functions to allow  add/remove objects outside the context of scene loading. This is where the reference count lets utility functions know if an object is used at all, for example if an application wants to sweep remove all images that aren't referenced by any scene objects.

Obviously, this doesn't have to be shared_ptr, it's just the only facility available right now.

I've removed all external access to the reference objects so there should be no leaks. I have a different design in mind using a non-owning reference count. But we can refactor to that later. There's a few other projects needing the scene + GLTF loading sooner than later.

src/ppx/scene/scene_material.cpp
chaoticbob: Left over game programmer paranoia. Removed.

src/ppx/scene/scene_material.cpp
```diff
@@ -0,0 +1,217 @@
...
+// -------------------------------------------------------------------------------------------------
+Texture::Texture(
+    const scene::ImageRef   image,
+    const scene::SamplerRef sampler)
+    : mImage(image),
+      mSampler(sampler)
+{
+}
+
+Texture::~Texture()
```
Keenuts, line 74: There is no overridden function, those virtual dtors are not mandatory AFAIK.
And if you need one, what about using `virtual ~Image() = default;` since those are empty?

src/ppx/scene/scene_material.cpp
```diff
@@ -0,0 +1,217 @@
...
+    float                    texCoordRotate,
+    float2                   texCoordScale)
+    : mTexture(texture),
+      mTexCoordTranslate(texCoordTranslate),
+      mTexCoordRotate(texCoordRotate),
+      mTexCoordScale(texCoordScale)
+{
+}
+
+TextureView::~TextureView()
```
Keenuts, line 97: same here.

include/ppx/scene/scene_material.h
```diff
@@ -0,0 +1,325 @@
...
+
+// Material Factory
+//
+// Customizable factory that provides implementations of materials. An application
+// can inherit this class to provide implementations of materials as it sees fit.
+//
+// Materials must be uniquely identifiable by their Material Ident string.
+//
+// Materials that do not take any parameters, such as the default ErrorMaterial
+// class, can the same shared copy across all of its instances.
```
Keenuts, line 305: can -> get no?

src/ppx/scene/scene_scene.cpp
```diff
@@ -0,0 +1,126 @@
...
+namespace scene {
+
+Scene::Scene(std::unique_ptr<scene::ResourceManager>&& resourceManager)
+    : mResourceManager(std::move(resourceManager))
+{
+}
+
+Scene::~Scene()
+{
+    mResourceManager->DestroyAll();
```
Keenuts, line 13: This is probably not necessary -> the manager is deleted in this dtor, which will clear the vectors and delete resources.

src/ppx/scene/scene_material.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+    if (materialIdent == PPX_MATERIAL_IDENT_UNLIT) {
+        material = scene::MakeRef(new scene::UnlitMaterial());
+    }
+    else if (materialIdent == PPX_MATERIAL_IDENT_STANDARD) {
+        material = scene::MakeRef(new scene::StandardMaterial());
+    }
+    else {
+        if (!mErrorMaterial) {
+            mErrorMaterial = scene::MakeRef(new scene::ErrorMaterial());
```
Keenuts, line 208: This bit I'm not sure about. We'll need to see how it's used, but for other materials, we return a ref this factory doesn't own anymore. But here, it's still owned.
Because we use shared_ptr, the refcount will save us if the factory gets deleted, but it might feel wrong.

Maybe this caching should be done in the resource manager & the scene loader handle this logic:
 -  iterates over GLTF nodes
 - if it has material: query the resource for the material. If not yet there, gets created.
 - if it has no material: query the RM for the error material, if not yet there, gets created.

Keenuts (changes requested):
Overall OK to move this forward.
Did a full review so a few new comments.

src/ppx/scene/scene_material.cpp
```diff
@@ -0,0 +1,217 @@
...
+    : mImage(pImage),
+      mImageView(pImageView)
+{
+}
+
+Image::~Image()
+{
+    if (mImageView) {
+        auto pDevice = mImageView->GetDevice();
+        pDevice->DestroySampledImageView(mImageView);
```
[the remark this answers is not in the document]
chaoticbob: I think that would be good and possibly fold everything under one idiom for object creation and destruction. The CreateDestroyTraits were to mirror what game engines were doing. We're past that now so, shared_ptr is more acceptable.

include/ppx/scene/scene_config.h
```diff
@@ -0,0 +1,192 @@
...
+    {
+        VertexAttributeFlags flags = {};
+        flags.bits.texCoords       = true;
+        flags.bits.normals         = true;
+        flags.bits.tangents        = true;
+        flags.bits.colors          = true;
+        return flags;
+    }
+
+    VertexAttributeFlags& TexCoords(bool enable = true)
```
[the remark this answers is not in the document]
chaoticbob: Allows someone to chain call it similar how some of the other Options and CreateInfos class work.

src/ppx/scene/scene_material.cpp
chaoticbob: This is a left over paranoia from when certain compilers/linkers had a lot of issue with classes without a fully specified destructor.

include/ppx/scene/scene_config.h
```diff
@@ -0,0 +1,192 @@
...
+            bool tangents  : 1;
+            bool colors    : 1;
+        } bits;
+        uint32_t mask = 0;
+    };
+
+    VertexAttributeFlags(uint32_t initialMask = 0)
+        : mask(initialMask) {}
+
+    ~VertexAttributeFlags() {}
```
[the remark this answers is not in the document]
chaoticbob: Removed. The rules for aggregate types a bit fuzzy to me, hopefully nothing else in the class disqualifies it.

src/ppx/scene/scene_material.cpp
```diff
@@ -0,0 +1,217 @@
...
+    : mSampler(pSampler)
+{
+}
+
+Sampler::~Sampler()
+{
+    if (mSampler) {
+        auto pDevice = mSampler->GetDevice();
+        pDevice->DestroySampler(mSampler);
+        mSampler.Reset();
```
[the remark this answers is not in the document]
chaoticbob: Removed.

include/ppx/scene/scene_config.h
```diff
@@ -0,0 +1,192 @@
...
+        {
+            bool texCoords : 1;
+            bool normals   : 1;
+            bool tangents  : 1;
+            bool colors    : 1;
+        } bits;
+        uint32_t mask = 0;
+    };
+
+    VertexAttributeFlags(uint32_t initialMask = 0)
```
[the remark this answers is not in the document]
chaoticbob: Done.

include/ppx/scene/scene_config.h
```diff
@@ -0,0 +1,192 @@
...
+    typename ObjectRefT = std::shared_ptr<ObjectT>>
+ObjectRefT MakeRef(ObjectT* pObject)
+{
+    return ObjectRefT(pObject);
+}
+
+// -------------------------------------------------------------------------------------------------
+// Helper Structs
+// -------------------------------------------------------------------------------------------------
+struct VertexAttributeFlags
```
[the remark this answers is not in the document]
chaoticbob: I don't think this is necessary, GetRequiredVertexAttributes() isn't called through specific material classes. The loader just seems everything as scene::Material. If this changes, we can revisit this.

include/ppx/scene/scene_config.h
```diff
@@ -0,0 +1,192 @@
...
+    {
+        VertexAttributeFlags flags = {};
+        flags.bits.texCoords       = false;
+        flags.bits.normals         = false;
+        flags.bits.tangents        = false;
+        flags.bits.colors          = false;
+        return flags;
+    }
+
+    static VertexAttributeFlags EnableAll()
```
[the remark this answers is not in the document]
chaoticbob: This is a good suggestion. Changed to All().

include/ppx/scene/scene_config.h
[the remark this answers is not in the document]
chaoticbob: I can make this change in a follow up PR. There's a few places in the loader that I want to be careful about when applying this change.

src/ppx/scene/scene_material.cpp
chaoticbob: Removed.

src/ppx/scene/scene_material.cpp
chaoticbob: Removed.

include/ppx/scene/scene_material.h
chaoticbob: Removed since this comment doesn't apply anymore.

src/ppx/scene/scene_scene.cpp
chaoticbob: Removed.

src/ppx/scene/scene_material.cpp
chaoticbob: Changed it so that ErrorMaterial gets a new'd copy like the other materials. I think if we see the ErroMaterial, we'll just fix the material. Don't think it will break anyone's memory budget.

include/ppx/scene/scene_scene.h
Keenuts, line 1: maybe the filename should just be `scene`?

include/ppx/scene/scene_config.h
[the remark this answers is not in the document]
Keenuts: The case I was thinking about was in the MaterialFactory, where you instantiate the object to get the attributes:

`UnlitMaterial().getRequiredVertexAttributes()`

But up to you.

Keenuts (approved):
LGTM, up to you for the last 2 points. I'll let the others review and approve to merge.

wangra-google (approved):

apazylbe (approved):

include/ppx/scene/scene_config.h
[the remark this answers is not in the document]
chaoticbob: Good point. Serves me right that my own code would make a liar out of it. When I do the refactor on exposing the bits, I'll take care of this as well.

include/ppx/scene/scene_scene.h
chaoticbob: The reasoning for all the annoying prefix is that Visual Studio gets really confused  when toggling header/source if there are two files of the same name. It's unfortunate, but this is why grfx, vk, and dx have prefixes. `scene` is generic enough that we're likely to use it again for something.

include/ppx/scene/scene_material.h
```diff
@@ -0,0 +1,325 @@
...
+#define PPX_MATERIAL_IDENT_UNLIT    "ppx_material_ident:unlit"
+#define PPX_MATERIAL_IDENT_STANDARD "ppx_material_ident:standard"
+
+namespace ppx {
+namespace scene {
+
+// Image
+//
+// This class wraps a grfx::Image and a grfx::SampledImage objects to make
+// pathing access to the GPU pipelines easier.
```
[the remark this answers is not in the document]
chaoticbob: There could probably be a better word then *pathing*. In this case it's the literal usage of pathing: images require images views to visible to pipelines (via descriptor sets).

src/ppx/scene/scene_material.cpp
```diff
@@ -0,0 +1,205 @@
...
+// -------------------------------------------------------------------------------------------------
+Texture::Texture(
+    const scene::ImageRef   image,
+    const scene::SamplerRef sampler)
+    : mImage(image),
+      mSampler(sampler)
+{
+}
+
+// -------------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
chaoticbob: I don't thinks. If you're asking why this isn't just in the header. It's for two reasons: 1) reduce potential build time on incremental builds 2) VS debugger can every now and then has a hard time with header breakpoints.

src/ppx/scene/scene_material.cpp
```diff
@@ -0,0 +1,205 @@
...
+    float                    texCoordRotate,
+    float2                   texCoordScale)
+    : mTexture(texture),
+      mTexCoordTranslate(texCoordTranslate),
+      mTexCoordRotate(texCoordRotate),
+      mTexCoordScale(texCoordScale)
+{
+}
+
+// -------------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
chaoticbob: See above.

src/ppx/scene/scene_mesh.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+namespace ppx {
+namespace scene {
+
+// -------------------------------------------------------------------------------------------------
+// MeshData
+// -------------------------------------------------------------------------------------------------
+MeshData::MeshData(
+    const scene::VertexAttributeFlags& availableVertexAttributes,
+    grfx::Mesh*                        pGpuMesh)
```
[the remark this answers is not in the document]
chaoticbob: It's because of the destroy pattern that we have in grfx::Device. It's still heavily raw pointer based because there was a desire to have on demand destroy (because of Vulkan). There was a larger discussion earlier in the PR regarding if we should change that to something like shared_ptr.

src/ppx/scene/scene_mesh.cpp
```diff
@@ -0,0 +1,142 @@
...
+        mBoundingBox.Expand(batch.GetBoundingBox().GetMax());
+    }
+}
+
+std::vector<const scene::Material*> Mesh::GetMaterials() const
+{
+    std::vector<const scene::Material*> materials;
+    for (auto& batch : mBatches) {
+        if (IsNull(batch.GetMaterial())) {
+            // @TODO: need a better way to handle missing materials
```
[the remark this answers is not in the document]
chaoticbob: It's not necessarily an error, possibly a warning. Need to see how it plays out when we do encounter thise case. This case shuold never happen from a loader since the loader is required to set ErrorMaterial if there's missing material. But if scenes are created dynamically, it's feasible that someone could accidentally not assign a material.

src/ppx/camera.cpp
```diff
@@ -38,7 +38,7 @@ void Camera::LookAt(const float3& eye, const float3& target, const float3& up)
     mEyePosition          = eye;
     mTarget               = target;
     mWorldUp              = up;
-    mViewDirection        = glm::normalize(mEyePosition - mTarget);
+    mViewDirection        = glm::normalize(mTarget - mEyePosition);
```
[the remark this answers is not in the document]
chaoticbob: Yeah, it came from my terrible rebase that I was complaining about.

PR #309, opened 2023-10-05
Changes to Buffer and Geometry to accommodate faster bulk writes

angela28chen:
Changes:

* Added `maxIndexCount` and `maxVertexCount` to `GeometryOptions` to allow caller to set pre-determined sizes and only do one initial resize of the buffer. Note: This is faster than reserving the space at the beginning and resizing the vector each append.
* Some changes to `Buffer` to accommodate pre-determined sizes and add functionality for overwriting data within the buffer.
* Fixed typo for `mNormalBufferIndex`

Changes for Buffer element count logic:
* Added `GetSizeOfData()` to report the size of data that has been written, useful for when partial elements have been written
* Changed `GetElementCount()` logic, should only be called when buffer contains an integer number of elements.
* Changed `AppendDataToVertexBuffer()` to call `GetSizeOfData()` instead of `GetElementCount()` since it is called to write partial elements.

### DIFF
--- a/include/ppx/geometry.h
+++ b/include/ppx/geometry.h
@@ -106,8 +106,8 @@ struct GeometryOptions
 
     // NOTE: Setting max index/vertex counts will set up index/vertex
     // buffers for "Use #2"
-    GeometryOptions& MaxIndexCount(uint32_t count);
-    GeometryOptions& MaxVertexCount(uint32_t count);
+    GeometryOptions& SetMaxIndexCount(uint32_t count);
+    GeometryOptions& SetMaxVertexCount(uint32_t count);
 
 private:
     GeometryOptions& AddAttribute(grfx::VertexSemantic semantic, grfx::Format format);
@@ -154,9 +154,9 @@ public:
     public:
         Buffer() {}
         Buffer(BufferType type, uint32_t elementSize, uint32_t maxElementCount = 0)
-            : mType(type), mElementSize(elementSize), mMaxElementCount(maxElementCount)
+            : mType(type), mElementSize(elementSize), mFinalSizeSet(maxElementCount > 0)
         {
-            mData.resize(mMaxElementCount * mElementSize);
+            mData.resize(maxElementCount * mElementSize);
         }
         ~Buffer() {}
 
@@ -169,68 +169,64 @@ public:
         uint32_t    GetElementCount() const;
         uint32_t    GetDataSize() const;
 
-        // Append() will automatically adjust mOffset, but it may need to be
-        // manually adjusted if Overwrite() is used
-        void SetOffset(uint32_t offset) { mOffset = offset; }
-
         // Trusts that calling code is well behaved :)
         //
+        // Write(count, pValues...) is used to write a chunk of data at specified offset
+        // If called with offset=-1, will write at the end and adjust mOffset accordingly
         template <typename T>
-        void Append(const T& value)
-        {
-            uint32_t sizeOfValue = static_cast<uint32_t>(sizeof(T));
-
-            if (mMaxElementCount == 0) {
-                // Allocate storage for incoming data
-                SetSize(mOffset + sizeOfValue);
-            }
-
-            // Copy data
-            const void* pSrc = &value;
-            void*       pDst = mData.data() + mOffset;
-            memcpy(pDst, pSrc, sizeOfValue);
-
-            mOffset += sizeOfValue;
-        }
-
-        template <typename T>
-        void Append(uint32_t count, const T* pValues)
+        void Write(uint32_t count, const T* pValues, int32_t offset = -1)
         {
             uint32_t sizeOfValues = count * static_cast<uint32_t>(sizeof(T));
+            uint32_t writeOffset  = offset == -1 ? mOffset : offset;
 
-            if (mMaxElementCount == 0) {
-                // Allocate storage for incoming data
-                SetSize(mOffset + sizeOfValues);
+            if (offset == -1 && !mFinalSizeSet) {
+                // New data is being appended, need to allocate storage
+                SetSize(writeOffset + sizeOfValues);
+            }
+            else {
+                PPX_ASSERT_MSG(writeOffset + sizeOfValues <= GetSize(), "Attempting to write values of size " << sizeOfValues << " at offset " << offset << " will overflow the buffer of size " << GetSize());
             }
 
             // Copy data
             const void* pSrc = pValues;
-            void*       pDst = mData.data() + mOffset;
+            void*       pDst = mData.data() + writeOffset;
             memcpy(pDst, pSrc, sizeOfValues);
 
-            mOffset += sizeOfValues;
+            if (offset == -1) {
+                mOffset += sizeOfValues;
+            }
         }
 
-        // If Overwrite is used to write new data, calling code will also
-        // have to SetOffset() appropriately
-        //
+        // Write() is used to write data at the specified offset
+        // If called with offset=-1, will write at the end and adjust mOffset accordingly
         template <typename T>
-        void Overwrite(const T& value, size_t offset)
+        void Write(const T& value, int32_t offset = -1)
         {
             uint32_t sizeOfValue = static_cast<uint32_t>(sizeof(T));
+            uint32_t writeOffset = offset == -1 ? mOffset : offset;
 
-            PPX_ASSERT_MSG(offset + sizeOfValue <= GetSize(), "Attempting to overwrite with value of size " << sizeOfValue << " at offset " << offset << " will overflow the buffer of size " << GetSize());
+            if (offset == -1 && !mFinalSizeSet) {
+                // New data is being appended, need to allocate storage
+                SetSize(writeOffset + sizeOfValue);
+            }
+            else {
+                PPX_ASSERT_MSG(writeOffset + sizeOfValue <= GetSize(), "Attempting to write value of size " << sizeOfValue << " at offset " << offset << " will overflow the buffer of size " << GetSize());
+            }
 
             // Copy data
             const void* pSrc = &value;
-            void*       pDst = mData.data() + offset;
+            void*       pDst = mData.data() + writeOffset;
             memcpy(pDst, pSrc, sizeOfValue);
+
+            if (offset == -1) {
+                mOffset += sizeOfValue;
+            }
         }
 
     private:
-        BufferType        mType            = BUFFER_TYPE_VERTEX;
-        uint32_t          mElementSize     = 0;
-        uint32_t          mMaxElementCount = 0;
+        BufferType        mType         = BUFFER_TYPE_VERTEX;
+        uint32_t          mElementSize  = 0;
+        bool              mFinalSizeSet = false;
         std::vector<char> mData;
         uint32_t          mOffset = 0; // bytes
     };
@@ -301,8 +297,8 @@ public:
     void AppendEdge(const WireMeshVertexData& vtx0, const WireMeshVertexData& vtx1);
 
     // Directly overwrites index/vertex buffers
-    void SetIndexBuffer(const Geometry::Buffer* indexBuffer) { mIndexBuffer = *indexBuffer; }
-    void SetVertexBuffer(const Geometry::Buffer* vertexBuffer, size_t vertexBufferIndex) { mVertexBuffers[vertexBufferIndex] = *vertexBuffer; }
+    void SetIndexBuffer(const Geometry::Buffer& indexBuffer) { mIndexBuffer = indexBuffer; }
+    void SetVertexBuffer(const Geometry::Buffer& vertexBuffer, size_t vertexBufferIndex) { mVertexBuffers[vertexBufferIndex] = vertexBuffer; }
 
 private:
     // This is intialized to point to a static var of derived class of VertexDataProcessorBase
--- a/include/ppx/geometry.h
+++ b/include/ppx/geometry.h
@@ -59,6 +59,8 @@ struct GeometryOptions
     uint32_t                      vertexBindingCount                      = 0;
     grfx::VertexBinding           vertexBindings[PPX_MAX_VERTEX_BINDINGS] = {};
     grfx::PrimitiveTopology       primtiveTopology                        = grfx::PRIMITIVE_TOPOLOGY_TRIANGLE_LIST;
+    uint32_t                      maxIndexCount                           = 0;
+    uint32_t                      maxVertexCount                          = 0;
 
     // Creates a create info objects with a UINT16 or UINT32 index
     // type and position vertex attribute.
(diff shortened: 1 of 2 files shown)


angela28chen:
I don't know if there should be unit tests for this code?

include/ppx/geometry.h
```diff
@@ -157,6 +170,11 @@ class Geometry
         char*       GetData() { return DataPtr(mData); }
         const char* GetData() const { return DataPtr(mData); }
         uint32_t    GetElementCount() const;
+        uint32_t    GetSizeOfData() const;
```
chaoticbob, line 173: Should be `GetDataSize()` to match convention elsewhere.

include/ppx/geometry.h
```diff
@@ -165,35 +183,59 @@ class Geometry
...
         }
 
     private:
-        BufferType        mType        = BUFFER_TYPE_VERTEX;
-        uint32_t          mElementSize = 0;
+        BufferType        mType              = BUFFER_TYPE_VERTEX;
+        uint32_t          mElementSize       = 0;
+        uint32_t          mKnownElementCount = 0;
         std::vector<char> mData;
+        uint32_t          mOffset = 0;
```
chaoticbob, line 238: Add a comment to help future readers know that this is measured in bytes.

include/ppx/geometry.h
```diff
@@ -165,35 +183,59 @@ class Geometry
...
+            void*       pDst = mData.data() + offset;
+            memcpy(pDst, pSrc, sizeOfValue);
         }
 
     private:
-        BufferType        mType        = BUFFER_TYPE_VERTEX;
-        uint32_t          mElementSize = 0;
+        BufferType        mType              = BUFFER_TYPE_VERTEX;
+        uint32_t          mElementSize       = 0;
+        uint32_t          mKnownElementCount = 0;
```
chaoticbob, line 236: Can this be `mMaxElementCount` to keep the names across the various classes?

include/ppx/geometry.h
```diff
@@ -146,8 +153,14 @@ class Geometry
     {
     public:
         Buffer() {}
-        Buffer(BufferType type, uint32_t elementSize)
-            : mType(type), mElementSize(elementSize) {}
+        Buffer(BufferType type, uint32_t elementSize, uint32_t knownElementCount = 0)
```
chaoticbob, line 156: Can this be `maxElementCount` to keep the names consistent across the classes?

include/ppx/geometry.h
```diff
@@ -146,8 +153,14 @@ class Geometry
...
+            : mType(type), mElementSize(elementSize), mKnownElementCount(knownElementCount)
+        {
+            if (mKnownElementCount > 0) {
+                // Pre-determined final size of buffer
+                mData.resize(mKnownElementCount * mElementSize);
+            }
```
chaoticbob, line 162: Even though the check adds clarity, `mData.resize()` will just return if you give a 0.

include/ppx/geometry.h
angela28chen: Done

include/ppx/geometry.h
angela28chen: Done

include/ppx/geometry.h
angela28chen: Done

include/ppx/geometry.h
angela28chen: Done

include/ppx/geometry.h
angela28chen: Noted, thanks. Changed

include/ppx/geometry.h
```diff
@@ -157,6 +167,11 @@ class Geometry
         char*       GetData() { return DataPtr(mData); }
         const char* GetData() const { return DataPtr(mData); }
         uint32_t    GetElementCount() const;
+        uint32_t    GetDataSize() const;
+
+        // Append() will automatically adjust mOffset, but it may need to be
+        // manually adjusted if Overwrite() is used
+        void SetOffset(uint32_t offset) { mOffset = offset; }
```
wangra-google, line 174: it is risky to expose this as public since people can call this and then call Append() which is not the desired behavior i assume? can this be part of 'Overwrite()`?

include/ppx/geometry.h
```diff
@@ -261,6 +300,10 @@ class Geometry
     void AppendTriangle(const TriMeshVertexData& vtx0, const TriMeshVertexData& vtx1, const TriMeshVertexData& vtx2);
     void AppendEdge(const WireMeshVertexData& vtx0, const WireMeshVertexData& vtx1);
 
+    // Directly overwrites index/vertex buffers
```
wangra-google, line 303: why do we use pointer here? maybe better with const&? otherwise we need to check if the pointer is nullptr or not

src/ppx/geometry.cpp
```diff
@@ -54,22 +54,23 @@ class VertexDataProcessorBase
...
     // 0 is returned if the bufferIndex is ignored
     template <typename U>
     uint32_t AppendDataToVertexBuffer(Geometry* pGeom, uint32_t bufferIndex, const U& data)
     {
         if (bufferIndex != PPX_VALUE_IGNORED) {
             PPX_ASSERT_MSG((bufferIndex >= 0) && (bufferIndex < pGeom->mVertexBuffers.size()), "buffer index is not valid");
+            uint32_t originalSize = pGeom->mVertexBuffers[bufferIndex].GetDataSize();
             pGeom->mVertexBuffers[bufferIndex].Append(data);
-            return pGeom->mVertexBuffers[bufferIndex].GetElementCount();
+            return pGeom->mVertexBuffers[bufferIndex].GetDataSize() - originalSize;
```
wangra-google, line 66: in the case of `mMaxElementCount > 0`, and since we pre reserved the max size,  `GetDataSize` would always return the max size before & after the Append(), and this function here would return 0?

wangra-google (changes requested):

include/ppx/geometry.h
```diff
@@ -146,8 +153,11 @@ class Geometry
     {
     public:
         Buffer() {}
-        Buffer(BufferType type, uint32_t elementSize)
-            : mType(type), mElementSize(elementSize) {}
+        Buffer(BufferType type, uint32_t elementSize, uint32_t maxElementCount = 0)
+            : mType(type), mElementSize(elementSize), mMaxElementCount(maxElementCount)
```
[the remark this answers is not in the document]
angela28chen: I think I had the name as "mKnownElementCount / mFinalElementCount" at some point earlier to try and mitigate that confusion, but changed it to this for consistency. 

I don't actually want to prevent people from calling Append if the value is not zero, I think there could be cases where the final size is known but the user still wants to use Append functions.

include/ppx/geometry.h
angela28chen: I'm not sure it's possible since I think `Overwrite()` can be used in different scenarios:
1. Writing new data to buffer that has already been resized at the start (mOffset needs to be adjusted) (in my example, the part where the single sphere is repeated into the larger buffer)
2. Overwriting existing data in buffer at a point earlier than mOffset (mOffset does not need to be adjusted) (in my example, the part where unique position data is overwriting the standard position data)

I understand this is risky and that's why the old flow of using Append instead of Overwrite and not setting a final size is intended as the less risky but slower option.

src/ppx/geometry.cpp
angela28chen: Actually in the case of `mMaxElementCount > 0`, the behaviour of `GetDataSize` and `Append` are changed so that `mOffset` is used to determine what the true size of the data is, not the size of the data vector. So before and after `Append()`, `mOffset` will be different, and this line will not be 0.

include/ppx/geometry.h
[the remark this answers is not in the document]
wangra-google: `mMaxElementCount` here means 2 things:
- the reserved size in bytes if it is not 0
- a bool indicate the reserved size is not used if it is 0
But if the user is not familiar with this class, by purely looking at header, the 2nd one above is not obvious, and imagine in the future if someone wants to get the max elem count, adding a new public functions to get the mMaxElementCount would be wrong. 

I guess if you don't keep `mMaxElementCount` as a member, there won't be a consistency issue? 

Let's say if `mMaxElementCount` is not zero, If people call `Append()' after `mData` reaching the max size, and there will be no resize (because `mMaxElementCount != 0`), and the source value would be copied to somewhere out of the range which might cause a mem corruption?

src/ppx/geometry.cpp
wangra-google: i see, make sense

include/ppx/geometry.h
wangra-google: I think the problem is that the user needs to dig into the implementation to understand how to use this function which is something we should avoid because this could make the class difficult to use and error-prone. 
Wondering if it is only related to `Overwrite`, can it be somehow a input arg of `Overwrite`?

include/ppx/geometry.h
```diff
@@ -154,9 +154,9 @@ class Geometry
     public:
         Buffer() {}
         Buffer(BufferType type, uint32_t elementSize, uint32_t maxElementCount = 0)
```
wangra-google, line 156: better to have some comments on what does `maxElementCount = 0` mean

include/ppx/geometry.h
```diff
@@ -169,68 +169,64 @@ class Geometry
...
+            }
         }
 
     private:
-        BufferType        mType            = BUFFER_TYPE_VERTEX;
-        uint32_t          mElementSize     = 0;
-        uint32_t          mMaxElementCount = 0;
+        BufferType        mType         = BUFFER_TYPE_VERTEX;
+        uint32_t          mElementSize  = 0;
+        bool              mFinalSizeSet = false;
```
wangra-google, line 229: - if this value is only set in constructor, you can put `const` here
- i feel `mUseFixedSize` might be better in this case. When I see `mFinalSizeSet`, i would assume the `final size` is set but might be changed. The goal here is to reduce confusion for people who are not familiar with the class, but i am open to discussions :)

include/ppx/geometry.h
```diff
@@ -169,68 +169,64 @@ class Geometry
...
         uint32_t    GetDataSize() const;
 
-        // Append() will automatically adjust mOffset, but it may need to be
-        // manually adjusted if Overwrite() is used
-        void SetOffset(uint32_t offset) { mOffset = offset; }
-
         // Trusts that calling code is well behaved :)
         //
+        // Write(count, pValues...) is used to write a chunk of data at specified offset
+        // If called with offset=-1, will write at the end and adjust mOffset accordingly
```
wangra-google, line 175: probably better with "will `append` at the end and adjust mOffset accordingly" so that it is clearer that it is appending not overwriting the value

include/ppx/geometry.h
```diff
@@ -169,68 +169,64 @@ class Geometry
...
+            if (offset == -1) {
+                mOffset += sizeOfValues;
+            }
         }
 
-        // If Overwrite is used to write new data, calling code will also
-        // have to SetOffset() appropriately
-        //
+        // Write() is used to write data at the specified offset
+        // If called with offset=-1, will write at the end and adjust mOffset accordingly
```
wangra-google, line 201: same as above

include/ppx/geometry.h
```diff
@@ -169,68 +169,64 @@ class Geometry
...
         template <typename T>
-        void Overwrite(const T& value, size_t offset)
+        void Write(const T& value, int32_t offset = -1)
```
wangra-google, line 203: Can this function call Write(1, &value, offset) instead of duplicating the code?

include/ppx/geometry.h
```diff
@@ -102,6 +104,11 @@ struct GeometryOptions
     GeometryOptions& AddTangent(grfx::Format format = grfx::FORMAT_R32G32B32A32_FLOAT);
     GeometryOptions& AddBitangent(grfx::Format format = grfx::FORMAT_R32G32B32_FLOAT);
 
+    // NOTE: Setting max index/vertex counts will set up index/vertex
+    // buffers for "Use #2"
```
apazylbe, line 108: It's kind of unclear of what "Use #2" means here, as fas as I understand it refers to line 148-151. But i think it is better to be explicit.

include/ppx/geometry.h
```diff
@@ -157,43 +167,68 @@ class Geometry
...
         uint32_t    GetElementCount() const;
+        uint32_t    GetDataSize() const;
 
         // Trusts that calling code is well behaved :)
         //
+        // Write(count, pValues...) is used to write a chunk of data at specified offset
+        // If called with offset=-1, will write at the end and adjust mOffset accordingly
         template <typename T>
-        void Append(const T& value)
+        void Write(uint32_t count, const T* pValues, int32_t offset = -1)
```
apazylbe, line 177: A few thoughts:
This breaks the pattern and uses  in the comment above the class. 
At the minimum the class comments need to be updated.
But stepping back for a second, can you accomplish the same thing with the Use #1 with just setting data and managing it directly. The comment above specifically says "Mixing them will most likely lead to undefined behavior." and I think this new function specifically combines two uses?

include/ppx/geometry.h
```diff
@@ -157,43 +167,68 @@ class Geometry
         char*       GetData() { return DataPtr(mData); }
         const char* GetData() const { return DataPtr(mData); }
         uint32_t    GetElementCount() const;
+        uint32_t    GetDataSize() const;
 
         // Trusts that calling code is well behaved :)
         //
+        // Write(count, pValues...) is used to write a chunk of data at specified offset
+        // If called with offset=-1, will write at the end and adjust mOffset accordingly
```
apazylbe, line 175: mOffset is a private field, and this is a public function. I think it should be more verbose about what happens to offset. For example "updates internal offset that points to the end of the written data" or something like that.

include/ppx/geometry.h
```diff
@@ -157,43 +167,68 @@ class Geometry
...
-            mData.resize(offset + sizeOfValue);
             // Copy data
-            const void* pSrc = &value;
-            void*       pDst = mData.data() + offset;
-            memcpy(pDst, pSrc, sizeOfValue);
+            const void* pSrc = pValues;
+            void*       pDst = mData.data() + writeOffset;
+            memcpy(pDst, pSrc, sizeOfValues);
+
+            if (offset == -1) {
```
apazylbe, line 195: It looks like mOffset is not updated if offset is provided, what happens if offset > mOffset, does the mOffset become meaningless?

include/ppx/geometry.h
```diff
@@ -261,6 +296,10 @@ class Geometry
     void AppendTriangle(const TriMeshVertexData& vtx0, const TriMeshVertexData& vtx1, const TriMeshVertexData& vtx2);
     void AppendEdge(const WireMeshVertexData& vtx0, const WireMeshVertexData& vtx1);
 
+    // Directly overwrites index/vertex buffers
+    void SetIndexBuffer(const Geometry::Buffer& indexBuffer) { mIndexBuffer = indexBuffer; }
```
apazylbe, line 300: There are implementation details on top of this class: "Implementation Notes: Do not modify the vertex buffers directly, use the Append* functions"
Is there a reason to do this?

apazylbe (changes requested):
I left some comments, but on a high level I am not sure if added complexity is justified here. Particularly for the buffer resizing, overwriting: the Buffer comment explicitly establishes two uses either to use Append and rely on the Buffer internals to manage data, or set size and get data pointer and memcpy. Could you use the buffer in the second way to get the same thing done for optimizations?

angela28chen:
Note: Closing, going with solution proposed in #318 instead to make fewer changes to geometry code

PR #312, opened 2023-10-06
geometry: change resize policy to be double

Keenuts:
The initial resize policy for the buffer was "minimalist". Only resized to match the added element.

Changed this policy to be "double the size". Every time we reach the size limit, the underlying container size is doubled.

This reduces the benchmark loading time from 30s to 15s in debug. (367M resize calls to 725).

### DIFF
--- a/include/ppx/geometry.h
+++ b/include/ppx/geometry.h
@@ -157,6 +157,15 @@ public:
         const char* GetData() const { return DataPtr(mData); }
         uint32_t    GetElementCount() const;
 
+        // Changes the buffer size to `size`.
+        // If this operations grows the vector, this class considers all the elements
+        // to be initialized. It is the caller responsibility to initialize those elements.
+        void SetSize(uint32_t size)
+        {
+            mUsedSize = size;
+            mData.resize(size);
+        }
+
         // Trusts that calling code is well behaved :)
         template <typename T>
         void Append(const T& value)
--- a/include/ppx/geometry.h
+++ b/include/ppx/geometry.h
@@ -159,23 +159,19 @@ public:
 
         // Trusts that calling code is well behaved :)
         template <typename T>
-        inline void Append(const T& value)
+        void Append(const T& value)
         {
             Append(1, &value);
         }
 
         template <typename T>
-        inline void Append(uint32_t count, const T* pValues)
+        void Append(uint32_t count, const T* pValues)
         {
             uint32_t sizeOfValues = count * static_cast<uint32_t>(sizeof(T));
-            if (mUsedSize + sizeOfValues > mData.size()) {
+            if ((mUsedSize + sizeOfValues) > mData.size()) {
                 mData.resize(std::max<size_t>(mUsedSize + sizeOfValues, mData.size() * 2));
             }
 
-            //// Current size
-            // size_t offset = mData.size();
-            //// Allocate storage for incoming data
-            // mData.resize(offset + sizeOfValues);
             //  Copy data
             const void* pSrc = pValues;
             void*       pDst = mData.data() + mUsedSize;
--- a/include/ppx/geometry.h
+++ b/include/ppx/geometry.h
@@ -152,47 +152,41 @@ public:
 
         BufferType  GetType() const { return mType; }
         uint32_t    GetElementSize() const { return mElementSize; }
-        uint32_t    GetSize() const { return CountU32(mData); }
-        void        SetSize(uint32_t size) { mData.resize(size); }
+        uint32_t    GetSize() const { return static_cast<uint32_t>(mUsedSize); }
         char*       GetData() { return DataPtr(mData); }
         const char* GetData() const { return DataPtr(mData); }
         uint32_t    GetElementCount() const;
 
         // Trusts that calling code is well behaved :)
-        //
         template <typename T>
-        void Append(const T& value)
+        inline void Append(const T& value)
         {
-            uint32_t sizeOfValue = static_cast<uint32_t>(sizeof(T));
-
-            // Current size
-            size_t offset = mData.size();
-            // Allocate storage for incoming data
-            mData.resize(offset + sizeOfValue);
-            // Copy data
-            const void* pSrc = &value;
-            void*       pDst = mData.data() + offset;
-            memcpy(pDst, pSrc, sizeOfValue);
+            Append(1, &value);
         }
 
         template <typename T>
-        void Append(uint32_t count, const T* pValues)
+        inline void Append(uint32_t count, const T* pValues)
         {
             uint32_t sizeOfValues = count * static_cast<uint32_t>(sizeof(T));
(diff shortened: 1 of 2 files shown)


Keenuts:
@angela28chen this PR is an alternative to the one you are working on.
This is a simpler solution, but it has a big drawback: we can over-allocate.
We could add a shrink-to-fit like function the app could call if that's an issue.

The main advantage is we don't change the API, this is an internal change only

angela28chen:
I think that both this and the change I'm working on can be useful. This can improve the allocation speed for applications that make use of a lot of AppendVertexData() calls, and the other one is more precise and intended for use cases where the final size is known.

Is there an advantage to not expanding the API? I was under the impression that my change would not affect previous style of usage for Geometry objects, it would just add a second case for when the size is known. Though maybe it could be slightly more confusing for new implementations?

include/ppx/geometry.h
```diff
@@ -152,47 +152,41 @@ class Geometry
...
+        inline void Append(uint32_t count, const T* pValues)
         {
             uint32_t sizeOfValues = count * static_cast<uint32_t>(sizeof(T));
-
-            // Current size
-            size_t offset = mData.size();
-            // Allocate storage for incoming data
-            mData.resize(offset + sizeOfValues);
-            // Copy data
+            if (mUsedSize + sizeOfValues > mData.size()) {
```
chaoticbob, line 171: nit: Can you add parens around the + operation for readability?

include/ppx/geometry.h
```diff
@@ -152,47 +152,41 @@ class Geometry
...
+                mData.resize(std::max<size_t>(mUsedSize + sizeOfValues, mData.size() * 2));
+            }
+
+            //// Current size
+            // size_t offset = mData.size();
+            //// Allocate storage for incoming data
+            // mData.resize(offset + sizeOfValues);
```
chaoticbob, line 178: Can this be removed since it's not used anymore?

include/ppx/geometry.h
```diff
@@ -152,47 +152,41 @@ class Geometry
...
+        uint32_t    GetSize() const { return static_cast<uint32_t>(mUsedSize); }
         char*       GetData() { return DataPtr(mData); }
         const char* GetData() const { return DataPtr(mData); }
         uint32_t    GetElementCount() const;
 
         // Trusts that calling code is well behaved :)
-        //
         template <typename T>
-        void Append(const T& value)
+        inline void Append(const T& value)
```
chaoticbob, line 162: IIRC, inline is implicit for functions that appear within the class definition.

Keenuts:
> I think that both this and the change I'm working on can be useful. This can improve the allocation speed for applications that make use of a lot of AppendVertexData() calls, and the other one is more precise and intended for use cases where the final size is known.
> 
> Is there an advantage to not expanding the API? I was under the impression that my change would not affect previous style of usage for Geometry objects, it would just add a second case for when the size is known. Though maybe it could be slightly more confusing for new implementations?

You are correct when you say that those 2 solution don't "conflict" as yes, both could live together.
It's more from on a gain vs complexity perspective:

The issue we have is that the code in debug is slow. This seems to be caused by 2 main issues:
 - the allocation count is exuberantly high: 370 millions vs 700 with this PR.
 - the small many append operations (also 370 millions).

A perf report after this PR shows most of the overhead comes from:
 - `ppx::Geometry::Buffer::GetElementCount()`
 - `ppx::Geometry::Buffer::Append`


Once the number of allocation reduced to something smaller (here 700), the other issue is the number of calls to append.
I tried rebasing this PR on top of you PR, and I don't see a big improvement (vs the initial improvement both PRs can provide which is reducing the number of allocations).

So as both seem to provide equivalent performances, this one has the benefit to be simpler (less code to maintain).
Let me know if you find different results!

include/ppx/geometry.h
Keenuts: done.

angela28chen:
> I tried rebasing this PR on top of you PR, and I don't see a big improvement (vs the initial improvement both PRs can provide which is reducing the number of allocations).

I am not sure exactly what you are comparing, could you please clarify this part? I tried to compare my optimizations (`maxElementCount` to reduce # of allocations, then creating one sphere and 
copying it N times then overwriting position data) to this PR (doubling allocation if necessary at every append) and I still notice a significant performance difference (~2x in debug). Please see attached photos.
maxbuffersizes_timing_302
doubling_timing_312

I understand that BigWheels is generally designed for simplicity over speed and that it works well for simpler benchmarks, but to me it seems like a good idea to offer some speedier alternatives for more complex benchmarks while keeping the original stuff around for the lighter benchmarks.

include/ppx/geometry.h
```diff
@@ -152,47 +152,37 @@ class Geometry
 
         BufferType  GetType() const { return mType; }
         uint32_t    GetElementSize() const { return mElementSize; }
-        uint32_t    GetSize() const { return CountU32(mData); }
-        void        SetSize(uint32_t size) { mData.resize(size); }
```
apazylbe, line 156: Any reason to remove `SetSize`? This would break Use # 1 in the comment above this class.

Keenuts: The only reason was this is not used anywhere, so untested, so I removed it.
I can either remove the comment, or restore this.

Keenuts: Angela will need this, so re-adding it.

Keenuts:
> I am not sure exactly what you are comparing, could you please clarify this part? I tried to compare my optimizations (`maxElementCount` to reduce # of allocations, then creating one sphere and copying it N times then overwriting position data) to this PR (doubling allocation if necessary at every append) and I still notice a significant performance difference (~2x in debug). Please see attached photos. 


For future reference, turns out performance were similar between this PR and the alternative.
There is a new PR (#318) that will be rebased on top of this.

include/ppx/geometry.h
```diff
@@ -152,47 +152,46 @@ class Geometry
...
         char*       GetData() { return DataPtr(mData); }
         const char* GetData() const { return DataPtr(mData); }
         uint32_t    GetElementCount() const;
 
+        // Changes the buffer size to `size`.
+        // If this operations grows the vector, this class considers all the elements
+        // to be initialized. It is the caller responsibility to initialize those elements.
+        void SetSize(uint32_t size)
+        {
+            mUsedSize = size;
```
apazylbe, line 165: Wondering if we should emit a warning if size is less than the current size? Or perhaps if the caller is going this way then we really trust them to know what they are doing?

apazylbe (approved):

include/ppx/geometry.h
Keenuts: I do see why we would want that, but it's also a valid behavior to shrink an array.
I see this buffer more as a small wrapper over vector, which allows the same resize & append behaviors as we do without warning.

PR #313, opened 2023-10-10
Scene Graph + Renderer Part 2

chaoticbob:
***NOTE: ERROR_SCENE_* enums aren't final yet**

This CL contain updates to scene related classes to help with shader arguments. The GLTF loader and samples will come after this CL.

- Made camera's near and far clip values accesible.
- Added tex coord transform calculation to texture view classes.
- Added two functions so the renderer can ask a material if it has params or textures in case it wants to use different pipeline for no params or not textures
- Added DebugMaterial
- Added functions to Scene and ResourceManger that returns an index and map to populate shader arguments


include/ppx/scene/scene_material.h
```diff
@@ -167,6 +170,12 @@ class Material
     virtual std::string GetIdentString() const = 0;
 
     virtual scene::VertexAttributeFlags GetRequiredVertexAttributes() const = 0;
+
+    // Returns true if material has paramters for shader
+    virtual bool HasParams() const { return false; }
```
dnovillo, line 175: Should this be pure virtual?  May be easier to spot problems when a derived class doesn't implement this.

include/ppx/scene/scene_material.h
```diff
@@ -190,6 +199,24 @@ class ErrorMaterial
...
+// Implementations should render position, tex coord, normal, and tangent as color output.
+//
+class DebugMaterial
+    : public scene::Material
+{
+public:
+    DebugMaterial() {}
+    virtual ~DebugMaterial() {}
+
+    virtual std::string GetIdentString() const override { return PPX_MATERIAL_IDENT_DEBUG; }
```
dnovillo, line 213: `override` is not strictly needed here, right? Given that you're actually overriding the function.  But it doesn't hurt, I guess.

include/ppx/scene/scene_resource_manager.h
```diff
@@ -46,30 +46,47 @@ class ResourceManager
...
     void DestroyAll();
 
+    // The following functions return an array of each resource in the
+    // order of iteration in the map.
+    //
+    // The results of these functions are used to build the resource
+    // and parameter arrays that are fed as arguments into the shader
+    // for indexing.
+    //
+    // The expectation is tht these functions are only called when
```
dnovillo, line 80: s/tht/that/

src/ppx/scene/scene_resource_manager.cpp
```diff
@@ -98,5 +98,41 @@ void ResourceManager::DestroyAll()
...
+        objectPtrs.push_back(it.second.get());
+    }
+    return objectPtrs;
+}
+
+std::vector<const scene::Image*> ResourceManager::GetImages() const
+{
+    std::vector<const scene::Image*> objectPtrs;
+    for (auto& it : mImages) {
+        objectPtrs.push_back(it.second.get());
```
dnovillo, line 114: This is traversing unordered maps.  Does it matter that separate invocations may change the order in which these vectors are built up?

src/ppx/scene/scene_scene.cpp
```diff
@@ -122,5 +133,50 @@ ppx::Result Scene::AddNode(scene::NodeRef&& node)
     return ppx::SUCCESS;
 }
 
+scene::ResourceArrayIndexMap<scene::Sampler> Scene::GetSamplersArrayIndexMap() const
+{
+    auto array = mResourceManager->GetSamplers();
+
+    std::unordered_map<const scene::Sampler*, uint32_t> indexMap;
+    //
```
dnovillo, line 141: Missing comment?

dnovillo (changes requested):

include/ppx/scene/scene_material.h
chaoticbob: Missing override will cause build failure on Android.

src/ppx/scene/scene_resource_manager.cpp
chaoticbob: It shouldn't matter. The  calling code uses the vector as the basis for indexing. An index map is built from that which gets fed into the shaders.

include/ppx/scene/scene_material.h
chaoticbob: I don't think that's necessary. My expectation is that if you're implementing a material, I will trust that you will know if it's going to use params, textures or both.

chaoticbob: If people feel really strongly about ADTs I'll change it.

include/ppx/scene/scene_resource_manager.h
chaoticbob: Fixed.

src/ppx/scene/scene_scene.cpp
chaoticbob: No, this is just to side step clang-format aligning the variable names so far over to the right.

include/ppx/scene/scene_material.h
Keenuts: I'm in favor of the pure virtual: when building an interface, I'm in favor or making each non-universal function pure virtual. As the default behavior here is not always subset of the correct child behavior.

src/ppx/scene/scene_material.cpp
```diff
@@ -108,6 +125,12 @@ scene::VertexAttributeFlags UnlitMaterial::GetRequiredVertexAttributes() const
     return attrFlags;
 }
 
+bool UnlitMaterial::HasTextures() const
+{
+    bool hasBaseColorTex = HasBaseColorTexture();
+    return hasBaseColorTex;
```
Keenuts, line 131: nit: `return HasBaseColorTexture()` ?

src/ppx/scene/scene_material.cpp
```diff
@@ -126,6 +149,17 @@ scene::VertexAttributeFlags StandardMaterial::GetRequiredVertexAttributes() cons
     return attrFlags;
 }
 
+bool StandardMaterial::HasTextures() const
+{
+    bool hasBaseColorTex         = HasBaseColorTexture();
```
Keenuts, line 154: that would be nice if c++ had the same 'default-is-immutable' as Rust 😥 I want to put const in front of each one of those variables, but it's noisy.

src/ppx/scene/scene_scene.cpp
```diff
@@ -122,5 +133,50 @@ ppx::Result Scene::AddNode(scene::NodeRef&& node)
     return ppx::SUCCESS;
 }
 
+scene::ResourceArrayIndexMap<scene::Sampler> Scene::GetSamplersArrayIndexMap() const
```
Keenuts, line 136: I'm not sure I follow, why is this indexing is required?

My shader resources are kept in the resource manager, so unordered map: meaning I don't get to keep the order I added them with.

But here you rebuild the index depending on the map iteration. So this index is in fact not given by the user (depending on the shader binding indices), but computed from the unordered_map.
Hence I'm a bit lost: what is this index map providing?
Also, there is one function to return the vector from the map, so index depends on the map ordering. But here you use this vector, to create another map to explicitly store those indices as value in the map? Is this to do some kind of lookup?

How do you plan to use those?

include/ppx/scene/scene_material.h
chaoticbob: I will change it to a pure virtual interface to keep you folks happy.

src/ppx/scene/scene_material.cpp
chaoticbob: Yeah, I agree. I've been adding const onto things slowly.

src/ppx/scene/scene_scene.cpp
chaoticbob: I apologize that this is a bit cryptic, but it should make more sense once the samples are in.  The idea here is to keep the bindings or array offsets agnostic since the scene can have arbitrary number of objects.

Each draw call has a push constants block that looks something like this:
```
struct Draw {
  uint instanceIndex; // This indexes into a StructuredBuffer for modeMatrix
  uint materialIndex; // This indexes into a StructuredBuffer for material params, texture index, sampler in dex
};
```

In the draw loop:
```
for (pNode in DrawableNodes) {
   instanceIndex = NodeIndexMap[pNode];
  materialIndex = MaterialIndexMap[pNode->material];
}
```
`NodeIndexMap` is the `.second` in the returned pair. It maps the pointer to a specifc index in the vector ( `.first` in the returned pair).

When populating the resource for the GPU the vector is walked because it has the index and the pointer to the resource.

include/ppx/scene/scene_material.h
chaoticbob: Functions are now pure virtual.

include/ppx/scene/scene_material.h
```diff
@@ -190,6 +199,24 @@ class ErrorMaterial
...
 
+// Debug Material
+//
+// Implementations should render position, tex coord, normal, and tangent as color output.
+//
+class DebugMaterial
+    : public scene::Material
+{
+public:
+    DebugMaterial() {}
```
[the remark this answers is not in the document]
chaoticbob: Changed all empty ctors/dtors to default.

src/ppx/scene/scene_material.cpp
chaoticbob: This is to make debugging a bit more convenient when needing to set a breakpoint to see the value. Otherwise it requires digging through the other function. I can remove it later, but would prefer it left in for now while the renderer is being written.

src/ppx/scene/scene_material.cpp
```diff
@@ -183,22 +205,25 @@ scene::VertexAttributeFlags MaterialFactory::GetRequiredVertexAttributes(const s
     return attrFlags;
 }
 
-scene::MaterialRef MaterialFactory::CreateMaterial(
+scene::Material* MaterialFactory::CreateMaterial(
```
apazylbe, line 208: Since we are switching from shader_ptr to raw pointer, its up to the callers to manage its lifetime?

include/ppx/scene/scene_scene.h
```diff
@@ -24,6 +24,9 @@
 namespace ppx {
 namespace scene {
 
+template <typename ResObjT>
+using ResourceArrayIndexMap = std::pair<std::vector<const ResObjT*>, std::unordered_map<const ResObjT*, uint32_t>>;
```
apazylbe, line 28: Could this be a struct with meaningful names and comments? 
Or at least a comment on top of this.

src/ppx/scene/scene_scene.cpp
Keenuts: Right, so you need to have an index<>resource pair kept. And because you need a contiguous `[0;n[` indices ranges at each drawcall, you use this vector: the set for quick insert/removal, the vector to get this contiguous range of indices.
And on top, the map to quickly query which index a value has.

Because the resource manager doesn't handle deletion, this could be simplified with the map + counter. When adding a value, add it to the map, and store the counter as value. Then increment counter.
If you add deletion in the future (not used in BigWheels in any project yet), then it would become more complex. But for now, this seems to be the simples option vs a set, then a vector, then a map no?

chaoticbob: Was thinking more about this. The array isn't actually needed at all after the map to index is built. The continuous range isn't a strict requirement as long as the highest index is below the max number of descriptor bindings. 

My completely untested worry about the counter is that might overflow the descriptor bindings. 

Regarding deletion, it's not a problem if you just repopulate th descriptor bindings. There's a performance cost, I can only speculate what that is right now would like to measure for sure and find out.

include/ppx/scene/scene_scene.h
chaoticbob: Simplified and added comment.

src/ppx/scene/scene_material.cpp
chaoticbob: The ref is still used in other places.

include/ppx/scene/scene_resource_manager.h
```diff
@@ -46,30 +46,47 @@ class ResourceManager
...
+    // The following functions return an array of each resource in the
+    // order of iteration in the map.
+    //
+    // The results of these functions are used to build the resource
+    // and parameter arrays that are fed as arguments into the shader
+    // for indexing.
+    //
+    // The expectation is that these functions are only called when
+    // the arguments to the shader need updating. This should only be
+    // during loads or there's a change to scene. So it should be
```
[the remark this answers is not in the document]
chaoticbob: Done

include/ppx/scene/scene_material.h
```diff
@@ -167,6 +170,12 @@ class Material
     virtual std::string GetIdentString() const = 0;
 
     virtual scene::VertexAttributeFlags GetRequiredVertexAttributes() const = 0;
+
+    // Returns true if material has paramters for shader
```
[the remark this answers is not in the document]
chaoticbob: Fixed.

src/ppx/scene/scene_scene.cpp
```diff
@@ -122,5 +163,50 @@ ppx::Result Scene::AddNode(scene::NodeRef&& node)
...
 }
 
+scene::ResourceIndexMap<scene::Sampler> Scene::GetSamplersArrayIndexMap() const
+{
+    auto array = mResourceManager->GetSamplers();
+
+    std::unordered_map<const scene::Sampler*, uint32_t> indexMap;
+    //
+    const uint32_t objectCount = CountU32(array);
+    for (uint32_t idx = 0; idx < objectCount; ++idx) {
```
Keenuts, line 173: Since the array form is not useful once we have this map, shall you change the map creation to:

```cpp
size_t index = 0;
std::unordered_map<const scene::Sampler*, uint32_t> indexMap;
for (const auto& item : mSamplers()) {
     indexMap[item] = index;
     ++index;
}
```

?

src/ppx/scene/scene_mesh.cpp
```diff
@@ -23,52 +23,107 @@ namespace scene {
...
+        attr.format                = scene::kVertexPositionFormat;
+        attr.binding               = scene::kVertexPositionBinding;
+        attr.offset                = PPX_APPEND_OFFSET_ALIGNED;
+        attr.inputRate             = grfx::VERTEX_INPUT_RATE_VERTEX;
+        attr.semantic              = grfx::VERTEX_SEMANTIC_POSITION;
+        positionBinding.AppendAttribute(attr);
+    }
+    mVertexBindings.push_back(positionBinding);
+
+    if (availableVertexAttributes.mask != 0) {
```
Keenuts, line 45: nit: you can early-return to un-nest.

src/ppx/scene/scene_material.cpp
```diff
@@ -183,22 +205,25 @@ scene::VertexAttributeFlags MaterialFactory::GetRequiredVertexAttributes(const s
...
     else if (materialIdent == PPX_MATERIAL_IDENT_STANDARD) {
-        material = scene::MakeRef(new scene::StandardMaterial());
+        pMaterial = new scene::StandardMaterial();
+    }
+    else if (materialIdent == PPX_MATERIAL_IDENT_DEBUG) {
+        pMaterial = new scene::DebugMaterial();
     }
     else {
-        material = scene::MakeRef(new scene::ErrorMaterial());
+        pMaterial = new scene::ErrorMaterial();
```
Keenuts, line 223: nit:
What about returning error if the value is `MATERIAL_IDENT_ERROR` and assert otherwise?
This way if we see an error material, we know it's an handled error, like something wrong in the GLTF scene, and not us who made a typo when using the fucntion?

Keenuts (approved):
Some nits and things I trust you to change, but overall LGTM, let's see how the next PR goes.

src/ppx/scene/scene_material.cpp
chaoticbob: I think a warning is appropriate. Missing material is valid usage, so the loader shouldn't stop loading. The default ErrorMaterial renders flat purple to indicate something went wrong visually.

src/ppx/scene/scene_scene.cpp
chaoticbob: It could, but the reason why there's all the annoying `blahCount` stuff is for debugging should a breakpoint need to be placed in there. In case we have an IDE that isn't awesome at letting us view STL containers.

chaoticbob: Sorry, I misread what you said originally and my brain processed while I was driving. Good suggestion. I've made the changes.

src/ppx/scene/scene_mesh.cpp
chaoticbob: Done.

src/ppx/scene/scene_material.cpp
chaoticbob: The loader will log a warning if there's a missing material.

dnovillo (approved):

PR #318, opened 2023-10-12
Fixes and speed improvement for graphics_pipeline setup with minimal geometry changes

angela28chen:
`Geometry` changes:
* In buffer `Append()`, when capacity is reached, it is doubled to reduce the number of allocations needed in subsequent `Append()` calls. This is slightly slower than resize, but it allows for simple direct memcpys to `mData`.
* Adding `SetIndexBuffer()` and `SetVertexBuffer()` to Geometry so that the buffers can be directly manipulated by the application for specialized optimization that cannot be done by accessing buffers through `Append()`. (example: `Geometry::GetIndexBuffer` -> `Buffer::SetSize`-> `Buffer::GetData` -> `mpmcpy` -> `Geometry::SetIndexBuffer`)

`vk_graphics_pipeline` changes:
* Fix: LOD0 has the highest detail and LOD2 has the lowest.
* A bunch of optimizations for the vertex buffers
  * Create the vertex data for a single sphere, then copy that N times to fill out the entire vertex buffer
  * Make another pass that overwrites the position data to move spheres to specific positions on grid
* Reduce repeated appends to index vectors


angela28chen:
Thank you everyone for the feedback on #309 and #312. This PR is building off of Nathan's doubling strategy (but for reserve instead of resize) in #312 and my optimizations for vk_graphics_pipeline in #302 to speed up the setup time. It has much fewer changes to the geometry code, so it should hopefully address Nathan and Aliya's concerns regarding adding unnecessary complexity to BigWheels

Debug setup timings (On Windows)
* (1:05 min) main 
* (40 sec) double resize policy #312
* (25 sec) double reserve + benchmark sphere optimizations (this PR)

Release build on Windows timing for this PR: ~8sec.
Note that these are probably a lot faster on Linux.

main
doubling
minimal_geom_changes

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -549,37 +646,44 @@ void ProjApp::Setup()
...
                     lowPrecisionInterleaved.AppendIndicesTriangle(v0 + i * sphereVertexCount, v1 + i * sphereVertexCount, v2 + i * sphereVertexCount);
-                    lowPrecisionPositionPlanar.AppendIndicesTriangle(v0 + i * sphereVertexCount, v1 + i * sphereVertexCount, v2 + i * sphereVertexCount);
                     highPrecisionInterleaved.AppendIndicesTriangle(v0 + i * sphereVertexCount, v1 + i * sphereVertexCount, v2 + i * sphereVertexCount);
-                    highPrecisionPositionPlanar.AppendIndicesTriangle(v0 + i * sphereVertexCount, v1 + i * sphereVertexCount, v2 + i * sphereVertexCount);
                 }
             }
-            // Create a giant vertex buffer to accommodate all copies of the sphere mesh
+
+            // Overwrite the position vertex buffers in the geometries
+            lowPrecisionInterleaved.SetVertexBuffer(0, lowInterleavedPositionBuffer);
```
apazylbe, line 677: Would you be able to remove the need to set these if Get*Buffer returned non-const? I think it's just SetSize that you can't call now?

include/ppx/geometry.h
```diff
@@ -182,9 +176,13 @@ class Geometry
 
             // Current size
             size_t offset = mData.size();
-            // Allocate storage for incoming data
+            // Reserve double storage if necessary
+            if ((offset + sizeOfValues) > mData.capacity()) {
+                mData.reserve(std::max<size_t>(offset + sizeOfValues, mData.size() * 2));
```
Keenuts, line 181: Why using reserve?
On linux, debug build:
 - before this PR: 30s
 - with this PR: 11s
 - with this PR, keeping the resize as in #312 : 7s

angela28chen: I thought I had a reason for this but I can't remember it anymore, so I think it's fine to use resize if the functionality is the same.

I am calling the `Buffer::SetSize()` that is removed in #312 though, so I'd prefer if that function is kept in.

Keenuts: Sure, I removed `SetSize()` because we had no users, but since you are now using it, I'll add it back.

benchmarks/graphics_pipeline/main.cpp
angela28chen: Yes, I chose to use setters and make copies of the buffers rather than changing the getters to be non-const, but I think that can be an option too. Is that preferred?

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -743,6 +847,74 @@ void ProjApp::Setup()
...
+
+    for (size_t vertexBufferIndex = 0; vertexBufferIndex < nVertexBufferCount; vertexBufferIndex++) {
+        if (vertexBufferIndex == 0 && !copyFirstBuffer) {
+            continue;
+        }
+
+        const char*      srcBufferDataPtr = srcGeom.GetVertexBuffer(vertexBufferIndex)->GetData();
+        size_t           srcBufferSize    = srcGeom.GetVertexBuffer(vertexBufferIndex)->GetSize();
+        size_t           dstBufferSize    = srcBufferSize * repeatCount;
+        Geometry::Buffer dstBuffer        = *(dstGeom.GetVertexBuffer(vertexBufferIndex));
```
wangra-google, line 878: probably Geometry::Buffer& to avoid an extra copy of dstBuffer?

benchmarks/graphics_pipeline/main.cpp
apazylbe: I think it's better to convert Get*Buffer to be non-const instead of creating copies of buffer and overriding them, it sounds like it should be a valid usage to manipulate data in the buffers, so there should geometry should allow doing that. Would that work for your use case?

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -528,18 +534,109 @@ void ProjApp::Setup()
...
+                GeometryOptions::PositionPlanarU32()
+                    .AddTexCoord()
+                    .AddNormal()
+                    .AddTangent(),
+                &highPrecisionPositionPlanarSingleSphere));
+
+            // Full geometries
             Geometry lowPrecisionInterleaved;
-            PPX_CHECKED_CALL(Geometry::Create(GeometryOptions::InterleavedU32(grfx::FORMAT_R16G16B16_FLOAT).AddTexCoord(grfx::FORMAT_R16G16_FLOAT).AddNormal(grfx::FORMAT_R8G8B8A8_SNORM).AddTangent(grfx::FORMAT_R8G8B8A8_SNORM), &lowPrecisionInterleaved));
+            PPX_CHECKED_CALL(Geometry::Create(
```
apazylbe, line 575: Could we change this slightly to improve readability?
For example:
```
Geometry lowPrecisionInterleavedSingleSphere;
Geometry lowPrecisionInterleaved;
{
     GeometryOptions lowPrecisionInterleavedOptions = GeometryOptions::InterleavedU32(grfx::FORMAT_R16G16B16_FLOAT)
                    .AddTexCoord(grfx::FORMAT_R16G16_FLOAT)
                    .AddNormal(grfx::FORMAT_R8G8B8A8_SNORM)
                    .AddTangent(grfx::FORMAT_R8G8B8A8_SNORM);
     PPX_CHECKED_CALL(Geometry::Create(lowPrecisionInterleavedOptions, lowPrecisionInterleavedSingleSphere));
     PPX_CHECKED_CALL(Geometry::Create(lowPrecisionInterleavedOptions, lowPrecisionInterleaved));
}
```
and similarly for others.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -743,6 +847,74 @@ void ProjApp::Setup()
...
+void ProjApp::OverwritePositionData(Geometry::Buffer& dstBuffer, const TriMeshVertexData& vtx, size_t elementIndex)
+{
+    size_t elementSize = dstBuffer.GetElementSize();
+    size_t offset      = elementSize * elementIndex;
+
+    char* dstBufferDataPtr = dstBuffer.GetData();
+
+    const void* pSrc = &(vtx.position);
+    void*       pDst = dstBufferDataPtr + offset;
+    memcpy(pDst, pSrc, sizeof(float3));
```
apazylbe, line 903: should this be `sizeof(vtx.position)`?

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -549,37 +646,44 @@ void ProjApp::Setup()
...
-                    vertexDataCompressed.texCoord = half2(glm::packHalf1x16(vertexData.texCoord.x), glm::packHalf1x16(vertexData.texCoord.y));
-                    vertexDataCompressed.normal   = i8vec4(MapFloatToInt8(vertexData.normal.x), MapFloatToInt8(vertexData.normal.y), MapFloatToInt8(vertexData.normal.z), MapFloatToInt8(1.0f));
-                    vertexDataCompressed.tangent  = i8vec4(MapFloatToInt8(vertexData.tangent.x), MapFloatToInt8(vertexData.tangent.y), MapFloatToInt8(vertexData.tangent.z), MapFloatToInt8(vertexData.tangent.a));
-                    lowPrecisionInterleaved.AppendVertexData(vertexDataCompressed);
-                    lowPrecisionPositionPlanar.AppendVertexData(vertexDataCompressed);
-
-                    highPrecisionInterleaved.AppendVertexData(vertexData);
-                    highPrecisionPositionPlanar.AppendVertexData(vertexData);
+
+                    OverwritePositionData(lowInterleavedPositionBuffer, vertexDataCompressed, i * sphereVertexCount + vertexIndex);
```
apazylbe, line 658: It's a little hard to understand the overall picture of what's happening and what's important, perhaps breaking into separate functions will help? or adding comments at the top or in place will help?

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -743,6 +847,74 @@ void ProjApp::Setup()
     }
 }
 
+void ProjApp::RepeatGeometryNonPositionVertexData(Geometry& dstGeom, const Geometry& srcGeom, size_t repeatCount)
+{
+    size_t nVertexBufferCount = srcGeom.GetVertexBufferCount();
+    PPX_ASSERT_MSG(nVertexBufferCount == dstGeom.GetVertexBufferCount(), "Mismatched source and destination vertex data format");
+
+    bool copyFirstBuffer;
```
apazylbe, line 855: Perhaps we could simplify this a little? 
```
PPX_ASSERT(nVertexBufferCount > 0, "Geometry cannot have 0 vertex buffers");
// If there is one interleaved (1 vb), repeat position data as well
// For position planar (2 vb) & planar (1+ vb), repeat only non-position vertex data, starting from buffer 1
size_t firstBufferToCopy = (nVertexBufferCount == 1) ? 0 : 1;

for (size_t vertexBufferIndex = firstBufferToCopy; vertexBufferIndex < nVertexBufferCount; vertexBufferIndex++) {
```

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -743,6 +847,74 @@ void ProjApp::Setup()
...
+            void*       pDst = dstBuffer.GetData() + offset;
+            memcpy(pDst, pSrc, srcBufferSize);
+            offset += srcBufferSize;
+        }
+
+        dstGeom.SetVertexBuffer(vertexBufferIndex, dstBuffer);
+    }
+}
+
+void ProjApp::OverwritePositionData(Geometry::Buffer& dstBuffer, const TriMeshVertexData& vtx, size_t elementIndex)
```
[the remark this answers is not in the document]
angela28chen: Changed to non-member function and made it into a template function

include/ppx/geometry.h
```diff
@@ -109,7 +109,10 @@ struct GeometryOptions
 //! @class Geometry
 //!
 //! Implementation Notes:
-//!   - Do not modify the vertex buffers directly, use the Append* functions
+//!   - Recommended to avoid modifying the index/vertex buffers directly and to use
+//!     the Append* functions instead (for smaller geometries especially)
+//!   - Only use SetIndexBuffer/SetVertexBuffer if direct access to buffers is needed
+//!     (as described in Use #2 of Buffer below)
```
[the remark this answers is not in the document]
angela28chen: This usage is now changed and the comment is updated accordingly

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -260,6 +260,12 @@ class ProjApp
     void SetupFullscreenQuads();
 
     void CreateFullscreenQuadsPipelines();
+
+    void RepeatGeometryNonPositionVertexData(Geometry& dstGeom, const Geometry& srcGeom, size_t repeatCount);
```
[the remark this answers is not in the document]
angela28chen: Added

benchmarks/graphics_pipeline/main.cpp
angela28chen: Obsolete due to setter changes

benchmarks/graphics_pipeline/main.cpp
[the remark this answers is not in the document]
angela28chen: Fixed

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -260,6 +260,12 @@ class ProjApp
...
+
+    void OverwritePositionData(Geometry::Buffer& dstBuffer, const TriMeshVertexData& vtx, size_t elementIndex);
```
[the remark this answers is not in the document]
angela28chen: I did some refactoring in this PR

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -528,18 +534,109 @@ void ProjApp::Setup()
             const uint32_t sphereVertexCount = mesh.GetCountPositions();
             const uint32_t sphereTriCount    = mesh.GetCountTriangles();
 
+            PPX_LOG_INFO("LOD: " << lod.name);
+            PPX_LOG_INFO("  Sphere vertex count: " << sphereVertexCount << " | triangle count: " << sphereTriCount);
```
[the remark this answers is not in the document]
angela28chen: I did some refactoring in this PR & added comments

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -528,18 +534,109 @@ void ProjApp::Setup()
...
+            PPX_CHECKED_CALL(Geometry::Create(
+                GeometryOptions::PositionPlanarU32(grfx::FORMAT_R16G16B16_FLOAT)
+                    .AddTexCoord(grfx::FORMAT_R16G16_FLOAT)
+                    .AddNormal(grfx::FORMAT_R8G8B8A8_SNORM)
+                    .AddTangent(grfx::FORMAT_R8G8B8A8_SNORM),
+                &lowPrecisionPositionPlanarSingleSphere));
+
+            Geometry highPrecisionInterleavedSingleSphere;
+            PPX_CHECKED_CALL(Geometry::Create(
+                GeometryOptions::InterleavedU32()
```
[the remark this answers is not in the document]
angela28chen: Added comments

benchmarks/graphics_pipeline/main.cpp
angela28chen: Thanks, changed

benchmarks/graphics_pipeline/main.cpp
angela28chen: Fixed

benchmarks/graphics_pipeline/main.cpp
angela28chen: Added more comments to this section along with diagrams, let me know if it's still unclear!

benchmarks/graphics_pipeline/main.cpp
angela28chen: Changed, thanks

angela28chen:
Thank you for your reviews. After seeing some comments around clarity I decided to bundle some refactoring for `main.cpp` into this PR: 
* Break down `Setup()` into smaller pieces
* Move helper functions and special classes/structs into a separate file
* Added more comments

Hopefully this isn't too annoying to review, it's a lot of code moved but I didn't change much functionally. I've also updated the PR description.

Non-trivial code changes since last review:
* In `Geometry`, using non-const getter for vertex/index buffers instead of getter+setter
* Sphere vertex buffer related helper functions are now no longer member functions

Keenuts (approved):

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -701,49 +454,296 @@ void ProjApp::Setup()
...
+
+    // LODs for spheres
+    mSphereLODs.push_back(LOD{/* longitudeSegments = */ 50, /* latitudeSegments = */ 50, kAvailableLODs[0]});
+    mSphereLODs.push_back(LOD{/* longitudeSegments = */ 20, /* latitudeSegments = */ 20, kAvailableLODs[1]});
+    mSphereLODs.push_back(LOD{/* longitudeSegments = */ 10, /* latitudeSegments = */ 10, kAvailableLODs[2]});
+    PPX_ASSERT_MSG(mSphereLODs.size() == kAvailableLODs.size(), "LODs for spheres must be the same as the available LODs");
+
+    // Create the meshes
+    uint32_t meshIndex = 0;
+    for (LOD lod : mSphereLODs) {
```
apazylbe, line 515: This is still way too much for a single function to do. We have to scroll 150+ lines to see from the end of this for loop to see how it started. The code will be read a lot more times than it is written, so would we be able to simplify organize this a bit?
- Put it into a separate file for geometry setup: for ex mesh_setup.h, mesh_setup.cpp
Add code in a new file where it only exposes a single function which returns the overall objects you need in the Application (which I think are sphere meshes):
```
static const GeometryOptions kLowPrecisionPositionPlanarOptions = GeometryOptions::PositionPlanarU32(grfx::FORMAT_R16G16B16_FLOAT)
                                                                .AddTexCoord(grfx::FORMAT_R16G16_FLOAT)
                                                                .AddNormal(grfx::FORMAT_R8G8B8A8_SNORM)
                                                                .AddTangent(grfx::FORMAT_R8G8B8A8_SNORM);

CreateSpheresForLod() {
	CreateSingleSpheresGeometries();
	PopulateSingleSpheresVertexData();
	CopySingleSphereToMultiSphere();
	UpdateSphereGeometryInPlace();

	for all meshes to create:
		grfx_util::CreateMeshFromGeometry(mesh);
}

CreateSphereMeshes() {
	// setup lods
	for (LOD lod : mSphereLods) {
		CreateSpheresForLod();
	}
}
```

or maybe even add
```
enum class PrecisionType {
    PRECISION_TYPE_LOW_PRECISION = 0;
};

enum class VertexLayoutType {
    VERTEX_LAYOUT_TYPE_INTERLEAVED = 0;
    VERTEX_LAYOUT_TYPE_PLANAR = 1;
};

GeometryOptions GetGeometryOptions(PrecisionType precisionType, VertexLayoutType vertexLayoutType) {
    
}
```

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -701,49 +454,296 @@ void ProjApp::Setup()
...
+            for (uint32_t j = 0; j < sphereVertexCount; ++j) {
+                TriMeshVertexData vertexData = {};
+                mesh.GetVertexData(j, &vertexData);
+                vertexData.position = modelMatrix * float4(vertexData.position, 1);
+
+                TriMeshVertexDataCompressed vertexDataCompressed;
+                vertexDataCompressed.position = half3(glm::packHalf1x16(vertexData.position.x), glm::packHalf1x16(vertexData.position.y), glm::packHalf1x16(vertexData.position.z));
+
+                size_t elementIndex = firstVertexOfCurrentSphere + j;
+                OverwritePositionData(lowInterleavedPositionBufferPtr, vertexDataCompressed, elementIndex);
```
apazylbe, line 644: From what I can see `lowInterleavedPositionBufferPtr` is only used here and defined 50 lines above, perhaps it is define it on top of this line, or just pass `lowPrecisionInterleaved.GetVertexBuffer(0)` or geometry object itself and have OverWritePositionData get the vertex buffer itself.
Otherwise you have to scroll far up to see where this variable is defined

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -701,49 +454,296 @@ void ProjApp::Setup()
...
+        Geometry::Buffer* highInterleavedPositionBufferPtr = highPrecisionInterleaved.GetVertexBuffer(0);
+        Geometry::Buffer* highPlanarPositionBufferPtr      = highPrecisionPositionPlanar.GetVertexBuffer(0);
+
+        // Resize empty Position Planar vertex buffers
+        lowPlanarPositionBufferPtr->SetSize(sphereVertexCount * kMaxSphereInstanceCount * lowPlanarPositionBufferPtr->GetElementSize());
+        highPlanarPositionBufferPtr->SetSize(sphereVertexCount * kMaxSphereInstanceCount * highPlanarPositionBufferPtr->GetElementSize());
+
+        // Iterate through the full vertex buffers, changing position data and appending indices.
+        //
+        // i : sphere index
```
apazylbe, line 607: I think variables like i and j can be used for simple short loops. Here I think it's more beneficial to name them what they are: sphereIndex, vertexIndex, etc in the code below. sphereIndex loop is quite long, and you can't even easily search to see where it is used.

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -701,49 +454,296 @@ void ProjApp::Setup()
...
+        // |---------------------------i(0)--------------------------| ... |--------------i(kMaxSphereInstanceCount-1)---------------|
+        //
+        for (uint32_t i = 0; i < kMaxSphereInstanceCount; i++) {
+            uint32_t index = sphereIndices[i];
+            uint32_t x     = (index % (grid.xSize * grid.ySize)) / grid.ySize;
+            uint32_t y     = index % grid.ySize;
+            uint32_t z     = index / (grid.xSize * grid.ySize);
+
+            // Model matrix to be applied to the sphere mesh
+            float4x4 modelMatrix = glm::translate(float3(x * grid.step, y * grid.step, z * grid.step));
```
apazylbe, line 630: I think even this can be a small helper function: GetModelMatrix()

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -701,49 +454,296 @@ void ProjApp::Setup()
         grfx::ShaderModuleCreateInfo shaderCreateInfo = {static_cast<uint32_t>(bytecode.size()), bytecode.data()};
         PPX_CHECKED_CALL(GetDevice()->CreateShaderModule(&shaderCreateInfo, &mPsShaders[j]));
     }
+}
 
-    CreateSpherePipelines();
+void ProjApp::SetupFullscreenQuadsResources()
```
apazylbe, line 459: Sounds like it is just shaders being created? Maybe `CreateFullscreenQuadsShaders()`?

apazylbe (changes requested):

benchmarks/graphics_pipeline/helper.h
```diff
@@ -0,0 +1,186 @@
...
+    uint32_t    latitudeSegments;
+    std::string name;
+};
+
+// ----------------------------------------------------------------------------
+// Camera
+// ----------------------------------------------------------------------------
+static constexpr float kCameraSpeed = 0.2f;
+
+class FreeCamera
```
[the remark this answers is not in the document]
wangra-google: +1 feels really weird to see this defined in the helper.h

src/ppx/geometry.cpp
```diff
@@ -1064,6 +1064,15 @@ const Geometry::Buffer* Geometry::GetVertexBuffer(uint32_t index) const
     return pBuffer;
 }
 
+Geometry::Buffer* Geometry::GetVertexBuffer(uint32_t index)
+{
+    Geometry::Buffer* pBuffer = nullptr;
+    if (IsIndexInRange(index, mVertexBuffers)) {
+        pBuffer = &mVertexBuffers[index];
+    }
+    return pBuffer;
```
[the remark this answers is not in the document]
angela28chen: Added an assert to this one and the other const version of this function

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -400,34 +242,80 @@ void ProjApp::Config(ppx::ApplicationSettings& settings)
...
+        mGraphicsPipelinesIndexer.AddDimension(kAvailablePsShaders.size());
+        mGraphicsPipelinesIndexer.AddDimension(kAvailableVbFormats.size());
+        mGraphicsPipelinesIndexer.AddDimension(kAvailableVertexAttrLayouts.size());
+    }
+
+    SetupSkyboxResources();
+    SetupSkyboxMeshes();
+    CreateSkyboxPipelines();
+
+    SetupSphereResources();
```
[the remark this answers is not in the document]
angela28chen: Changed

benchmarks/graphics_pipeline/helper.h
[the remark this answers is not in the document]
angela28chen: Done

benchmarks/graphics_pipeline/helper.h
```diff
@@ -0,0 +1,186 @@
...
+    // `mPhi` (latitude) is an angle in the range [0, pi].
+    float mTheta;
+    float mPhi;
+};
+
+// ----------------------------------------------------------------------------
+// MultiDimensional Indexer
+// ----------------------------------------------------------------------------
+
+class MultiDimensionalIndexer
```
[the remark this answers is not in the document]
angela28chen: Done

benchmarks/graphics_pipeline/main.cpp
```diff
@@ -701,49 +454,296 @@ void ProjApp::Setup()
...
+        // These planar index buffers are the same as the interleaved ones
+        *(lowPrecisionPositionPlanar.GetIndexBuffer())  = *(lowPrecisionInterleaved.GetIndexBuffer());
+        *(highPrecisionPositionPlanar.GetIndexBuffer()) = *(highPrecisionInterleaved.GetIndexBuffer());
+
+        // Create a giant vertex buffer for each vb type to accommodate all copies of the sphere mesh
+        PPX_CHECKED_CALL(grfx_util::CreateMeshFromGeometry(GetGraphicsQueue(), &lowPrecisionInterleaved, &mSphereMeshes[meshIndex++]));
+        PPX_CHECKED_CALL(grfx_util::CreateMeshFromGeometry(GetGraphicsQueue(), &lowPrecisionPositionPlanar, &mSphereMeshes[meshIndex++]));
+        PPX_CHECKED_CALL(grfx_util::CreateMeshFromGeometry(GetGraphicsQueue(), &highPrecisionInterleaved, &mSphereMeshes[meshIndex++]));
+        PPX_CHECKED_CALL(grfx_util::CreateMeshFromGeometry(GetGraphicsQueue(), &highPrecisionPositionPlanar, &mSphereMeshes[meshIndex++]));
     }
```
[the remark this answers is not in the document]
angela28chen: Noted, I added a TODO but I think I'd rather leave that alone for this PR

benchmarks/graphics_pipeline/main.cpp
angela28chen: I tried to refactor the sphere mesh code into a separate class, PTAL and let me know if this works

benchmarks/graphics_pipeline/main.cpp
angela28chen: Obsolete

benchmarks/graphics_pipeline/main.cpp
angela28chen: Fixed, kept i/j/k for diagram in comments but changed the code implementation

benchmarks/graphics_pipeline/main.cpp
angela28chen: Changed, refactored a grid class

benchmarks/graphics_pipeline/main.cpp
angela28chen: Changed

angela28chen:
I tried to refactor the code and separate pieces out into multiple other files as suggested, and I really tried to break down the sphere mesh code into smaller pieces. Please let me know if this is a more readable structure

src/ppx/geometry.cpp
```diff
@@ -1057,11 +1057,14 @@ uint32_t Geometry::GetVertexCount() const
 
 const Geometry::Buffer* Geometry::GetVertexBuffer(uint32_t index) const
```
wangra-google, line 1058: If we never return nullptr, would it be better to return Buffer& instead of Buffer*. I am always worried about possible nullptr when i see pointers.  (maybe not necessarily in this change since there might be many other places using this function)

wangra-google (approved):

src/ppx/geometry.cpp
angela28chen: Noted, I agree that it should not happen in this PR though since the function is used elsewhere.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -0,0 +1,779 @@
...
+        grfx::DescriptorSetLayoutCreateInfo layoutCreateInfo = {};
+        layoutCreateInfo.flags.bits.pushable                 = true;
+        layoutCreateInfo.bindings.push_back(grfx::DescriptorBinding(0, grfx::DESCRIPTOR_TYPE_UNIFORM_BUFFER));
+        layoutCreateInfo.bindings.push_back(grfx::DescriptorBinding(1, grfx::DESCRIPTOR_TYPE_SAMPLED_IMAGE));
+        layoutCreateInfo.bindings.push_back(grfx::DescriptorBinding(2, grfx::DESCRIPTOR_TYPE_SAMPLER));
+        PPX_CHECKED_CALL(GetDevice()->CreateDescriptorSetLayout(&layoutCreateInfo, &mSkyBox.descriptorSetLayout));
+    }
+
+    // Shaders
+    GraphicsBenchmarkApp::SetupShader("Benchmark_SkyBox.vs", &mVSSkybox);
```
apazylbe, line 194: Is "GraphicsBenchmarkApp::" necessary here?

benchmarks/graphics_pipeline/SphereMesh.cpp
```diff
@@ -0,0 +1,255 @@
...
+    PrepareFullGeometries();
+
+    // Iterate through the spheres to adjust data unique to each sphere
+    for (uint32_t sphereIndex = 0; sphereIndex < mSphereCount; sphereIndex++) {
+        WriteSpherePosition(grid, sphereIndex);
+        AppendSphereIndicesToInterleaved(sphereIndex);
+    }
+
+    // These planar index buffers are the same as the interleaved ones
+    *(mLowPlanar.GetIndexBuffer())  = *(mLowInterleaved.GetIndexBuffer());
```
apazylbe, line 79: We shouldn't use this API like this. Since we are modifying the whole object we should either use SetIndexBuffer() or use the existing API to append indices and remove non-const GetIndexBuffer(). This behaviour might be difficult to catch if something goes wrong. 

On a separate note: we are adding a non const GetVertexBuffer here that will also allow this behaviour, which perhaps should be converted to Geometry SetVertexSize() and GetVertexBufferData() instead. But we can keep that as is and revisit in the future.

angela28chen: OK, I specifically want to avoid calling `Append*()` many times here, and (correct me if I'm wrong) I don't think that it can be done safely with the chunk writing `AppendIndicesU32()` which requires `uint32_t*` when the buffer data is stored as `char` vector. 
So I'll go back to adding SetIndexBuffer() and remove non-const GetIndexBuffer() then?

benchmarks/graphics_pipeline/helper.h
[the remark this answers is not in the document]
angela28chen: Thank you for the suggestions, these were incorporated. Resolving outdated thread

benchmarks/graphics_pipeline/SphereMesh.h
```diff
@@ -0,0 +1,173 @@
...
+// Visualization of full buffers internal structure:
+// i : sphere index
+// j : vertex index within one sphere
+// k : triangle index within one sphere
+// v0, v1, v2: the three elements of triangle k
+//
+// Full VERTEX buffers contain a total of (mSphereCount * mSingleSphereVertexCount) vertices arranged like so:
+//
+// | j(0) | j(1) | ... | j(mSingleSphereVertexCount-1) | ... | j(0) | j(1) | ...  | j(mSingleSphereVertexCount-1) |
+// |-----------------------i(0)------------------------| ... |------------------i(mSphereCount)-------------------|
```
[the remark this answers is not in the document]
angela28chen: Thanks, fixed

benchmarks/graphics_pipeline/SphereMesh.h
```diff
@@ -0,0 +1,173 @@
...
+//
+// Full INDEX buffers contain a total of (mSphereCount * mSingleSphereTriCount * 3) indices arranged like so:
+//
+// | v0 | v1 | v2 |        ...        | v0 | v1 | v2 | ... | v0 | v1 | v2 |        ...        | v0 | v1 | v2 |
+// |     k(0)     | ... | k(mSingleSphereTriCount-1) | ... |     k(0)     | ... | j(mSingleSphereTriCount-1) |
```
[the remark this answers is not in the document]
angela28chen: Thanks, fixed

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
angela28chen: Thanks, removed from `SetupShader()` calls

angela28chen:
Some minor fixes, and added `SetIndexBuffer()` while removing non-const `GetIndexBuffer()`

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
```diff
@@ -0,0 +1,242 @@
...
+const grfx::Api kApi = grfx::API_DX_12_0;
+#elif defined(USE_VK)
+const grfx::Api kApi = grfx::API_VK_1_1;
+#endif
+
+static constexpr uint32_t kMaxSphereInstanceCount  = 3000;
+static constexpr uint32_t kSeed                    = 89977;
+static constexpr uint32_t kMaxFullscreenQuadsCount = 1000;
+
+static constexpr char* kShaderBaseDir = "benchmarks/shaders";
```
apazylbe, line 39: this produces a compiler warning. You need to add a const after constexpr
https://stackoverflow.com/questions/54258241/warning-iso-c-forbids-converting-a-string-constant-to-char-for-a-static-c

apazylbe (approved):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
angela28chen: Thanks, I'll change this. Did you see this warning on Linux? I don't think I am able to see this on Windows when building with VS

angela28chen:
I just realized this refactoring actually breaks the configuration: fullscreen quad set to a solid color
I'm looking into this now

wangra-google (approved):

apazylbe (approved):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
apazylbe: Yes, on linux

PR #319, opened 2023-10-12
Use vulkan descriptor arrays instead of flattening them

apazylbe:
In order to fully support resource arrays, also require and enable shader*ArrayDynamicIndexing device features.
This allows one to use push constants to select texture within an array of textures like in 24_push_constants

Does not add support for using push descriptors with arrays.

Dynamic indexing features are not required for just using shader resource arrays with constant indices.
Currently, from what I can tell, only 24_push_constants requires enabling shaderSampledImageArrayDynamicIndexing and there are some small number of gpus that don't support those features (software lavapipe).

We could handle this in a few different ways:
1) Enable dynamic indexing features across the board, fail on the devices that don't support them. This would mean that BigWheels won't run on some devices (lavapipe)
2) Let applications make sure the used features are supported, and provide a way for applications to enable those device features. 
3) Only flatten resource arrays on devices that support dynamic indexing features. This would require compiling/loading shaders conditionally, and might get messy.

### DIFF
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -17,6 +17,7 @@ jobs:
     steps:
       - name: Setup necessary packages
         run: |
+          sudo add-apt-repository ppa:kisak/kisak-mesa -y
           sudo apt update && sudo apt install libxrandr-dev libxinerama-dev libx11-dev libxcursor-dev libxi-dev libx11-xcb-dev clang \
           mesa-vulkan-drivers
       - name: Setup Vulkan SDK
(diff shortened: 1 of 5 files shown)


apazylbe:
Ah, looks like the features aren't supported for our github CI. Maybe more of an argument for option #2?

chaoticbob:
> Ah, looks like the features aren't supported for our github CI. Maybe more of an argument for option #2?

It might be a somewhat painful transition, but we should just assume those are supported. The argument for this is that it better aligns with D3D12 - in which these features are implicit. For the devices we'll care about those options will be present.

chaoticbob (commented):
LGTM so far

chaoticbob:
Tested this against my local branch and it's working as expected for Vulkan.

apazylbe:
the CI is now good after upgrading to the new mesa vulkan drivers. I will finish testing this on a few devices/projects and then set to ready to review

PR #320, opened 2023-10-16
More efficient SampleIBLTexture()

chaoticbob:
- Replaced CartesianToSpherical with obstensibly more efficient DirectionToLatLongUV

### DIFF
--- a/assets/common/shaders/ppx/Math.hlsli
+++ b/assets/common/shaders/ppx/Math.hlsli
@@ -16,8 +16,8 @@
 #define MATH_HLSLI
 
 #define EPSILON 0.00001
-#define PI      3.1415292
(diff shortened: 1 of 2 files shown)


assets/common/shaders/ppx/Math.hlsli
```diff
@@ -17,6 +17,7 @@
 
 #define EPSILON 0.00001
 #define PI      3.1415292
+#define INV_PI  (1.0 / 3.1415292)
```
wangra-google, line 20: 3.141592? the 1st `2` is a typo?

chaoticbob: Yep. Guess it's been there for a while.

Keenuts (approved):

PR #323, opened 2023-10-23
Split textures and shaders from simulation object.

dnovillo:
This narrows the sharing surface between the simulator and shader and texture classes.  Only what is strictly needed is passed into object instantiations.

- For textures: pass the device.
- For shaders: pass the device and descriptor pool.
- Remove the initializer for textures.  They are all written to by the simulation loop, so it's pointless to clear them.

There is more untanglement and simplification to come.


projects/fluid_simulation/shaders.h
```diff
@@ -58,21 +57,29 @@ class Texture
...
+
+    uint32_t GetWidth() const { return mTexture->GetWidth(); }
+    uint32_t GetHeight() const { return mTexture->GetHeight(); }
+
+    // Compute and return the size of the texture normalized to the resolution.
+    // given in pixels.  This maps the size of the texture to the normalized
+    // coordinates ([-1, 1], [-1, 1]).
+    ppx::float2 GetNormalizedSize(ppx::uint2 resolution) const
+    {
+        return ppx::float2(GetWidth() * 2.0f / static_cast<float>(resolution.x), GetHeight() * 2.0f / static_cast<float>(resolution.y));
```
apazylbe, line 80: Not sure if I follow the math here, are the values supposed to be in the range of [-1, 1]? 
If the width is equal to resolution wouldn't it be 2?

projects/fluid_simulation/shaders.h
```diff
@@ -202,8 +205,11 @@ class ComputeShader : public Shader
     // dr    Dispatch record to use.
     void Dispatch(const PerFrame& frame, const std::unique_ptr<ComputeDispatchRecord>& dr);
 
+    ComputeResources* GetResources() const { return mResources; }
```
apazylbe, line 208: const ComputeResources?

projects/fluid_simulation/shaders.h
```diff
@@ -618,11 +624,15 @@ class GraphicsShader : public Shader
     // coord    Normalized coordinate where to draw the texture.
     std::unique_ptr<GraphicsDispatchRecord> GetDR(Texture* image, ppx::float2 coord)
     {
-        return std::make_unique<GraphicsDispatchRecord>(this, image, coord);
+        return std::make_unique<GraphicsDispatchRecord>(this, image, coord, mResolution);
     }
 
+    GraphicsResources* GetResources() const { return mResources; }
```
apazylbe, line 630: const GraphicsResources?

projects/fluid_simulation/shaders.h
dnovillo: Yes.  Resolution is 2, since it spans from -1 to +1.

dnovillo:
Apologies for the additional patch pushed into this PR.  I cherry-picked into the wrong branch.

Keenuts:
> Apologies for the additional patch pushed into this PR. I cherry-picked into the wrong branch.

Is this a new version ready for review then?

dnovillo:
> > Apologies for the additional patch pushed into this PR. I cherry-picked into the wrong branch.
> 
> Is this a new version ready for review then?

Not yet.  I'll send an update to this today.

projects/fluid_simulation/shaders.h
dnovillo: Done.

dnovillo:
> > > Apologies for the additional patch pushed into this PR. I cherry-picked into the wrong branch.
> > 
> > 
> > Is this a new version ready for review then?
> 
> Not yet. I'll send an update to this today.

It's now ready.  Thanks.

apazylbe (approved):

projects/fluid_simulation/shaders.h
Keenuts: Got confused too, definitely thought a -1 was missing since the comments mentions the interval `[-1;1]`.
Maybe the comment should be rephrased?

dnovillo: Sure. I'll update the comment.  Thanks.

PR #326, opened 2023-10-31
Add and modify shaders used for fullscreen quads

angela28chen:
General changes:
* Adding knob `--fullscreen-quads-type` to toggle between shader type (noise / solid color / texture)
* Refactoring shaders related to quads
* Updated UI visible/invisible logic for knobs related to quads
* Instead of rebuilding pipelines for fullscreen quads during Render(), all 3 are built during setup and just switched between

Solid color shader
* Now visibly different hues for consecutive quads
* Reduced options: `Red, Blue, Green, White`

Texture loading shader
* Same texture loaded for every quad, 1:1 pixel:texel
* Added new large texture, texture loading shader

Minor changes
* Moved skybox texture to the benchmarks asset folder


angela28chen:
Note: I added a separate knob for the type since there will be a texture option in the future

assets/benchmarks/shaders/Benchmark_SolidColor.hlsl
```diff
@@ -22,18 +25,20 @@ struct ColorParams
...
 
-struct VSOutput {
-    float4 position : SV_POSITION;
-};
-
-VSOutput vsmain(float4 Position : POSITION)
+// returns zigzag value between (0.5 ~ 1.0) in steps of 0.1 between consecutive seeds
+// this is chosen to try and reduce flashing effect when the slider is dragged slowly
+// but so that color difference is still visible to the eye between consecutive quads
+float intensity(uint32_t seed)
```
wangra-google, line 31: can this be done on the cpu side instead of in the shader?

angela28chen: Sure, moved this part out from shader

wangra-google (approved):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -444,42 +471,55 @@ void GraphicsBenchmarkApp::SetupSpheresPipelines()
...
+        grfx::GraphicsPipelineCreateInfo2 gpCreateInfo  = {};
+        gpCreateInfo.VS                                 = {mVSQuads.Get(), "vsmain"};
+        gpCreateInfo.PS                                 = {mQuadsPs[i].Get(), "psmain"};
+        gpCreateInfo.vertexInputState.bindingCount      = 1;
+        gpCreateInfo.vertexInputState.bindings[0]       = mFullscreenQuads.vertexBinding;
+        gpCreateInfo.topology                           = grfx::PRIMITIVE_TOPOLOGY_TRIANGLE_STRIP;
+        gpCreateInfo.polygonMode                        = grfx::POLYGON_MODE_FILL;
+        gpCreateInfo.cullMode                           = grfx::CULL_MODE_BACK;
+        gpCreateInfo.frontFace                          = grfx::FRONT_FACE_CW;
+        gpCreateInfo.depthReadEnable                    = true;
```
wangra-google, line 514: depth read should be false

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -444,42 +471,55 @@ void GraphicsBenchmarkApp::SetupSpheresPipelines()
...
+        gpCreateInfo.depthWriteEnable                   = false;
+        gpCreateInfo.blendModes[0]                      = grfx::BLEND_MODE_NONE;
+        gpCreateInfo.outputState.renderTargetCount      = 1;
+        gpCreateInfo.outputState.renderTargetFormats[0] = GetSwapchain()->GetColorFormat();
+        gpCreateInfo.outputState.depthStencilFormat     = GetSwapchain()->GetDepthFormat();
```
wangra-google, line 519: in this case we probably dont need the depth attachment

assets/benchmarks/shaders/Benchmark_Texture.hlsl
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "VsOutput.hlsli"
+
+Texture2D Tex0 : register(t0);
+
+float4 psmain(VSOutputPos input) : SV_TARGET
+{
+    return Tex0.Load(uint3(input.position.x, input.position.y, /* mipmap */ 0));
```
wangra-google, line 21: sorry i couldn't find the vs Benchmark_VsSimpleQuads.vs, but how do you calculate the texture coord?

angela28chen: I did not add texture coord to the vertex buffer, I am just using the position data here as coordinates for which pixel of the texture to load. I tested this manually with the `--resolution NxM` flag for different resolutions and it seems to be behaving as expected?

angela28chen:
Note: I am not sure about if there is still a validation error on Linux relating to push constant size now that there are three separate pipelines, please let me know if that shows up.

assets/benchmarks/shaders/Benchmark_Texture.hlsl
angela28chen: `Benchmark_VsSimpleQuads.vs` file is added in this PR, just not with the most recent commit

angela28chen:
I thought it made more sense to include the texture shader implementation in this PR as well so I merged the two.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
angela28chen: If I remove depth stuff I actually get different behaviour for the single renderpass setting.

There are a bunch of draw calls of the same plane at the same depth. Without depth testing there is no order to which shows up on top and I've noticed that the color stays the same. Probably the first quad drawn always shows at the front, which I assume isn't the behaviour we want for these quads.

assets/benchmarks/shaders/Benchmark_Texture.hlsl
wangra-google: with the current position, the part covers the full screen will have a range of [-1, 1] whereas if we want to have 1to1 mapping between the texel and pixel, the range should be [0,1]?

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
wangra-google: hmm, that is weird. without depth testing, the order should be following the drawing order, so the last drawcall should be shown. you can verify this with renderdoc capture to see if this is the case.

wangra-google:
> I thought it made more sense to include the texture shader implementation in this PR as well so I merged the two.

make sense, thanks!

assets/benchmarks/shaders/Benchmark_Texture.hlsl
angela28chen: I see your point but I think that something else must be going on since this code produces a result that looks correct and doesn't look like I am accessing the texture out of bounds. I'm not an expert but I think it might have something to do with the HLSL POSITION->SV_POSITION change that the vertex shader is doing?

angela28chen: As in I think `input.position` here is in range `0,0` to `window dimensions`, and it's going to access texture pixel with pixel coordinates (range `0,0` to texture size `3840-1, 2160-1`)

wangra-google: if you could share a renderdoc capture, i can take a look

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
angela28chen: I tried this and I looked at pixel history, it is only touched by the draw call for the first quad.

wangra-google: the multipass one is wrong, we should not do clear when we begin a new pass, should be also `ATTACHMENT_LOAD_OP_DONT_CARE` for each renderpass, and we don't need the depth attachment

assets/benchmarks/shaders/Benchmark_Texture.hlsl
wangra-google: I see, because we use SV_POSITION as pixel input which gives us the pixel location in the range of [0-screen width/height], and we use .Load which takes the pixel location instead of the texture coord to load the image. it works well. thanks!

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
angela28chen: OK, after our discussion I played around with this a little more and figured out that disabling depth read is fine but changing this depth stencil format will cause weird behaviour on my machine. So I am thinking of removing the depth read but leaving the format alone for now. This is how the other BigWheels benchmarks related to texture loading seem to be handling it (`texture_load` and `texture_sample`)

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
angela28chen: Fixed

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
wangra-google: yeah, i checked the code, with the bigwheels framework, we can only disable the depth attachment for the swapchain which will also disable the depth for the sphere pass which is not we want. So for now it is fine by just setting the depth read to false.

wangra-google (approved):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -778,15 +834,36 @@ void GraphicsBenchmarkApp::RecordCommandBufferSpheres(PerFrame& frame)
...
-        // Solid color
-        float3 colorValues = kFullscreenQuadsColorsValues[pFullscreenQuadsColor->GetIndex()];
-        frame.cmd->PushGraphicsConstants(mFullscreenQuads.pipelineInterface, 3, &colorValues);
-    }
-    else {
-        // Noise
-        uint32_t noiseQuadRandomSeed = (uint32_t)seed;
-        frame.cmd->PushGraphicsConstants(mFullscreenQuads.pipelineInterface, 1, &noiseQuadRandomSeed);
+    switch (pFullscreenQuadsType->GetIndex()) {
+        case 0: {
```
apazylbe, line 838: nit: Consider making this an enum? There are a few places where an index is mentioned and we need to remember which index corresponds to which type

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
```diff
@@ -64,20 +64,21 @@ static constexpr std::array<const char*, 3> kAvailableLODs = {
 
 static constexpr uint32_t kMeshCount = kAvailableVbFormats.size() * kAvailableVertexAttrLayouts.size() * kAvailableLODs.size();
 
-static constexpr std::array<const char*, 6> kFullscreenQuadsColors = {
+static constexpr std::array<const char*, 3> kFullscreenQuadsType = {
```
apazylbe, line 67: nit: kFullScreenQuadsTypes?

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
```diff
@@ -152,33 +152,34 @@ class GraphicsBenchmarkApp
...
+    Texture                                                             mNormalMapTexture;
+    Texture                                                             mMetalRoughnessTexture;
+    Texture                                                             mQuadsTexture;
+    Entity                                                              mSkyBox;
+    Entity                                                              mSphere;
+    Entity2D                                                            mFullscreenQuads;
+    bool                                                                mEnableMouseMovement = true;
+    std::vector<grfx::BufferPtr>                                        mDrawCallUniformBuffers;
+    std::array<grfx::GraphicsPipelinePtr, kPipelineCount>               mPipelines;
+    std::array<grfx::GraphicsPipelinePtr, kFullscreenQuadsType.size()>  mQuadsPipelines;
```
apazylbe, line 174: nit: this is a long block that's a bit hard to read. Perhaps we should split those into logical groups and add a comment? 
For ex:
```
// Full screen quads resources
std::array<grfx::GraphicsPipelinePtr, kFullscreenQuadsType.size()>  mQuadsPipelines;
...
```

apazylbe (approved):
A few comments, otherwise lgtm

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
angela28chen: OK, split by skybox/sphere/quads.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
angela28chen: Fixed

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
angela28chen: OK, I added a enum class

angela28chen:
Should be ready for merge now

PR #330, opened 2023-11-02
Add checkbox for replacing sphere textures with 1x1 white texture

angela28chen:
Added
* Checkbox `pAllTexturesToWhite`
* Functionality and visibility logic for new knob

Also:
* Refactor graphics benchmark to make use of `grfx::Texture` and remove struct `Texture`
* Minor fix for using `FullscreenQuadsType` enum class in visibility logic


angela28chen:
I believe this only affects the spheres when the pixel shader `Benchmark_PsMemBound` is selected so I've nested the new knob under that in the UI.

angela28chen:
Note: I don't notice the performance impact of this knob at all, don't know if that is an issue?

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
```diff
@@ -208,6 +216,7 @@ class GraphicsBenchmarkApp
     std::shared_ptr<KnobCheckbox>              pFullscreenQuadsSingleRenderpass;
     std::shared_ptr<KnobCheckbox>              pAlphaBlend;
     std::shared_ptr<KnobCheckbox>              pDepthTestWrite;
+    std::shared_ptr<KnobCheckbox>              pAllTexturesToWhite;
```
wangra-google, line 219: probably better to rename to pAllTexuresTo1x1

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -33,6 +33,11 @@ void GraphicsBenchmarkApp::InitKnobs()
     pKnobPs->SetDisplayName("Pixel Shader");
     pKnobPs->SetFlagDescription("Select the pixel shader for the graphics pipeline.");
 
+    pAllTexturesToWhite = GetKnobManager().CreateKnob<ppx::KnobCheckbox>("all-textures-to-white", false);
```
wangra-google, line 36: all-textures-to-1x1

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -33,6 +33,11 @@ void GraphicsBenchmarkApp::InitKnobs()
...
+    pAllTexturesToWhite->SetDisplayName("All Textures To White");
```
wangra-google, line 37: All Textures To 1x1

wangra-google:
> Note: I don't notice the performance impact of this knob at all, don't know if that is an issue?
it only affect the perf when the shader is actually mem fetch bound. don't worry about this for now, we can do more tests once this PR is submitted.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
angela28chen: OK, done

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
angela28chen: Done

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
angela28chen: Done & also changed the flag

wangra-google (approved):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -260,6 +265,22 @@ void GraphicsBenchmarkApp::SetupSphereResources()
         samplerCreateInfo.maxLod                  = FLT_MAX;
         PPX_CHECKED_CALL(GetDevice()->CreateSampler(&samplerCreateInfo, &mMetalRoughnessTexture.sampler));
     }
+    {
+        // 1x1 White Texture
+        grfx_util::ImageOptions options = grfx_util::ImageOptions().MipLevelCount(0);
```
apazylbe, line 270: I think you can also use `PPX_CHECKED_CALL(grfx_util::CreateTexture1x1<uint8_t>(GetDevice()->GetGraphicsQueue(), {255, 255, 255, 255}, &m1x1WhiteTexture));`
Example in 16_gbuffer 
https://github.com/search?q=repo%3Agoogle%2Fbigwheels%20m1x1WhiteTexture&type=code

angela28chen: Thanks, I didn't know this existed. Changed

angela28chen:
I did some refactoring of all the textures used in the graphics benchmark to make use of bigwheels `grfx::Texture` and remove the `Texture` struct defined in the benchmark. Also decoupled samplers from textures.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -120,6 +125,24 @@ void GraphicsBenchmarkApp::Setup()
...
+        samplerCreateInfo.minLod                  = 0;
+        samplerCreateInfo.maxLod                  = FLT_MAX;
+        PPX_CHECKED_CALL(GetDevice()->CreateSampler(&samplerCreateInfo, &mLinearSampler));
+
+        samplerCreateInfo            = {};
+        samplerCreateInfo.magFilter  = grfx::FILTER_NEAREST;
+        samplerCreateInfo.minFilter  = grfx::FILTER_NEAREST;
+        samplerCreateInfo.mipmapMode = grfx::SAMPLER_MIPMAP_MODE_NEAREST;
+        samplerCreateInfo.minLod     = 0.0f;
+        samplerCreateInfo.maxLod     = 1.0f;
```
apazylbe, line 143: I am curious, is there a reason maxLod is 1 here and FLT_MAX for linear?

apazylbe (approved):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
wangra-google: in the case of 1x1 texture, this value doesn't really matter since there is only 1 mip level. But actually we could just use the linear sampler directly for the 1x1 case. No need to switch the sampler in the 1x1 case

angela28chen: I see, OK. I've removed `mNearestSampler`

angela28chen:
Ready to be merged

wangra-google (approved):

PR #333, opened 2023-11-06
Update the asset file path for Android Benchmark app setup

elviscapiaq:


benchmarks/graphics_pipeline/build.gradle
```diff
@@ -1,5 +1,6 @@
 ArrayList projectAssets = [
     "benchmarks/shaders/spv",
+    "benchmarks/textures",
```
wangra-google, line 3: there are some textures under the `Wood` folder, are they used by benchmarks? if so, should we add the subfolder? if not, shall we remove those textures? https://github.com/google/bigwheels/tree/main/assets/benchmarks/textures/Wood

angela28chen: I think these are used by the benchmark `texture_transfer_cpu_to_gpu`

wangra-google: good to know, thanks

elviscapiaq: I believe when you add a folder. all files inside it and its subfolders are also added, so we shouldn't need to add the `Wood` subfolder.

elviscapiaq:
I tested this last update on the Benchmark app and it is working as expected now.

wangra-google (approved):

wangra-google:
> I tested this last update on the Benchmark app and it is working as expected now.

LGTM

angela28chen (approved):

apazylbe (approved):

elviscapiaq:
Closing this PR since I created it over my local main branch and had several issues to rebase it.
I created a new PR with this same change: https://github.com/google/bigwheels/pull/335
I am sorry for this.

PR #334, opened 2023-11-08
Add XR support to benchmark

wangra-google:
Need to generate project with 
```
cmake -B build -G "Visual Studio 16 2019" -A x64 -DPPX_BUILD_XR=1
```

Then change `settings.xr.enable` to `true` to enable XR mode 

The benchmark xr version is done within the original project (instead of duplicating the project and have a `xr` version like fishtornado and cube) because the benchmark itself is WIP, and this could avoid duplicating the changes in 2 projects. 

The downside is that there is a lot of `#if defined(PPX_BUILD_XR)` in the .h/.cpp files (this was the main reason why the cube_xr and fishtornado_xr were in separate projects)

I will do another pass in a separate PR to move all #if defined(PPX_BUILD_XR) into the base class or framework so that we don't do this ifdefs in the application.

### DIFF
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
@@ -657,7 +657,8 @@ void GraphicsBenchmarkApp::Render()
     grfx::SwapchainPtr swapchain  = GetSwapchain(currentViewIndex);
 
 #if defined(PPX_BUILD_XR)
-    if (swapchain->ShouldSkipExternalSynchronization()) {
+    if (IsXrEnabled()) {
+        PPX_ASSERT_MSG(swapchain->ShouldSkipExternalSynchronization(), "XRComponent should not be nullptr when XR is enabled!");
         // No need to
         // - Signal imageAcquiredSemaphore & imageAcquiredFence.
         // - Wait for imageAcquiredFence since xrWaitSwapchainImage is called in AcquireNextImage.
@@ -904,8 +905,8 @@ void GraphicsBenchmarkApp::RecordCommandBufferSkybox(PerFrame& frame)
     frame.cmd->BindIndexBuffer(mSkyBox.mesh);
     frame.cmd->BindVertexBuffers(mSkyBox.mesh);
     {
-        SkyboxData data{};
-        data.MVP = frame.sceneData.viewProjectionMatrix * glm::scale(float3(500.0f, 500.0f, 500.0f));
+        SkyboxData data = {};
+        data.MVP        = frame.sceneData.viewProjectionMatrix * glm::scale(float3(500.0f, 500.0f, 500.0f));
         mSkyBox.uniformBuffer->CopyFromSource(sizeof(data), &data);
 
         frame.cmd->PushGraphicsUniformBuffer(mSkyBox.pipelineInterface, /* binding = */ 0, /* set = */ 0, /* bufferOffset = */ 0, mSkyBox.uniformBuffer);
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
@@ -99,6 +99,12 @@ void GraphicsBenchmarkApp::Config(ppx::ApplicationSettings& settings)
     settings.grfx.api                   = kApi;
     settings.grfx.enableDebug           = false;
     settings.grfx.swapchain.depthFormat = grfx::FORMAT_D32_FLOAT;
+#if defined(PPX_BUILD_XR)
+    // XR specific settings
+    settings.grfx.pacedFrameRate   = 0;
+    settings.xr.enable             = false; // Change this to true to enable the XR mode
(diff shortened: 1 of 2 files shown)


benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -764,8 +904,8 @@ void GraphicsBenchmarkApp::RecordCommandBufferSkybox(PerFrame& frame)
     frame.cmd->BindIndexBuffer(mSkyBox.mesh);
     frame.cmd->BindVertexBuffers(mSkyBox.mesh);
     {
-        SkyboxData data = {};
-        data.MVP        = mCamera.GetViewProjectionMatrix() * glm::scale(float3(500.0f, 500.0f, 500.0f));
+        SkyboxData data{};
```
angela28chen, line 907: Nit: Is there a reason this was changed from `SkyboxData data = {}` ? I thought that one was recommended syntax for initializing structures. go/totw/88#best-practices-for-initialization

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -586,14 +600,79 @@ void GraphicsBenchmarkApp::ProcessQuadsKnobs()
...
+    submitInfo.signalSemaphoreCount = 0;
+    submitInfo.ppSignalSemaphores   = nullptr;
+
+    submitInfo.pFence = frame.uiRenderCompleteFence;
+
+    PPX_CHECKED_CALL(GetGraphicsQueue()->Submit(&submitInfo));
+}
+#endif
+
 void GraphicsBenchmarkApp::Render()
```
angela28chen, line 638: This was a bit hard to read with 5 blocks of `#if defined(PPX_BUILD_XR)`, I wonder if it would just be better if there was a separate `RenderXR` function for the case of (`PPX_BUILD_XR==1` and `IsXrEnabled()==true`) ?

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -586,14 +600,79 @@ void GraphicsBenchmarkApp::ProcessQuadsKnobs()
...
+#endif
+
     uint32_t           imageIndex = UINT32_MAX;
-    PPX_CHECKED_CALL(swapchain->AcquireNextImage(UINT64_MAX, frame.imageAcquiredSemaphore, frame.imageAcquiredFence, &imageIndex));
-    // Wait for and reset image acquired fence
-    PPX_CHECKED_CALL(frame.imageAcquiredFence->WaitAndReset());
+    grfx::SwapchainPtr swapchain  = GetSwapchain(currentViewIndex);
+
+#if defined(PPX_BUILD_XR)
+    if (swapchain->ShouldSkipExternalSynchronization()) {
```
angela28chen, line 660: Should this also have `IsXrEnabled()` ?

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
wangra-google: oops, changed by mistake, reverted

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
wangra-google: fixed

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
wangra-google: i would prefer to avoid duplicating the code so that in the future when people changes the Render() function, they don't need to do the change in 2 places. Although I do agree that the `#if defined(PPX_BUILD_XR)` is annoying. I am planning to do another pass to move all `#if defined(PPX_BUILD_XR)` into the base class or framework so that we don't do this ifdefs in the application.

angela28chen (approved):
LGTM

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -99,6 +99,12 @@ void GraphicsBenchmarkApp::Config(ppx::ApplicationSettings& settings)
     settings.grfx.api                   = kApi;
     settings.grfx.enableDebug           = false;
     settings.grfx.swapchain.depthFormat = grfx::FORMAT_D32_FLOAT;
+#if defined(PPX_BUILD_XR)
+    // XR specific settings
+    settings.grfx.pacedFrameRate   = 0;
+    settings.xr.enable             = false; // Change this to true to enable the XR mode
```
apazylbe, line 105: I am curious on how this will be handled in the future, will you add a flag to enable this? So that you don't have to modify the code to run in XR mode

apazylbe (approved):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
wangra-google: yeah, i think that could be an option.

PR #337, opened 2023-11-09
Fix the xr version of benchmark on android

wangra-google:


apazylbe (approved):
lgtm as a temporary fix until we address this properly

PR #338, opened 2023-11-10
Benchmark: Shows min, max and average submission time on CPU

elviscapiaq:


wangra-google (approved):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -831,28 +836,35 @@ void GraphicsBenchmarkApp::Render()
...
 
-    grfx::SubmitInfo submitInfo   = {};
-    submitInfo.commandBufferCount = 1;
-    submitInfo.ppCommandBuffers   = &frame.cmd;
+    size_t submissionCount = pSubmissionCount->GetValue();
+    mSubmissionTimes.resize(submissionCount);
+    for (size_t i = 0; i < submissionCount; i++) {
+        grfx::SubmitInfo submitInfo   = {};
+        submitInfo.commandBufferCount = 1;
+        submitInfo.ppCommandBuffers   = &frame.cmd;
```
Keenuts, line 844: This is not legal in Vulkan, and triggerd VUID-vkQueueSubmit-pCommandBuffers-00071.
When you call submit, the command buffer goes from executable to pending until its execution finished (https://docs.vulkan.org/spec/latest/chapters/cmdbuffers.html#:~:text=Pending,the%20commands%20recorded%20to%20it.)
This means the next loop will call submit with a buffer having the "pending" state, which is incorrect.

Seems like adding the VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT bit is enough to solve this. I'd be tempted to add this bit by default on all primary command buffers we create with BigWheels, but unsure if this has any drawbacks. Probably perf-wise since the driver cannot assume the command-buffer remains unchanged while recording?
(@chaoticbob might know)

wangra-google: Thanks for mentioning this. it is indeed a problem, and we want to avoid using VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT because there could be some extra memory allocations at driver level. I had some discussions with Shan, and will change the design of this knob a bit.

elviscapiaq:
Thanks for all your feedback. I updated this PR such that now we are only showing the min, max and average submission time for the CPU avoiding the drawbacks of using VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT . Also, I rebased it.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -854,7 +854,21 @@ void GraphicsBenchmarkApp::Render()
     }
     submitInfo.pFence = frame.renderCompleteFence;
 
+    Timer timerSubmit;
+    timerSubmit.Start();
     PPX_CHECKED_CALL(GetGraphicsQueue()->Submit(&submitInfo));
+    double t = timerSubmit.MillisSinceStart();
+
+    uint64_t frameCount     = GetFrameCount();
+    mSubmissionTime.average = (mSubmissionTime.average * frameCount) / (frameCount + 1) + t / (frameCount + 1);
```
apazylbe, line 863: Should we use the existing metric framework for this: https://github.com/google/bigwheels/blob/main/src/ppx/application.cpp#L1747 ? 
That way we can also export this metric

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -906,6 +920,24 @@ void GraphicsBenchmarkApp::DrawExtraInfo()
     uint64_t frequency = 0;
     GetGraphicsQueue()->GetTimestampFrequency(&frequency);
 
+    ImGui::Columns(2);
```
wangra-google, line 923: `ImGui::Columns(2);` is only required once at the beginning, the other ones are not needed

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -854,7 +854,21 @@ void GraphicsBenchmarkApp::Render()
     }
     submitInfo.pFence = frame.renderCompleteFence;
 
+    Timer timerSubmit;
+    timerSubmit.Start();
     PPX_CHECKED_CALL(GetGraphicsQueue()->Submit(&submitInfo));
+    double t = timerSubmit.MillisSinceStart();
+
+    uint64_t frameCount     = GetFrameCount();
```
wangra-google, line 862: mSubmissionTime.average = (mSubmissionTime.average * frameCount + t) / (frameCount + 1);

apazylbe (commented):
Also don't forget to update the PR description

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
angela28chen: I feel like there's other metrics in this benchmark that would be good to use the metrics framework for. Maybe we should just do some refactoring in a separate PR after this to address them?

angela28chen (approved):
LGTM (Though I'm guessing the title of this PR needs to be updated since it's no longer about a submission count knob?)

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
wangra-google: yeah, separate PR would be better.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
elviscapiaq: Done.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
elviscapiaq: Done.

elviscapiaq:
I updated the PR based on the feedback and also rebased it.

wangra-google:
> I updated the PR based on the feedback and also rebased it.

LGTM

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
apazylbe: sounds good

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -860,7 +860,21 @@ void GraphicsBenchmarkApp::Render()
     }
     submitInfo.pFence = frame.renderCompleteFence;
 
+    Timer timerSubmit;
```
apazylbe, line 863: Could you add an include for this?

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -860,7 +860,21 @@ void GraphicsBenchmarkApp::Render()
...
+    timerSubmit.Start();
```
apazylbe, line 864: Start() does seem to return a result, perhaps worth asserting it is a success?

Keenuts (approved):
LGTM. Nothing to add outside of what's already said.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
elviscapiaq: Sound good to me. Updated!

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
elviscapiaq: Thanks for the reminder. Done.

elviscapiaq:
I've incorporated the feedback and performed a rebase.

apazylbe (approved):

PR #342, opened 2023-11-14
Knob: type deduction for create/init.

hysw:
Knob types are defined in header already,
let compiler deduce the type from knob variable
when initializing them.

PR #344, opened 2023-11-14
Knob: non-string dropdown.

hysw:


hysw:
This change stack on top of #342

angela28chen:
I am in favour of this change, I think it's useful to organize the name-values for entries together rather than in separate halves. It's more readable and would probably reduce error in future for if you only adjusted one half but forgot the other.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -670,7 +666,7 @@ void GraphicsBenchmarkApp::ProcessKnobs()
     const bool enableSpheres = pEnableSpheres->GetValue();
     // If the LOD is empty, assuming we skipped the setup for all sphere resources at start
     // So need to do the intial setup for all resources here
-    const bool spheresAreSetUp               = !mSphereLODs.empty();
+    const bool spheresAreSetUp               = mSpheresAreSetUp;
```
angela28chen, line 669: What about just removing `spheresAreSetUp` and using `mSpheresAreSetUp` in the lines below.

angela28chen (approved):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
hysw: The logic here need to be reworked, not just renaming, i.e. `SetupSpheresPipelines` still throw away a PipelineInterface without freeing it, I'm trying to not change much code so I trace back source PR to figure out what's the intension of existing code.

angela28chen: I see

include/ppx/knob.h
```diff
@@ -275,7 +303,12 @@ class KnobDropdown final
     // Needed for setting from flags but use is discouraged otherwise
     void SetIndex(const std::string& newValue)
```
apazylbe, line 304: Should this be `SetIndexByValue` and `SetIndexByName` to provide flexibility? Although if there is no use for that, we can always add that later.

apazylbe (approved):

include/ppx/knob.h
hysw: Nothing uses it, so deleted instead, we can add `SetIndexBy{Name,Value}` when we needs them.

elviscapiaq (approved):
I like this change which would help to organize pairs of name-values. Also, it can reduce errors in the future.

wangra-google (approved):

PR #347, opened 2023-11-14
A few changes to XR handling

apazylbe:
Swapchain:
- Add Wait() method to swapchain, it is a no-op for non-XR. For XR it calls xrWaitSwapchainImage. Since the command buffer can be recorded between xrAcquireSwapchainImage and xrWaitSwapchainImage, putting xrWaitSwapchainImage in AcquireNextImage is unnecessarily restrictive.
- Add xrReleaseSwapchainImage to Present(). It is weird not to have a function for this and to do this in the main application loop, since the application already calls Acquire.

Remove calling Render() twice for XR and the viewIndex logic. It is very much application/project specific logic of how it decides to render two views, and can be easily handled in the application level. It also becomes more restrictive where application decides to render differently (for example, using a single command buffer for UI and both views, or with multiview

### DIFF
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
@@ -755,6 +755,7 @@ void GraphicsBenchmarkApp::RecordAndSubmitCommandBufferGUIXR(PerFrame& frame)
         frame.uiCmd->EndRenderPass();
     }
     PPX_CHECKED_CALL(frame.uiCmd->End());
+    uiSwapchain->Wait(imageIndex);
 
     grfx::SubmitInfo submitInfo     = {};
     submitInfo.commandBufferCount   = 1;
@@ -767,29 +768,42 @@ void GraphicsBenchmarkApp::RecordAndSubmitCommandBufferGUIXR(PerFrame& frame)
     submitInfo.pFence = frame.uiRenderCompleteFence;
 
     PPX_CHECKED_CALL(GetGraphicsQueue()->Submit(&submitInfo));
+    uiSwapchain->Present(imageIndex, 0, nullptr);
 }
 #endif
+
+void GraphicsBenchmarkApp::DispatchRender()
+{
+    if (IsXrEnabled()) {
+        size_t viewCount = static_cast<uint32_t>(GetXrComponent().GetViewCount());
+        for (size_t view = 0; view < viewCount; view++) {
+            mViewIndex = view;
+            Render();
+        }
+    }
+    else {
+        Render();
+    }
+}
 
 void GraphicsBenchmarkApp::Render()
 {
     ProcessInput();
     ProcessKnobs();
 
-    PerFrame& frame            = mPerFrame[0];
-    uint32_t  currentViewIndex = 0;
+    PerFrame& frame = mPerFrame[0];
 
 #if defined(PPX_BUILD_XR)
     // Render UI into a different composition layer.
     if (IsXrEnabled()) {
-        currentViewIndex = GetXrComponent().GetCurrentViewIndex();
-        if ((currentViewIndex == 0) && GetSettings()->enableImGui) {
+        if ((mViewIndex == 0) && GetSettings()->enableImGui) {
             RecordAndSubmitCommandBufferGUIXR(frame);
         }
     }
 #endif
 
     uint32_t           imageIndex = UINT32_MAX;
-    grfx::SwapchainPtr swapchain  = GetSwapchain(currentViewIndex);
+    grfx::SwapchainPtr swapchain  = GetSwapchain(mViewIndex);
 
 #if defined(PPX_BUILD_XR)
     if (IsXrEnabled()) {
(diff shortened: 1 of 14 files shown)


projects/04_cube_xr/CubeXr.cpp
```diff
@@ -201,17 +201,27 @@ void CubeXrApp::Setup()
     mScissorRect = {0, 0, GetWindowWidth(), GetWindowHeight()};
 }
 
+void CubeXrApp::DispatchRender()
```
hysw, line 204: Maybe this belongs to `XR[Aware]Application` or something.
i.e. CubeXrApp -> XRApplication -> Application

apazylbe: Perhaps, and this PR is definitely no the end result of what XR should look like, but maybe just the next step.
We still need to figure out what that proper "end" state looks like. I think if we have `CubeXrApp -> XRApplication -> Application` we would want XRApplication to support both XR and non-XR. There are also different ways XR can be handled by application (just doing what it does now with full submits per view, recording one command buffer for everything, multiview), so I think this per view logic belongs in the derived application level.
I am kinda thinking maybe we should add another class (like Renderer), where application just calls renderer->Render(), and there can be specific classes like XrRenderer, XrMultiviewRenderer, etc, then we can also use that renderer in a different application to render to a specific composition layer

projects/fishtornado_xr/FishTornado.cpp
```diff
@@ -1057,9 +1074,10 @@ void FishTornadoApp::Render()
         submitInfo.pFence = frame.uiRenderCompleteFence;
 
         PPX_CHECKED_CALL(GetGraphicsQueue()->Submit(&submitInfo));
+        PPX_CHECKED_CALL(uiSwapchain->Present(uiImageIndex, 0, nullptr));
     }
 
-    grfx::SwapchainPtr swapchain = GetSwapchain(currentViewIndex);
+    grfx::SwapchainPtr swapchain = GetSwapchain(mViewIndex);
 
     UpdateTime();
```
wangra-google, line 1083: sorry i am not able to add comment to the line directly, but the comment in line 1069 is not valid anymore with this change, so the comment needs to be be changed.

include/ppx/grfx/grfx_swapchain.h
```diff
@@ -146,6 +146,8 @@ class Swapchain
         grfx::Fence*     pFence,     // Wait fence
         uint32_t*        pImageIndex);
 
+    Result Wait(uint32_t);
```
wangra-google, line 149: the input arg uint32_t is never used, so maybe better remove it?

wangra-google: the name `Wait` is a bit confusing. User might be confused what is it waiting for, and if it is a cpu or gpu wait. maybe add some comments or rename it something like "XRCPUWaitSwapchainImage" ?

projects/04_cube_xr/CubeXr.cpp
```diff
@@ -330,36 +342,31 @@ void CubeXrApp::Render()
...
     submitInfo.pFence = frame.renderCompleteFence;
-
+    swapchain->Wait(imageIndex);
     PPX_CHECKED_CALL(GetGraphicsQueue()->Submit(&submitInfo));
 
-    // No need to present when XR is enabled.
     if (!IsXrEnabled()) {
         PPX_CHECKED_CALL(swapchain->Present(imageIndex, 1, &frame.renderCompleteSemaphore));
     }
     else {
```
wangra-google, line 358: not sure if it is a good idea, but just want to hear your opinions :): since the `Present` is not a real Present in the case of XR swapchain, and the all the args are ignored anyways, maybe it is cleaner to have just 1 unified `Present` code path here for both XR and non-XR? Because otherwise, maybe it is clearer to have another XR specific function here instead of naming it `Present` with ignored inputs?

wangra-google:
> Swapchain:
> 
> * Add Wait() method to swapchain, it is a no-op for non-XR. For XR it calls xrWaitSwapchainImage. Since the command buffer can be recorded between xrAcquireSwapchainImage and xrWaitSwapchainImage, putting xrWaitSwapchainImage in AcquireNextImage is unnecessarily restrictive.
> * Add xrReleaseSwapchainImage to Present(). It is weird not to have a function for this and to do this in the main application loop, since the application already calls Acquire.
> 
> Remove calling Render() twice for XR and the viewIndex logic. It is very much application/project specific logic of how it decides to render two views, and can be easily handled in the application level. It also becomes more restrictive where application decides to render differently (for example, using a single command buffer for UI and both views, or with multiview

Thanks for the refactoring, and overall it is much clearer this way. Added some comments.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -825,14 +839,16 @@ void GraphicsBenchmarkApp::Render()
...
-        const glm::mat4 p                    = GetXrComponent().GetProjectionMatrixForCurrentViewAndSetFrustumPlanes(PPX_CAMERA_DEFAULT_NEAR_CLIP, PPX_CAMERA_DEFAULT_FAR_CLIP);
+        const glm::mat4 v                    = GetXrComponent().GetViewMatrixForView(mViewIndex);
+        const glm::mat4 p                    = GetXrComponent().GetProjectionMatrixForViewAndSetFrustumPlanes(mViewIndex, PPX_CAMERA_DEFAULT_NEAR_CLIP, PPX_CAMERA_DEFAULT_FAR_CLIP);
         frame.sceneData.viewProjectionMatrix = p * v;
     }
 #endif
 
     RecordCommandBuffer(frame, swapchain, imageIndex);
 
+    swapchain->Wait(imageIndex);
```
Keenuts, line 850: Not sure if adding that brings real-life benefits. Does moving this into the acquire to have something closer to the non-XR variant affects performance?

Not a strong opinion, feel free to dismiss this.

apazylbe:
Thanks for the feedback everyone! I will incorporate these into separate changes since some might be more controversial/require more though. Closing this.

PR #350, opened 2023-11-15
Properly caching sphere pipelines.

hysw:
- Don't resource leak pipelines when enable/disable depth/alpha.
- Sphere pipelines are now created on demand.
- Pipeline are cleaned up during shutdown.

### DIFF
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
@@ -158,13 +158,6 @@ void GraphicsBenchmarkApp::Setup()
         mMeshesIndexer.AddDimension(kAvailableVbFormats.size());
         mMeshesIndexer.AddDimension(kAvailableVertexAttrLayouts.size());
     }
-    // Graphics pipelines indexer
-    {
-        mGraphicsPipelinesIndexer.AddDimension(kAvailableVsShaders.size());
-        mGraphicsPipelinesIndexer.AddDimension(kAvailablePsShaders.size());
-        mGraphicsPipelinesIndexer.AddDimension(kAvailableVbFormats.size());
-        mGraphicsPipelinesIndexer.AddDimension(kAvailableVertexAttrLayouts.size());
-    }
     // Sampler
     {
         grfx::SamplerCreateInfo samplerCreateInfo = {};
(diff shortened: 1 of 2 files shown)


wangra-google (approved):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
```diff
@@ -164,7 +166,51 @@ class GraphicsBenchmarkApp
...
+        }
+
+        struct Hash
+        {
+            // Not a good hash function, but good enough.
+            size_t operator()(const SpherePipelineKey& key) const
+            {
+                size_t res = 0;
+
+                res = (res << 2) | key.ps;
```
Keenuts, line 200: why shift res if it's zero.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
```diff
@@ -164,7 +166,51 @@ class GraphicsBenchmarkApp
...
+                   vs == rhs.vs &&
+                   vertexFormat == rhs.vertexFormat &&
+                   vertexAttributeLayout == rhs.vertexAttributeLayout &&
+                   enableDepth == rhs.enableDepth &&
+                   enableAlphaBlend == rhs.enableAlphaBlend;
+        }
+
+        struct Hash
+        {
+            // Not a good hash function, but good enough.
```
Keenuts, line 195: Did you meant shifting by 8 bits?
Because as is, a pipeline with ps=1, vs=2 collides with ps=0,vs=2

Keenuts (approved):
I'd be fine with this change, just the hash part that seems to be problematic.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
hysw: Keeping it consistent with other field (ps is 2 bit at the moment), compiler will probably remove the (0<<2) anyway.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
hysw: Right now, `vs = {0,1}`, so it won't collide unless something went seriously wrong, and if we add more vertex shader variant and forget to update the hash, it will create a few hash collision which won't be terrible.

hysw: I'd like to use `boost::hash_combine` or something similar instead, although I'm not sure what is the licensing requirement for code snippets. They still have collision, just not obvious.

Keenuts: >  if we add more vertex shader variant and forget to update the hash, it will create a few hash collision which won't be terrible.

Doesn't this means we'd use the wrong pipeline, hence don't use the correct vertex shader?

Keenuts: > I'd like to use `boost::hash_combine` or something similar instead, although I'm not sure what is the licensing requirement for code snippets. They still have collision, just not obvious.

cppref is CC-BY-SA so I suppose no-no, but I the boost license seems to be OK

hysw: > > if we add more vertex shader variant and forget to update the hash, it will create a few hash collision which won't be terrible.
> 
> Doesn't this means we'd use the wrong pipeline, hence don't use the correct vertex shader?

That's the job of `operator==`

Keenuts: Oh right, of course... 🤦‍♂

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
Keenuts: Sad that log2 is not constexpr, otherwise you could do `(res << std::log2(size_ps))` etc

hysw: `constexpr log2` is not that hard to write.
Although, multiplication works fine, it probably adds a fraction of nano second to the run time which is fine.

elviscapiaq: I believe it would be worthwhile to verify that the number of elements for each option does not exceed the bits allocated in the hash. For example: vs.size() <= (1 << bitsForVs),   ps.size() <= (1 << bitsForPs), so we can avoid collisions and remember to update the hash computation.
Update: I think is was already solved.

elviscapiaq (approved):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -526,37 +535,67 @@ void GraphicsBenchmarkApp::SetupSpheresPipelines()
...
+Result GraphicsBenchmarkApp::CompileSpherePipeline(const SpherePipelineKey& key)
+{
+    if (mPipelines.find(key) != mPipelines.end()) {
+        return SUCCESS;
+    }
+    grfx::GraphicsPipelinePtr pipeline    = nullptr;
+    bool                      interleaved = key.vertexAttributeLayout == 0;
+    size_t                    meshIndex   = 2 * key.vertexFormat + (interleaved ? 0 : 1);
+    grfx::BlendMode           blendMode   = (key.enableAlphaBlend ? grfx::BLEND_MODE_ALPHA : grfx::BLEND_MODE_NONE);
+    // Interleaved pipeline
```
angela28chen, line 551: nit: Was this comment supposed to be inside the if block below?

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -526,37 +535,67 @@ void GraphicsBenchmarkApp::SetupSpheresPipelines()
...
+}
+
+Result GraphicsBenchmarkApp::CompileSpherePipeline(const SpherePipelineKey& key)
+{
+    if (mPipelines.find(key) != mPipelines.end()) {
+        return SUCCESS;
+    }
+    grfx::GraphicsPipelinePtr pipeline    = nullptr;
+    bool                      interleaved = key.vertexAttributeLayout == 0;
+    size_t                    meshIndex   = 2 * key.vertexFormat + (interleaved ? 0 : 1);
```
angela28chen, line 549: nit: maybe the `2` here can be `kAvailableVertexAttrLayouts.size()`

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -526,37 +535,67 @@ void GraphicsBenchmarkApp::SetupSpheresPipelines()
...
+    GetSpherePipeline();
+}
+
+Result GraphicsBenchmarkApp::CompileSpherePipeline(const SpherePipelineKey& key)
+{
+    if (mPipelines.find(key) != mPipelines.end()) {
+        return SUCCESS;
+    }
+    grfx::GraphicsPipelinePtr pipeline    = nullptr;
+    bool                      interleaved = key.vertexAttributeLayout == 0;
```
apazylbe, line 548: nit: put () around `key.vertexAttributeLayout == 0`

angela28chen (approved):
Thanks for fixing this. Just to make sure I understand the reasoning, is the reason for explicitly destroying the vulkan objects only so that the application does not crash on shutdown?

hysw:
> Thanks for fixing this. Just to make sure I understand the reasoning, is the reason for explicitly destroying the vulkan objects only so that the application does not crash on shutdown?

BigWheels automatically clean up the resource on shutdown so it is fine.
The problem is I can click Alpha Blend checkbox 50 times, and it creates 1200(24*50) pipelines objects.

apazylbe (approved):

angela28chen:
> BigWheels automatically clean up the resource on shutdown so it is fine.
The problem is I can click Alpha Blend checkbox 50 times, and it creates 1200(24*50) pipelines objects.

I think I understand the caching part fixing the leak for alpha blend / depth test rebuilding, I'm asking specifically about all the pipeline destroying added in `Shutdown()`?

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
hysw: Done.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
hysw: Done.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
hysw: Done.

hysw:
> I think I understand the caching part fixing the leak for alpha blend / depth test rebuilding, I'm asking specifically about all the pipeline destroying added in `Shutdown()`?

Removed `Shutdown()` part.

PR #352, opened 2023-11-20
Benchmark: Initialize spheres on demand 

elviscapiaq:
- It considerably reduces the startup time and memory by initializing only 100 spheres.

### DIFF
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
@@ -910,8 +910,7 @@ void GraphicsBenchmarkApp::ProcessKnobs()
 
     if (enableSpheres) {
         // Update sphere resources and mesh
-        const bool spheresAreSetUp = mSpheresAreSetUp;
-        if (!spheresAreSetUp) {
+        if (!mSpheresAreSetUp) {
             // This creates all resources.
             SetupSpheres();
         }
@@ -919,10 +918,6 @@ void GraphicsBenchmarkApp::ProcessKnobs()
             const bool requireMoreSpheres = sphereInstanceCountKnobChanged && (pSphereInstanceCount->GetValue() > mInitializedSpheres);
             if (requireMoreSpheres) {
                 SetupSphereMeshes();
-            }
-            // Rebuild pipelines
-            if (alphaBlendKnobChanged || depthTestWriteKnobChanged) {
-                SetupSpheresPipelines();
             }
             // Update descriptors
             if (allTexturesTo1x1KnobChanged) {
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
@@ -85,12 +85,12 @@ void GraphicsBenchmarkApp::InitKnobs()
     pKnobVertexAttrLayout->SetFlagDescription("Select the Vertex Attribute Layout for the graphics pipeline.");
     pKnobVertexAttrLayout->SetIndent(1);
 
-    GetKnobManager().InitKnob(&pSphereInstanceCount, "sphere-count", /* defaultValue = */ 50, /* minValue = */ 1, kMaxSphereInstanceCount);
+    GetKnobManager().InitKnob(&pSphereInstanceCount, "sphere-count", /* defaultValue = */ kDefaultSphereInstanceCount, /* minValue = */ 1, kMaxSphereInstanceCount);
     pSphereInstanceCount->SetDisplayName("Sphere Count");
     pSphereInstanceCount->SetFlagDescription("Select the number of spheres to draw on the screen.");
     pSphereInstanceCount->SetIndent(1);
 
-    GetKnobManager().InitKnob(&pDrawCallCount, "drawcall-count", /* defaultValue = */ kDefaultSphereInstanceCount, /* minValue = */ 1, kMaxSphereInstanceCount);
+    GetKnobManager().InitKnob(&pDrawCallCount, "drawcall-count", /* defaultValue = */ 1, /* minValue = */ 1, kMaxSphereInstanceCount);
     pDrawCallCount->SetDisplayName("DrawCall Count");
     pDrawCallCount->SetFlagDescription("Select the number of draw calls to be used to draw the `--sphere-count` spheres.");
     pDrawCallCount->SetIndent(1);
@@ -231,9 +231,7 @@ void GraphicsBenchmarkApp::Setup()
     SetupSkyBoxPipelines();
 
     if (pEnableSpheres->GetValue()) {
-        SetupSphereResources();
-        SetupSphereMeshes();
-        SetupSpheresPipelines();
+        SetupSpheres();
     }
 
     // =====================================================================
@@ -537,12 +535,6 @@ void GraphicsBenchmarkApp::SetupSkyBoxMeshes()
 
 void GraphicsBenchmarkApp::SetupSphereMeshes()
 {
-    const uint32_t requiredSphereCount = std::max<uint32_t>(pSphereInstanceCount->GetValue(), kDefaultSphereInstanceCount);
-    const uint32_t initSphereCount     = std::min<uint32_t>(kMaxSphereInstanceCount, 2 * std::max(requiredSphereCount, mInitializedSpheres));
-
-    OrderedGrid grid(initSphereCount, kSeed);
-    mInitializedSpheres = initSphereCount;
-
     GetDevice()->WaitIdle();
     // Destroy the meshes if they were created.
     for (auto& mesh : mSphereMeshes) {
@@ -552,8 +544,12 @@ void GraphicsBenchmarkApp::SetupSphereMeshes()
         }
     }
 
+    const uint32_t requiredSphereCount = std::max<uint32_t>(pSphereInstanceCount->GetValue(), kDefaultSphereInstanceCount);
+    const uint32_t initSphereCount     = std::min<uint32_t>(kMaxSphereInstanceCount, 2 * std::max(requiredSphereCount, mInitializedSpheres));
+
     // Create the meshes
-    uint32_t meshIndex = 0;
+    OrderedGrid grid(initSphereCount, kSeed);
+    uint32_t    meshIndex = 0;
     for (const auto& lod : kAvailableLODs) {
         PPX_LOG_INFO("LOD: " << lod.name);
         SphereMesh sphereMesh(/* radius = */ 1, lod.value.longitudeSegments, lod.value.latitudeSegments);
@@ -564,8 +560,7 @@ void GraphicsBenchmarkApp::SetupSphereMeshes()
         PPX_CHECKED_CALL(grfx_util::CreateMeshFromGeometry(GetGraphicsQueue(), sphereMesh.GetHighPrecisionInterleaved(), &mSphereMeshes[meshIndex++]));
         PPX_CHECKED_CALL(grfx_util::CreateMeshFromGeometry(GetGraphicsQueue(), sphereMesh.GetHighPrecisionPositionPlanar(), &mSphereMeshes[meshIndex++]));
     }
-
-    mSpheresAreSetUp = true;
+    mInitializedSpheres = initSphereCount;
 }
 
 void GraphicsBenchmarkApp::SetupFullscreenQuadsMeshes()
@@ -803,6 +798,15 @@ void GraphicsBenchmarkApp::SetupFullscreenQuadsPipelines()
     GetFullscreenQuadPipeline();
 }
 
+void GraphicsBenchmarkApp::SetupSpheres()
+{
+    SetupSphereResources();
+    SetupSphereMeshes();
+    // Pipelines must be setup after meshes to use in vertex bindings.
+    SetupSpheresPipelines();
+    mSpheresAreSetUp = true;
+}
+
 void GraphicsBenchmarkApp::UpdateOffscreenBuffer(grfx::Format format, int w, int h)
 {
     {
@@ -881,22 +885,14 @@ void GraphicsBenchmarkApp::ProcessKnobs()
     const bool enableSpheresKnobChanged       = pEnableSpheres->DigestUpdate();
     const bool sphereInstanceCountKnobChanged = pSphereInstanceCount->DigestUpdate();
 
-    const bool enableSpheres = pEnableSpheres->GetValue();
-    // If the LOD is empty, assuming we skipped the setup for all sphere resources at start
-    // So need to do the intial setup for all resources here
-    const bool spheresAreSetUp         = mSpheresAreSetUp;
-    const bool requireMoreSpheres      = sphereInstanceCountKnobChanged && (pSphereInstanceCount->GetValue() > mInitializedSpheres);
-    const bool updateSphereDescriptors = spheresAreSetUp && allTexturesTo1x1KnobChanged;
-    const bool setupSphereResources    = (!spheresAreSetUp) && enableSpheres;
-    const bool setupSphereMeshes       = ((!spheresAreSetUp) || requireMoreSpheres) && enableSpheres;
-    const bool rebuildSpherePipeline   = setupSphereMeshes || (spheresAreSetUp && (alphaBlendKnobChanged || depthTestWriteKnobChanged));
-
     // TODO: Ideally, the `maxValue` of the drawcall-count slider knob should be changed at runtime.
     // Currently, the value of the drawcall-count is adjusted to the sphere-count in case the
     // former exceeds the value of the sphere-count.
     if (pDrawCallCount->GetValue() > pSphereInstanceCount->GetValue()) {
         pDrawCallCount->SetValue(pSphereInstanceCount->GetValue());
     }
+
+    const bool enableSpheres = pEnableSpheres->GetValue();
 
     // Set visibilities
     if (enableSpheresKnobChanged) {
@@ -912,22 +908,27 @@ void GraphicsBenchmarkApp::ProcessKnobs()
     }
     pAllTexturesTo1x1->SetVisible(enableSpheres && (pKnobPs->GetValue() == SpherePS::SPHERE_PS_MEM_BOUND));
 
-    // Update sphere resources and mesh
-    if (setupSphereResources) {
-        SetupSphereResources();
-    }
-    if (setupSphereMeshes) {
-        SetupSphereMeshes();
-    }
-
-    // Rebuild pipelines
-    if (rebuildSpherePipeline) {
-        SetupSpheresPipelines();
-    }
-
-    // Update descriptors
-    if (updateSphereDescriptors) {
-        UpdateSphereDescriptors();
+    if (enableSpheres) {
+        // Update sphere resources and mesh
+        const bool spheresAreSetUp = mSpheresAreSetUp;
+        if (!spheresAreSetUp) {
+            // This creates all resources.
+            SetupSpheres();
+        }
+        else {
+            const bool requireMoreSpheres = sphereInstanceCountKnobChanged && (pSphereInstanceCount->GetValue() > mInitializedSpheres);
+            if (requireMoreSpheres) {
+                SetupSphereMeshes();
+            }
+            // Rebuild pipelines
+            if (alphaBlendKnobChanged || depthTestWriteKnobChanged) {
+                SetupSpheresPipelines();
+            }
+            // Update descriptors
+            if (allTexturesTo1x1KnobChanged) {
+                UpdateSphereDescriptors();
+            }
+        }
     }
 
     ProcessQuadsKnobs();
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
@@ -90,7 +90,7 @@ void GraphicsBenchmarkApp::InitKnobs()
     pSphereInstanceCount->SetFlagDescription("Select the number of spheres to draw on the screen.");
     pSphereInstanceCount->SetIndent(1);
 
-    GetKnobManager().InitKnob(&pDrawCallCount, "drawcall-count", /* defaultValue = */ 1, /* minValue = */ 1, kMaxSphereInstanceCount);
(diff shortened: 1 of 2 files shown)


src/ppx/knob.cpp
```diff
@@ -49,7 +49,6 @@ KnobCheckbox::KnobCheckbox(const std::string& flagName, bool defaultValue)
     : Knob(flagName, true), mValue(defaultValue), mDefaultValue(defaultValue)
 {
     SetFlagParameters("<true|false>");
-    RaiseUpdatedFlag();
```
angela28chen, line 52: Removing `RaiseUpdatedFlag()` from different knob classes is going to affect the behaviour of all knobs. I don't think this PR adequately addresses that.

elviscapiaq: Thanks for pointing it. I restored the calls to `RaiseUpdatedFlag()` and updated the PR.

wangra-google (approved):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -694,28 +700,34 @@ void GraphicsBenchmarkApp::ProcessKnobs()
...
+    const bool enableSpheresKnobChanged      = pEnableSpheres->DigestUpdate();
+    const bool initializedSpheresKnobChanged = pInitializedSpheres->DigestUpdate();
 
     const bool enableSpheres = pEnableSpheres->GetValue();
     // If the LOD is empty, assuming we skipped the setup for all sphere resources at start
     // So need to do the intial setup for all resources here
     const bool spheresAreSetUp               = !mSphereLODs.empty();
     const bool updateSphereDescriptors       = spheresAreSetUp && allTexturesTo1x1KnobChanged;
-    const bool setupSphereResourcesAndMeshes = (!spheresAreSetUp) && enableSpheres;
+    const bool setupSphereResourcesAndMeshes = (!spheresAreSetUp || initializedSpheresKnobChanged) && enableSpheres;
```
Keenuts, line 714: Sphere resources are not cleaned up. Right now, we re-create new descriptors while keeping the old ones, exhausting the pool. This needs to be fixed before this knob can be used.

The other issue is we initialize spheres twice by default:
 - once because the pEnableSpheres bit is set by default.
 - then here, because the initializedSpheresKnob by default is marked as changed.

Keenuts (changes requested):

hysw:
I don't think we need to have a separate knob for it
i.e. [eeb387c9](https://github.com/google/bigwheels/commit/eeb387c93a75b31e62adcfe3fe773a3d27d966d6)

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
elviscapiaq: Updated.

elviscapiaq:
Thank you for all your feedback. I have modified the essence of this PR and implemented on-demand initialization for the spheres. Please let me know your thoughts.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -544,19 +535,37 @@ void GraphicsBenchmarkApp::SetupSkyBoxMeshes()
...
 
     // Create the meshes
     uint32_t meshIndex = 0;
     for (const auto& lod : kAvailableLODs) {
         PPX_LOG_INFO("LOD: " << lod.name);
         SphereMesh sphereMesh(/* radius = */ 1, lod.value.longitudeSegments, lod.value.latitudeSegments);
         sphereMesh.ApplyGrid(grid);
-
         // Create a giant vertex buffer for each vb type to accommodate all copies of the sphere mesh
+        if (mSphereMeshes[meshIndex] != nullptr) {
```
Keenuts, line 551: Is there a case where not all meshes are initialized?
I think it would be similar to just add this before the loop:

```
for (auto& mesh : mSphereMeshes)
    GetDevice()->DestroyMesh(mesh);
```

elviscapiaq: I think at the beginning all meshes are nullptr,. Updated.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -544,15 +535,25 @@ void GraphicsBenchmarkApp::SetupSkyBoxMeshes()
...
+    const uint32_t requiredSphereCount = std::max<uint32_t>(pSphereInstanceCount->GetValue(), kDefaultSphereInstanceCount);
+    const uint32_t initSphereCount     = std::min<uint32_t>(kMaxSphereInstanceCount, 2 * std::max(requiredSphereCount, mInitializedSpheres));
+
+    OrderedGrid grid(initSphereCount, kSeed);
+    mInitializedSpheres = initSphereCount;
+
+    // Destroy the meshes if they were created.
+    for (auto& mesh : mSphereMeshes) {
+        if (mesh != nullptr) {
+            GetDevice()->DestroyMesh(mesh);
```
hysw, line 547: nit: reset it back to nullptr after you destroyed it, try not to leave dangling pointer around.

elviscapiaq: Done.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -874,13 +876,14 @@ void GraphicsBenchmarkApp::ProcessKnobs()
...
     const bool enableSpheresKnobChanged    = pEnableSpheres->DigestUpdate();
+    const bool requireMoreSpheres          = (pSphereInstanceCount->GetValue() > mInitializedSpheres);
 
     const bool enableSpheres = pEnableSpheres->GetValue();
     // If the LOD is empty, assuming we skipped the setup for all sphere resources at start
     // So need to do the intial setup for all resources here
     const bool spheresAreSetUp               = mSpheresAreSetUp;
     const bool updateSphereDescriptors       = spheresAreSetUp && allTexturesTo1x1KnobChanged;
-    const bool setupSphereResourcesAndMeshes = (!spheresAreSetUp) && enableSpheres;
+    const bool setupSphereResourcesAndMeshes = (!spheresAreSetUp || requireMoreSpheres) && enableSpheres;
```
angela28chen, line 886: I believe this change will cause `SetupSphereResources()` to be called more than once, but that has logic like loading shaders which probably should only be called once.

elviscapiaq: Update.

elviscapiaq:
> Note that because of the way imGUI sliders work, the value will be adjusted as the slider is dragged and not when it's let go. So with this change you will need to drag the slider ~4 times upwards and wait while it buffers to slowly make it to 3000, it's a little frustrating to use. Do we want to change it so that the # of spheres only change after the slider is let go? That can probably be done if a separate variable is kept.

Thanks for this feedback. I fixed this issue and now the spheres are initialized only when the slider is let go.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -544,15 +535,26 @@ void GraphicsBenchmarkApp::SetupSkyBoxMeshes()
...
 void GraphicsBenchmarkApp::SetupSphereMeshes()
 {
-    OrderedGrid grid(kMaxSphereInstanceCount, kSeed);
+    const uint32_t requiredSphereCount = std::max<uint32_t>(pSphereInstanceCount->GetValue(), kDefaultSphereInstanceCount);
+    const uint32_t initSphereCount     = std::min<uint32_t>(kMaxSphereInstanceCount, 2 * std::max(requiredSphereCount, mInitializedSpheres));
+
+    OrderedGrid grid(initSphereCount, kSeed);
+    mInitializedSpheres = initSphereCount;
+
+    // Destroy the meshes if they were created.
```
Keenuts, line 544: This triggers a Vulkan validation error:
The destroy can be called while the buffer is still used by the previous frame.
We might want to do a WaitForIdle (quick solution, don't need perf there) or something more fine grained to prevent those issues.

elviscapiaq: Done.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -1593,9 +1607,9 @@ void GraphicsBenchmarkApp::RecordCommandBufferSpheres(PerFrame& frame)
     frame.cmd->BindGraphicsDescriptorSets(mSphere.pipelineInterface, 1, &mSphere.descriptorSets.at(GetInFlightFrameIndex()));
 
     // Snapshot some scene-related values for the current frame
-    uint32_t currentSphereCount   = pSphereInstanceCount->GetValue();
+    uint32_t currentSphereCount   = std::min<uint32_t>(pSphereInstanceCount->GetValue(), mInitializedSpheres);
```
wangra-google, line 1610: in what case `pSphereInstanceCount` could be larger than `mInitializedSpheres`?

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -1593,9 +1607,9 @@ void GraphicsBenchmarkApp::RecordCommandBufferSpheres(PerFrame& frame)
...
     uint32_t currentDrawCallCount = pDrawCallCount->GetValue();
-    uint32_t mSphereIndexCount    = mSphereMeshes[meshIndex]->GetIndexCount() / kMaxSphereInstanceCount;
+    uint32_t mSphereIndexCount    = (mInitializedSpheres > 0 ? (mSphereMeshes[meshIndex]->GetIndexCount() / mInitializedSpheres) : 0);
```
wangra-google, line 1612: in what case `mInitializedSpheres` could be 0? If i understand correctly, only when `SetupSphereMeshes` is skipped,  mInitializedSpheres is 0? but in that case, `pEnableSpheres` is false, and  `RecordCommandBufferSpheres` would be skipped as well?

Keenuts: I don't think those case could be reached as-is. But since those values depends on elements external to this function, I would be in favor of keeping this check and min.
Just so we say "in case this happens, we have a defined behavior to handle it", even if if wouldn't happen today.

Keenuts (approved):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
wangra-google: if `mInitializedSpheres` should never be 0, maybe better with an assert?

angela28chen (approved):
Not something that needs to change for this PR, but I think that there might be a need for two different types of sliders for slider knobs, one where the value is reflected immediately during dragging and another where it's only updated once the slider is released. The current slider knob logic is not really consistent with itself. I think that `pSphereInstanceCount` shows this because `RecordCommandBufferSpheres()` will use the value (which is changing as the knob is getting dragged) whereas the vertex buffer will be using the value only after it has been marked as updated (which is when the knob is released).

So this is probably fine for this PR if dragging around the sphere slider isn't causing any index out of bounds crashes, but it's something that should be changed in knob implementation.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
elviscapiaq: Updated.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -874,14 +877,16 @@ void GraphicsBenchmarkApp::ProcessKnobs()
...
     // If the LOD is empty, assuming we skipped the setup for all sphere resources at start
     // So need to do the intial setup for all resources here
-    const bool spheresAreSetUp               = mSpheresAreSetUp;
-    const bool updateSphereDescriptors       = spheresAreSetUp && allTexturesTo1x1KnobChanged;
-    const bool setupSphereResourcesAndMeshes = (!spheresAreSetUp) && enableSpheres;
-    const bool rebuildSpherePipeline         = setupSphereResourcesAndMeshes || (spheresAreSetUp && (alphaBlendKnobChanged || depthTestWriteKnobChanged));
+    const bool spheresAreSetUp         = mSpheresAreSetUp;
+    const bool updateSphereDescriptors = spheresAreSetUp && allTexturesTo1x1KnobChanged;
+    const bool setupSphereResources    = (!spheresAreSetUp) && enableSpheres;
+    const bool setupSpheredMeshes      = ((!spheresAreSetUp) || requireMoreSpheres) && enableSpheres;
```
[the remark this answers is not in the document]
elviscapiaq: Done.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -874,14 +877,16 @@ void GraphicsBenchmarkApp::ProcessKnobs()
     const bool alphaBlendKnobChanged       = pAlphaBlend->DigestUpdate();
     const bool depthTestWriteKnobChanged   = pDepthTestWrite->DigestUpdate();
     const bool enableSpheresKnobChanged    = pEnableSpheres->DigestUpdate();
+    const bool requireMoreSpheres          = pSphereInstanceCount->DigestUpdate() && (pSphereInstanceCount->GetValue() > mInitializedSpheres);
```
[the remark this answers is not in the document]
elviscapiaq: Done.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
elviscapiaq: It can occur when we drag the `sphereCount` slider, because we only update the `initializedSpheres` value once the slider is released. In other words, we can drag the slider, and `sphereCount` can increase, but `initializedSpheres` would remain the same up to the slider is let go.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
Keenuts: Yes, an assert is also fine if that's an illegal behavior.
But I'd feel like having the slider default to 0 is a valid behavior.

wangra-google: the min value of pSphereInstanceCount is set to 1, so i guess 0 is not valid.

wangra-google: actually, maybe it is better to skip `RecordCommandBufferSpheres()` entirely if `mInitializedSpheres` is 0?

elviscapiaq: I agree with that change. Updated.

elviscapiaq:
PR was rebased.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -872,18 +875,21 @@ void GraphicsBenchmarkApp::ProcessKnobs()
...
-    const bool spheresAreSetUp               = mSpheresAreSetUp;
-    const bool updateSphereDescriptors       = spheresAreSetUp && allTexturesTo1x1KnobChanged;
-    const bool setupSphereResourcesAndMeshes = (!spheresAreSetUp) && enableSpheres;
-    const bool rebuildSpherePipeline         = setupSphereResourcesAndMeshes || (spheresAreSetUp && (alphaBlendKnobChanged || depthTestWriteKnobChanged));
+    const bool spheresAreSetUp         = mSpheresAreSetUp;
+    const bool requireMoreSpheres      = sphereInstanceCountKnobChanged && (pSphereInstanceCount->GetValue() > mInitializedSpheres);
+    const bool updateSphereDescriptors = spheresAreSetUp && allTexturesTo1x1KnobChanged;
+    const bool setupSphereResources    = (!spheresAreSetUp) && enableSpheres;
+    const bool setupSphereMeshes       = ((!spheresAreSetUp) || requireMoreSpheres) && enableSpheres;
+    const bool rebuildSpherePipeline   = setupSphereMeshes || (spheresAreSetUp && (alphaBlendKnobChanged || depthTestWriteKnobChanged));
```
apazylbe, line 892: Do we need to rebuild pipelines if there was an increase in the number of spheres but nothing else changed?

apazylbe, line 892: This is a bit hard to follow, and used later in the code where you have to scroll up to figure out when something is rebuilt. Also it's a little weird to update `mSpheresAreSetUp` in `SetupSphereMeshes()`, since mSpheresAreSetUp basically means everything is setup instead of just spheres. So if I call just `SetupSpheresMeshes()` mSpheresAreSetUp will be true. 
Maybe not for this PR, but we should  probably consider cleaning up the logic a bit, one alternative: https://github.com/google/bigwheels/commit/9e5c0b8319a8f6503d3631cee46285aab3c68fa5

apazylbe: Not directly related to this PR, but I wonder if it should also take into account the other knob changes (vertex buffer format, layout, and shaders) or if it can be ommitted altogether since `GetSpherePipeline()` in `RecordCommandBufferSpheres` will create the necessary pipeline anyway

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -90,7 +90,7 @@ void GraphicsBenchmarkApp::InitKnobs()
     pSphereInstanceCount->SetFlagDescription("Select the number of spheres to draw on the screen.");
     pSphereInstanceCount->SetIndent(1);
 
-    GetKnobManager().InitKnob(&pDrawCallCount, "drawcall-count", /* defaultValue = */ 1, /* minValue = */ 1, kMaxSphereInstanceCount);
+    GetKnobManager().InitKnob(&pDrawCallCount, "drawcall-count", /* defaultValue = */ kDefaultSphereInstanceCount, /* minValue = */ 1, kMaxSphereInstanceCount);
```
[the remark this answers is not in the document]
elviscapiaq: Good catch. Actually, the default value for pSphereInstanceCount shoud be kDefaultSphereInstanceCount but for drawcall should be 1. Updated.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -546,15 +537,27 @@ void GraphicsBenchmarkApp::SetupSkyBoxMeshes()
 
 void GraphicsBenchmarkApp::SetupSphereMeshes()
 {
-    OrderedGrid grid(kMaxSphereInstanceCount, kSeed);
+    const uint32_t requiredSphereCount = std::max<uint32_t>(pSphereInstanceCount->GetValue(), kDefaultSphereInstanceCount);
+    const uint32_t initSphereCount     = std::min<uint32_t>(kMaxSphereInstanceCount, 2 * std::max(requiredSphereCount, mInitializedSpheres));
+
+    OrderedGrid grid(initSphereCount, kSeed);
```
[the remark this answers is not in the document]
elviscapiaq: Done.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
elviscapiaq: That is a good point, actually we don't need to rebuild pipelines. so I applied your proposed clean up from the other comment.

elviscapiaq: Thank you for the comment. I applied your suggested clean up and the code looks more readable now.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -906,20 +908,27 @@ void GraphicsBenchmarkApp::ProcessKnobs()
...
-        SetupSpheresPipelines();
-    }
-
-    // Update descriptors
-    if (updateSphereDescriptors) {
-        UpdateSphereDescriptors();
+    if (enableSpheres) {
+        // Update sphere resources and mesh
+        const bool spheresAreSetUp = mSpheresAreSetUp;
+        if (!spheresAreSetUp) {
```
apazylbe, line 914: I think we can actually inline now and just check !mSpheresAreSetUp here

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -906,20 +908,27 @@ void GraphicsBenchmarkApp::ProcessKnobs()
...
+            // This creates all resources.
+            SetupSpheres();
+        }
+        else {
+            const bool requireMoreSpheres = sphereInstanceCountKnobChanged && (pSphereInstanceCount->GetValue() > mInitializedSpheres);
+            if (requireMoreSpheres) {
+                SetupSphereMeshes();
+            }
+            // Rebuild pipelines
+            if (alphaBlendKnobChanged || depthTestWriteKnobChanged) {
```
apazylbe, line 924: Sphere pipeline seems to be dependent on some other knobs like PS, VS, vertex buffer: https://github.com/google/bigwheels/blob/4c9fbb05e77d384456594e6b9ea165b7d0bbcca2/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp#L738-L747
Do we need to add those here, or perhaps just remove this since I think when we record a command buffer it will compile the necessary pipeline? 
@hysw, @angela28chen

hysw: No, this part can be removed.

`SetupSpheresPipelines()` should be called exactly once, the appropriate pipeline will be built on demand from current knob values.

elviscapiaq: Thanks @hysw  for confirming it. Updated.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
elviscapiaq: Done.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
elviscapiaq: Removed.

apazylbe (approved):

PR #353, opened 2023-11-21
Use metrics system for cpu submission time and bandwidth in the benchmark 

wangra-google:


mdagois:
LGTM for the ppx parts.
I am not too familiar with the graphics benchmark, so apazylbe and angela28chen are probably better fits for reviewing it.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
```diff
@@ -254,6 +247,20 @@ class GraphicsBenchmarkApp
     std::array<grfx::PipelineInterfacePtr, kFullscreenQuadsTypes.size()> mQuadsPipelineInterfaces;
     std::array<grfx::ShaderModulePtr, kFullscreenQuadsTypes.size()>      mQuadsPs;
 
+    // Metrics Data
+    struct MetricsData
+    {
+        enum EMetricsTypes : size_t
```
apazylbe, line 253: Is that "E" in front of "MetricTypes" for "Enum"? Is that a naming convention?

src/ppx/application.cpp
```diff
@@ -786,6 +786,15 @@ void Application::ShutdownMetrics()
     StopMetricsRun();
 }
 
+metrics::GaugeBasicStatistics Application::GetGaugeBasicStatistics(metrics::MetricID id) const
+{
+    if (!mStandardOpts.pEnableMetrics->GetValue()) {
```
apazylbe, line 791: Perhaps this should produce a "stronger" assert/error, cause otherwise the data is invalid, and you might not realize. I guess depends on the usage though, whether it is only called if metrics are enabled, or can be called as a noop otherwise.

apazylbe (changes requested):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -234,6 +234,56 @@ void GraphicsBenchmarkApp::Setup()
     }
 }
 
+void GraphicsBenchmarkApp::SetupMetrics()
+{
+    Application::SetupMetrics();
```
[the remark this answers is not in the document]
wangra-google: fixed

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -234,6 +234,56 @@ void GraphicsBenchmarkApp::Setup()
...
+
+    const float    gpuWorkDurationInSec = static_cast<float>(mGpuWorkDuration / static_cast<double>(frequency));
+    const uint32_t width                = GetSwapchain()->GetWidth();
+    const uint32_t height               = GetSwapchain()->GetHeight();
+    const uint32_t quadCount            = pFullscreenQuadsCount->GetValue();
+
+    if (quadCount) {
+        // Skip the first kSkipFrameCount frames after the knob of quad count being changed to avoid noise
+        constexpr uint32_t kSkipFrameCount  = 2;
+        static uint32_t    skipFrameCounter = 0;
```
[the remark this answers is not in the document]
wangra-google: changed to class member

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
wangra-google: yes, but not consistent with other enum naming in BW, removed E

src/ppx/application.cpp
wangra-google: changed into assert

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -234,6 +234,56 @@ void GraphicsBenchmarkApp::Setup()
...
+
+    metadata                                          = {ppx::metrics::MetricType::GAUGE, "Bandwidth", "GB/s", ppx::metrics::MetricInterpretation::HIGHER_IS_BETTER, {0.f, 10000.f}};
+    mMetricsData.metrics[MetricsData::kTypeBandwidth] = AddMetric(metadata);
+    PPX_ASSERT_MSG(mMetricsData.metrics[MetricsData::kTypeBandwidth] != ppx::metrics::kInvalidMetricID, "Failed to add Bandwidth metric");
+}
+
+void GraphicsBenchmarkApp::UpdateMetrics()
+{
+    uint64_t frequency = 0;
+    GetGraphicsQueue()->GetTimestampFrequency(&frequency);
```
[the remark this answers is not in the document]
wangra-google: fixed

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -1104,25 +1147,28 @@ void GraphicsBenchmarkApp::UpdateGUI()
...
-    ImGui::NextColumn();
+        ImGui::Text("CPU Max Submission Time");
+        ImGui::NextColumn();
+        ImGui::Text("%.2f ms", cpuSubmissionTime.max);
+        ImGui::NextColumn();
+    }
 
+    uint64_t frequency = 0;
+    PPX_CHECKED_CALL(GetGraphicsQueue()->GetTimestampFrequency(&frequency));
     ImGui::Columns(2);
```
angela28chen, line 1172: nit: maybe this line and line 1153 could just be moved to right above the `HasActiveMetricsRun` if block.

angela28chen (approved):
Looks good to me

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -234,6 +234,56 @@ void GraphicsBenchmarkApp::Setup()
...
+    if (quadCount) {
+        // Skip the first kSkipFrameCount frames after the knob of quad count being changed to avoid noise
+        constexpr uint32_t kSkipFrameCount  = 2;
+        static uint32_t    skipFrameCounter = 0;
+        if (pFullscreenQuadsCount->DigestUpdate()) {
+            skipFrameCounter = kSkipFrameCount;
+        }
+
+        if (skipFrameCounter == 0) {
+            const float dataWriteInGb = (static_cast<float>(width) * static_cast<float>(height) * 4.f * quadCount) / (1024.f * 1024.f * 1024.f);
```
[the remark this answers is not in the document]
wangra-google: this is not tracked by the metrics system since it depends on the number of quads, if the user changes the quad count, the average/min/max do not make much sense anymore.

On the other hand, i still want to show this value in the UI so that we know what is the total amount of data being output.

I could cache this value, but since it is not really expensive to calculate this, it is better to just leave it calculated two times.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
wangra-google: fixed

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
```diff
@@ -414,6 +414,20 @@ class GraphicsBenchmarkApp
     std::array<grfx::ShaderModulePtr, kFullscreenQuadsTypes.size()>      mQuadsPs;
 
     BlitContext mBlit;
+    // Metrics Data
+    struct MetricsData
+    {
+        enum MetricsTypes : size_t
```
apazylbe, line 420: nit: I think it should be MetricType since each value represents a single type, but feel free to ignore me here if you think otherwise

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
```diff
@@ -414,6 +414,20 @@ class GraphicsBenchmarkApp
...
+        {
+            kTypeCPUSubmissionTime = 0,
+            kTypeBandwidth,
+            kCount
+        };
+
+        ppx::metrics::MetricID metrics[kCount] = {};
+    };
+    MetricsData mMetricsData;
+    uint32_t    mSkipRecordBandwidthMetricFrameCounter = 0;
```
apazylbe, line 430: Could you add a short comment for this member?

include/ppx/metrics.h
```diff
@@ -172,7 +186,7 @@ class MetricGauge final : public Metric
 private:
     MetricMetadata               mMetadata;
     std::vector<TimeSeriesEntry> mTimeSeries;
-    Stats                        mBasicStats;
+    Stats                        mStats;
```
apazylbe, line 189: I think it was storing only the basic stats in this struct before, since you are splitting the Stats in two parts, one being the basic stats, should this just be GaugeBasicStatistics instead? And then we can actually construct Stats later. Otherwise it sounds like the complex stats are just not valid in this struct

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -291,6 +291,60 @@ void GraphicsBenchmarkApp::Setup()
...
+
+        metadata                                          = {ppx::metrics::MetricType::GAUGE, "Bandwidth", "GB/s", ppx::metrics::MetricInterpretation::HIGHER_IS_BETTER, {0.f, 10000.f}};
+        mMetricsData.metrics[MetricsData::kTypeBandwidth] = AddMetric(metadata);
+        PPX_ASSERT_MSG(mMetricsData.metrics[MetricsData::kTypeBandwidth] != ppx::metrics::kInvalidMetricID, "Failed to add Bandwidth metric");
+    }
+}
+
+void GraphicsBenchmarkApp::UpdateMetrics()
+{
+    if (HasActiveMetricsRun()) {
```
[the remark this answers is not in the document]
wangra-google: done

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
wangra-google: fixed

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
wangra-google: added as suggested

include/ppx/metrics.h
wangra-google: the original also stores both but the name was named as mBasicStats, but I do agree with you that there is no point of storing the complex here. Changed as suggested

include/ppx/metrics.h
```diff
@@ -167,12 +175,12 @@ class MetricGauge final : public Metric
...
     METRICS_NO_COPY(MetricGauge)
 
-    Stats ComputeStats() const;
+    GaugeComplexStatistics ComputeComplexStats() const;
 
 private:
     MetricMetadata               mMetadata;
     std::vector<TimeSeriesEntry> mTimeSeries;
-    Stats                        mBasicStats;
+    GaugeBasicStatistics         mBasic;
```
apazylbe, line 183: nit: can we keep the name as `mBasicStats`? Otherwise, it is not clear "basic" what?

apazylbe (approved):

include/ppx/metrics.h
wangra-google: fixed

PR #355, opened 2023-11-21
Add StandardKnobsDefaultValue in mSettings to allow derived classes overwrite default standard knob values

wangra-google:
Add StandardKnobsDefaultValue in mSettings to allow derived classes overwrite default standard knob values


include/ppx/application.h
```diff
@@ -527,6 +527,11 @@ class Application
     // Initializes standard knobs
     void InitStandardKnobs();
 
+    // This is used by derived class to overwrite standard knobs default value
+    virtual void SetupStandardKnobsDefaultValue()
```
hysw, line 531: Maybe `virtual void SetupStandardKnobsDefaultValue(StandardKnobsDefaultValue&)`?

wangra-google: i think the name was confusing, changed the name to `OverwriteStandardKnobsDefaultValue` to be clear that it is only for overwriting the default values

hysw: What I mean is passing down a reference instead of using `protected:`, the derived class should not touch the default value outside of this function anyway. Similar to `Config(ppx::ApplicationSettings&)`

hysw: Context: "Make classes' data members private, unless they are constants." [cppguide#Access_Control](https://google.github.io/styleguide/cppguide.html#Access_Control) (also in go/cstyle)
BigWheels does not follow all the google style guide for reasons, but I'd keep to the style guide when reasonable.

wangra-google: Thanks for the link. Make sense, changed as suggested

hysw (commented):
LGTM

Given this is a change on knob processing, I'd like to see Angela's opinion on it.

Keenuts:
What is the use case for this?
What I'd be worried about is unpredictable behaviors across samples: if I run 2 samples with the same set of flags, I'd expect both to run with the same options.

wangra-google:
> What is the use case for this? What I'd be worried about is unpredictable behaviors across samples: if I run 2 samples with the same set of flags, I'd expect both to run with the same options.

in the case of benchmark, to avoid duplication of calculating metrics (average, min, max) we start to use the metrics system to get some metrics and display in the UI, and we want to enable the metrics by default which is not the case now since it is disabled in the base class in application.cpp. I assume some other applications might have similar requirement to change the default behaviors of some of the StandardKnobs

wangra-google:
> > in the case of benchmark, to avoid duplication of calculating metrics (average, min, max) we start to use the metrics system to get some metrics and display in the UI, and we want to enable the metrics by default which is not the case now since it is disabled in the base class in application.cpp. I assume some other applications might have similar requirement to change the default behaviors of some of the StandardKnobs
> 
> Thanks for the context! Right I see. What about setting this behavior as default for all the benchmark applications?



> > in the case of benchmark, to avoid duplication of calculating metrics (average, min, max) we start to use the metrics system to get some metrics and display in the UI, and we want to enable the metrics by default which is not the case now since it is disabled in the base class in application.cpp. I assume some other applications might have similar requirement to change the default behaviors of some of the StandardKnobs
> 
> Thanks for the context! Right I see. What about setting this behavior as default for all the benchmark applications?



> > in the case of benchmark, to avoid duplication of calculating metrics (average, min, max) we start to use the metrics system to get some metrics and display in the UI, and we want to enable the metrics by default which is not the case now since it is disabled in the base class in application.cpp. I assume some other applications might have similar requirement to change the default behaviors of some of the StandardKnobs
> 
> Thanks for the context! Right I see. What about setting this behavior as default for all the benchmark applications?

This is not the only option i would like to change for benchmark, I am worried that some behaviors might not be desirable for all projects. I feel it would be more flexible to allow the derived class to change the settings as needed.

> > in the case of benchmark, to avoid duplication of calculating metrics (average, min, max) we start to use the metrics system to get some metrics and display in the UI, and we want to enable the metrics by default which is not the case now since it is disabled in the base class in application.cpp. I assume some other applications might have similar requirement to change the default behaviors of some of the StandardKnobs
> 
> Thanks for the context! Right I see. What about setting this behavior as default for all the benchmark applications?

As we discussed offline, I moved StandardKnobsDefaultValue to mSettings to allow derived classes overwrite default standard knob values so that it is more consistent with the other application settings

src/ppx/application.cpp
```diff
@@ -649,10 +649,13 @@ void Application::DispatchInitKnobs()
     InitKnobs();
 }
 
-void Application::DispatchConfig()
+void Application::DispatchConfigAndInitKnobs()
 {
     Config(mSettings);
 
+    // Knobs need to be set up before commandline parsing.
+    DispatchInitKnobs();
```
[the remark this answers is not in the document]
angela28chen: `mKnobManager.UpdateFromFlags(options);` now runs after the first usage of a knob value.

angela28chen (changes requested):
> What I'd be worried about is unpredictable behaviors across samples: if I run 2 samples with the same set of flags, I'd expect both to run with the same options.

I agree with this point, I think it would be inconsistent.

1) If we do go ahead with letting applications set their own default values for standard knobs, maybe it would help if the usage message printed by  `--help` indicates default values for all flags?
2) Instead of this change could we recommend a way for people to run benchmark with flags? E.g. add a README.md file and put the suggested command, something like `vk_graphics_benchmark.exe --enable-metrics`. I have been thinking that it's useful to have some documentation for benchmark too because I don't think there's easily accessible info right now about the controls (WASD, spacebar)

wangra-google:
> > What I'd be worried about is unpredictable behaviors across samples: if I run 2 samples with the same set of flags, I'd expect both to run with the same options.
> 
> I agree with this point, I think it would be inconsistent.
> 
> 1. If we do go ahead with letting applications set their own default values for standard knobs, maybe it would help if the usage message printed by  `--help` indicates default values for all flags?
> 2. Instead of this change could we recommend a way for people to run benchmark with flags? E.g. add a README.md file and put the suggested command, something like `vk_graphics_benchmark.exe --enable-metrics`. I have been thinking that it's useful to have some documentation for benchmark too because I don't think there's easily accessible info right now about the controls (WASD, spacebar)

Thanks for the suggestion, and I am open to any options, and it is great that we have the discussion here. :) Take the metrics for example, my concern is that if we start to use metrics system in the UI (which make sense to avoid duplicating the came calculation that already exist in metrics system), there will be an implicit dependency between the info shown in the UI and the metrics system. Without setting the default values at per-project basis, the users need to know (or check source code) if such dependency exists before knowing what command to use, or we have to document the command line for each project. For debugging purpose, having too many cmd specific args could be error prone as people would forget to add some args and some features could be disabled implicitly.

src/ppx/application.cpp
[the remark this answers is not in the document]
hysw: The change of initialization here is confusing.
`pAssetsPaths` sounds like a separate PR.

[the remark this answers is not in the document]
wangra-google: Thanks for pointing this out, it is indeed an issue. I did a cleanup of the DispatchConfig() so that it is only doing dispatching config instead of doing more than the name indicated to avoid confusion. And moved all the asset dir related code to a separate function with some comments so that it is clear what are the dependencies.

wangra-google:
> > > What I'd be worried about is unpredictable behaviors across samples: if I run 2 samples with the same set of flags, I'd expect both to run with the same options.
> > 
> > 
> > I agree with this point, I think it would be inconsistent.
> > 
> > 1. If we do go ahead with letting applications set their own default values for standard knobs, maybe it would help if the usage message printed by  `--help` indicates default values for all flags?
> > 2. Instead of this change could we recommend a way for people to run benchmark with flags? E.g. add a README.md file and put the suggested command, something like `vk_graphics_benchmark.exe --enable-metrics`. I have been thinking that it's useful to have some documentation for benchmark too because I don't think there's easily accessible info right now about the controls (WASD, spacebar)
> 
> Thanks for the suggestion, and I am open to any options, and it is great that we have the discussion here. :) Take the metrics for example, my concern is that if we start to use metrics system in the UI (which make sense to avoid duplicating the came calculation that already exist in metrics system), there will be an implicit dependency between the info shown in the UI and the metrics system. Without setting the default values at per-project basis, the users need to know (or check source code) if such dependency exists before knowing what command to use, or we have to document the command line for each project. For debugging purpose, having too many cmd specific args could be error prone as people would forget to add some args and some features could be disabled implicitly.



> > > What I'd be worried about is unpredictable behaviors across samples: if I run 2 samples with the same set of flags, I'd expect both to run with the same options.
> > 
> > 
> > I agree with this point, I think it would be inconsistent.
> > 
> > 1. If we do go ahead with letting applications set their own default values for standard knobs, maybe it would help if the usage message printed by  `--help` indicates default values for all flags?
> > 2. Instead of this change could we recommend a way for people to run benchmark with flags? E.g. add a README.md file and put the suggested command, something like `vk_graphics_benchmark.exe --enable-metrics`. I have been thinking that it's useful to have some documentation for benchmark too because I don't think there's easily accessible info right now about the controls (WASD, spacebar)
> 
> Thanks for the suggestion, and I am open to any options, and it is great that we have the discussion here. :) Take the metrics for example, my concern is that if we start to use metrics system in the UI (which make sense to avoid duplicating the came calculation that already exist in metrics system), there will be an implicit dependency between the info shown in the UI and the metrics system. Without setting the default values at per-project basis, the users need to know (or check source code) if such dependency exists before knowing what command to use, or we have to document the command line for each project. For debugging purpose, having too many cmd specific args could be error prone as people would forget to add some args and some features could be disabled implicitly.

With my latest commits, the commandline flags will always have the highest priority which means even the app changed the settings in the derived class's ::Config(), the commandline flags will overwrite the settings. In this way, 2 samples are running with the same set of flags, both will be running with the same options. On the other hand, if the commandline flags are not used, the application will be able to overwrite the default settings. 

I totally agree with adding a readme file to explain all the settings, but would be better in a separate PR.

Let me know if you have other concerns. thanks :)

include/ppx/application.h
```diff
@@ -323,6 +323,33 @@ struct ApplicationSettings
...
+    struct StandardKnobsDefaultValue
+    {
+        std::vector<std::string> assetsPaths     = {};
+        std::vector<std::string> configJsonPaths = {};
+        bool                     deterministic   = false;
+        bool                     enableMetrics   = false;
+        uint64_t                 frameCount      = UINT64_MAX;
+        int                      gpuIndex        = INT_MAX;
+#if !defined(PPX_LINUX_HEADLESS)
+        bool headless = false;
```
apazylbe, line 337: There is already `headless` option above in `ApplicationSettings`, so it seems weird to have this twice. Which one takes priority if there are different, can they even be different.

src/ppx/application.cpp
```diff
@@ -652,31 +652,6 @@ void Application::DispatchInitKnobs()
 void Application::DispatchConfig()
 {
     Config(mSettings);
-
-    if (mSettings.appName.empty()) {
```
apazylbe, line 656: I guess taking this out is a slight change from the original behaviour, since DispatchConfig can be overriden by applications, and now it can't modify this. But it might be for the best

src/ppx/application.cpp
```diff
@@ -1174,7 +1148,12 @@ int Application::Run(int argc, char** argv)
         return EXIT_FAILURE;
     }
 
+    // Call config.
```
apazylbe, line 1151: Run() seems to be a massive function. Perhaps not in this PR, but maybe we should extract pretty much line 1145-1233 into it's own function.

include/ppx/application.h
wangra-google: i feel in this case the knob one should be kept since this can be controlled both by cmdline and overwritten by the application. @angela28chen any thoughts on this?

src/ppx/application.cpp
wangra-google: make sense, i can do another PR to clean this up once this PR is submitted

include/ppx/application.h
angela28chen: I think that the knob one should be kept. Also, I think there is a bug in the `benchmarks\headless_compute` example probably introduced when I changed up the commandline parsing a while ago to make use of knobs. I think the line `settings.headless = true;` in there is ignored, but after this CL it can just be fixed to `settings.standardKnobsDefaultValue.headless = true;`

angela28chen (approved):
> Without setting the default values at per-project basis, the users need to know (or check source code) if such dependency exists before knowing what command to use, or we have to document the command line for each project. For debugging purpose, having too many cmd specific args could be error prone as people would forget to add some args and some features could be disabled implicitly.

Hmm I think you're right about it being more usable like this. OK, seems good to me, I'll put up a PR for adjusting the help message too after this (so there's an easy way to see what the defaults are without checking source code).

wangra-google:
> > Without setting the default values at per-project basis, the users need to know (or check source code) if such dependency exists before knowing what command to use, or we have to document the command line for each project. For debugging purpose, having too many cmd specific args could be error prone as people would forget to add some args and some features could be disabled implicitly.
> 
> Hmm I think you're right about it being more usable like this. OK, seems good to me, I'll put up a PR for adjusting the help message too after this (so there's an easy way to see what the defaults are without checking source code).

cool, thanks! :D

include/ppx/application.h
wangra-google: I feel that headless change would be a bit more involved, especially with the discussion by @hysw. I could do another PR to handle the headless, what do you think @apazylbe ?

Keenuts (approved):
LGTM, nothing more to say than what was already. Thanks!

include/ppx/application.h
```diff
@@ -323,6 +323,33 @@ struct ApplicationSettings
             uint32_t     imageCount  = 2;
         } swapchain;
     } grfx;
+
+    // Default values for standard knobs
+    struct StandardKnobsDefaultValue
+    {
+        std::vector<std::string> assetsPaths     = {};
+        std::vector<std::string> configJsonPaths = {};
```
angela28chen, line 331: Thinking about it, would these two flags `--asset-path` and `--config-json-path` ever require default values? They seem to be flags that would only be used in non-standard usage of benchmarks/projects.

hysw: It has benefit to specify all of them at the same place. If there's use case for them to be updated, they will remain null value

include/ppx/application.h
apazylbe: sgtm

apazylbe (approved):
thanks!

include/ppx/application.h
angela28chen: I see the point about it being nicer to have defaults in the same place, I'm just thinking that these 2 flags should never have non-empty defaults. Any knob value in the config file should just become that knob's default value (avoid having a default config file) and then assets used by default ought to be in the assets folder. I think these two flags should be excluded from `ApplicationSettings` to prevent application writers from setting defaults for them if this is an antipattern that we'd like to avoid.

hysw: If you want to prevent application writer from setting it, put a `const` modifier on it.
Also there does exist usecase for default asset paths/config paths, although to make it useful you need to make the last one take higher priority.
i.e. `--asset-path=platform-default --asset-path=application-default --asset-path=user-override1 --asset-path=user-override2`

hysw: I'd prefer to have default in flags and get rid of `Application::InitializeAssetDirs`

wangra-google: i will address the asset dir in a separate PR

PR #356, opened 2023-11-22
Dynamic rendering and a first version of a sample

apazylbe:
Also updates the version of ImGui to include dynamic rendering support.

Dynamic rendering requires:
- enabling the extension
- adding physical device features for dynamic rendering on device creation
- when creating a graphics pipeline passing a VkRenderingInfoKHR instead
of a renderpass
- Calling vkCmdBeginRenderingKHR and vkCmdEndRenderingKHR
- A dynamic render pass can span multiple command buffers, where the first
one started will be suspending, and the next one resuming.

Also adds a first version of a dynamic rendering sample:
- uses push descriptors for code simplicity, although arguably the sample
should just do one "novel" thing
- Renders the first sphere in one command buffer and the second in another
- ImGUI is rendered using a different render pass since it must not have
a depth attachment.

Required follow up: ClearRenderTarget currently does not work within
a dynamic render pass. We will need to store some information about
render area and render targets.


include/ppx/grfx/vk/vk_command.h
```diff
@@ -38,6 +38,9 @@ class CommandBuffer
     virtual void BeginRenderPassImpl(const grfx::RenderPassBeginInfo* pBeginInfo) override;
     virtual void EndRenderPassImpl() override;
 
+    virtual void BeginRenderingImpl(const grfx::RenderingInfo* renderingInfo) override;
```
chaoticbob, line 41: nit: should be `pRenderingInfo`

include/ppx/grfx/grfx_command.h
```diff
@@ -177,6 +177,21 @@ struct RenderPassBeginInfo
...
+struct RenderingInfo
+{
+    grfx::BeginRenderingFlags flags;
+    grfx::Rect                renderArea = {};
+
+    uint32_t                renderTargetCount                          = 0;
+    grfx::RenderTargetView* pRenderTargetViews[PPX_MAX_RENDER_TARGETS] = {};
+    grfx::DepthStencilView* pDepthStencilView                          = nullptr;
+
+    uint32_t                     RTVClearCount                          = 0;
```
chaoticbob, line 190: Can `renderTargetCount` be used to avoid needing this?

src/ppx/grfx/dx12/dx12_command.cpp
```diff
@@ -243,6 +243,95 @@ void CommandBuffer::EndRenderPassImpl()
...
+
+void CommandBuffer::BeginRenderingImpl(const grfx::RenderingInfo* pRenderingInfo)
+{
+    const grfx::RenderTargetClearValue& cv = pRenderingInfo->RTVClearValues[0];
+    D3D12_CLEAR_VALUE                   clearValue{DXGI_FORMAT_R32G32B32A32_FLOAT};
+    clearValue.Color[0]           = cv.r;
+    clearValue.Color[1]           = cv.g;
+    clearValue.Color[2]           = cv.b;
+    clearValue.Color[3]           = cv.a;
+    clearValue.DepthStencil.Depth = 1.0f;
```
chaoticbob, line 291: Should depth/stencil read their respective values from pRenderingInfo?

include/ppx/grfx/vk/vk_command.h
apazylbe: Done, thanks!

include/ppx/grfx/grfx_command.h
apazylbe: yes, changed

include/ppx/grfx/vk/vk_device.h
```diff
@@ -111,6 +111,11 @@ class Device
 
 extern PFN_vkCmdPushDescriptorSetKHR CmdPushDescriptorSetKHR;
 
+#if defined(VK_KHR_dynamic_rendering)
```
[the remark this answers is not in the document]
apazylbe: These aren't defined with Vulkan headers coming from Android NDK 23.1.7779620 which is currently default. 
I was trying to upgrade to 25.xxx, since that's what we mention in the project README anyway, but there were some problems with that, and I wasn't sure if we want to maintain compatibility with 23.xx and Vulkan headers that don't have this. If we are fine with upgrading to 25.xxx, I can do that in a follow up PR since I need to figure out some issues with the android build related to upgrading, and then remove those `#if`s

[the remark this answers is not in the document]
chaoticbob: Oh, right. I had forgotten that Android doesn't have a stable Vulkan SDK release. My bad.

Keenuts (approved):

PR #360, opened 2023-11-27
Minor refactoring: cleanup Application::Run()

wangra-google:
Minor refactoring: cleanup Application::Run()

### DIFF
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -550,7 +550,8 @@ private:
     bool   IsRunning() const;
 
     // DispatchConfig() allows derived classes to update settings
-    // This function will finalize the settings
+    // This is the final step of updating the settings
+    // based on the values defined in DispatchConfig() and mStandardOpts
     void FinalizeSettings();
 
     // Update the asset directories
@@ -567,8 +568,10 @@ private:
     // List gpus
     void ListGPUs() const;
 
-    // Returns true to exit the loop
-    bool Mainloop();
+    void FrameUpdate();
+    void FrameRender();
+
+    void Mainloop();
 
 #if defined(PPX_BUILD_XR)
     void InitializeXRComponentBeforeGrfxDeviceInit();
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -383,7 +383,6 @@ public:
     virtual void InitKnobs() {}
 
 protected:
-    virtual void DispatchConfig();
     virtual void DispatchSetup();
     virtual void DispatchShutdown();
     virtual void DispatchMove(int32_t x, int32_t y);
@@ -549,10 +548,9 @@ private:
     void   DestroyPlatformWindow();
     bool   IsRunning() const;
 
-    // DispatchConfig() allows derived classes to update settings
-    // This is the final step of updating the settings
-    // based on the values defined in DispatchConfig() and mStandardOpts
-    void FinalizeSettings();
+    // Config() allows derived classes to update settings
+    // Updates the standard application settings to reflect the knob values.
+    void UpdateStandardSettings();
 
     // Update the asset directories
     void UpdateAssetDirs();
@@ -568,10 +566,13 @@ private:
     // List gpus
     void ListGPUs() const;
 
-    void FrameUpdate();
-    void FrameRender();
+    // Process events which could change the Running status of the application
+    void ProcessEvents();
 
-    void Mainloop();
+    // Render the frame, handles both XR and non-XR cases
+    void RenderFrame();
+
+    void MainLoop();
 
 #if defined(PPX_BUILD_XR)
     void InitializeXRComponentBeforeGrfxDeviceInit();
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -535,6 +535,7 @@ public:
 private:
     void   InternalCtor();
     void   InitializeAssetDirs();
+    Result InitializeWindow();
     Result InitializePlatform();
     Result InitializeGrfxDevice();
     Result InitializeGrfxSurface();
@@ -548,6 +549,10 @@ private:
     void   DestroyPlatformWindow();
     bool   IsRunning() const;
 
+    // DispatchConfig() allows derived classes to update settings
+    // This function will finalize the settings
+    void FinalizeSettings();
+
     // Update the asset directories
     void UpdateAssetDirs();
 
@@ -558,6 +563,18 @@ private:
 
     // Initializes standard knobs
     void InitStandardKnobs();
+
+    // List gpus
+    void ListGPUs() const;
+
+    // Returns true to exit the loop
+    bool Mainloop();
+
+#if defined(PPX_BUILD_XR)
+    void InitializeXRComponentBeforeGrfxDeviceInit();
+    void InitializeXRComponentAndUpdateSettingsAfterGrfxDeviceInit();
+    void DestroyXRComponent();
+#endif
 
 private:
     friend struct WindowEvents;
--- a/src/ppx/application.cpp
+++ b/src/ppx/application.cpp
@@ -1142,6 +1142,268 @@ bool Application::IsRunning() const
     return mWindow->IsRunning();
 }
 
+void Application::FinalizeSettings()
+{
+    if (mSettings.appName.empty()) {
+        mSettings.appName = "PPX Application";
+    }
+    if (mSettings.window.title.empty()) {
+        mSettings.window.title = mSettings.appName;
+    }
+
+#if defined(PPX_LINUX_HEADLESS)
+    // Force headless if BigWheels was built without surface support.
+    mSettings.headless = true;
+#else
+    mSettings.headless = mStandardOpts.pHeadless->GetValue();
+#endif
+
+    // If command line argument provided width and height
+    auto       resolution        = mStandardOpts.pResolution->GetValue();
+    const bool hasResolutionFlag = (resolution.first > 0 && resolution.second > 0);
+    if (hasResolutionFlag) {
+        mSettings.window.width  = resolution.first;
+        mSettings.window.height = resolution.second;
+    }
+
+#if defined(PPX_BUILD_XR)
+    resolution = mStandardOpts.pXrUiResolution->GetValue();
+    if (resolution.first > 0 && resolution.second > 0) {
+        mSettings.xr.uiWidth  = resolution.first;
+        mSettings.xr.uiHeight = resolution.second;
+    }
+#endif
+
+    // Disable ImGui in headless or deterministic mode.
+    // ImGUI is not non-deterministic, but the visible informations (stats, timers) are.
+    if ((mSettings.headless || mStandardOpts.pDeterministic->GetValue()) && mSettings.enableImGui) {
+        mSettings.enableImGui = false;
+        PPX_LOG_WARN("Headless or deterministic mode: disabling ImGui");
+    }
+}
+
+#if defined(PPX_BUILD_XR)
+void Application::InitializeXRComponentBeforeGrfxDeviceInit()
+{
+    if (mSettings.xr.enable) {
+        XrComponentCreateInfo createInfo = {};
+        createInfo.api                   = mSettings.grfx.api;
+        createInfo.appName               = mSettings.appName;
+#if defined(PPX_ANDROID)
+        createInfo.androidContext = GetAndroidContext();
+        createInfo.colorFormat    = grfx::FORMAT_R8G8B8A8_SRGB;
+#else
+        createInfo.colorFormat = grfx::FORMAT_B8G8R8A8_SRGB;
+#endif
+        createInfo.depthFormat          = grfx::FORMAT_D32_FLOAT;
+        createInfo.refSpaceType         = XrRefSpace::XR_STAGE;
+        createInfo.viewConfigType       = XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO;
+        createInfo.enableDebug          = mSettings.grfx.enableDebug;
+        createInfo.enableQuadLayer      = mSettings.enableImGui;
+        createInfo.enableDepthSwapchain = mSettings.xr.enableDepthSwapchain;
+        const auto resolution           = mStandardOpts.pResolution->GetValue();
+        const bool hasResolutionFlag    = (resolution.first > 0 && resolution.second > 0);
+        if (hasResolutionFlag) {
+            createInfo.resolution.width  = mSettings.window.width;
+            createInfo.resolution.height = mSettings.window.height;
+        }
+        createInfo.uiResolution.width  = mSettings.xr.uiWidth;
+        createInfo.uiResolution.height = mSettings.xr.uiHeight;
+        createInfo.requiredExtensions  = mStandardOpts.pXrRequiredExtensions->GetValue();
+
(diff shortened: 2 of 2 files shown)


include/ppx/application.h
```diff
@@ -548,6 +549,10 @@ class Application
     void   DestroyPlatformWindow();
     bool   IsRunning() const;
 
+    // DispatchConfig() allows derived classes to update settings
+    // This function will finalize the settings
```
Keenuts, line 553: Could you explain what "finalizing the settings" means in this comment?

src/ppx/application.cpp
```diff
@@ -1273,6 +1213,262 @@ int Application::Run(int argc, char** argv)
...
+                    }
+                }
+            }
+            mXrComponent.EndFrame(mSwapchains, 0, mUISwapchainIndex);
+        }
+    }
+    else
+#endif
+    {
+        mWindow->ProcessEvent();
```
Keenuts, line 1311:
```suggestion
        mWindow->ProcessEvent();
        if (!IsRunning())
          return;
```

src/ppx/application.cpp
```diff
@@ -1352,143 +1537,9 @@ int Application::Run(int argc, char** argv)
...
-                mFirstFrameTime = nowMs / 1000.f;
-            }
-        }
-        // If we reach the maximum number of frames allowed
-        if ((mStandardOpts.pFrameCount->GetValue() > 0 && mFrameCount >= mStandardOpts.pFrameCount->GetValue()) ||
-            (nowMs / 1000.f) > mRunTimeSeconds) {
-            Quit();
+        const bool exitMainLoop = Mainloop();
+        if (exitMainLoop) {
+            break;
```
Keenuts, line 1542:
```suggestion
        Mainloop();
```

include/ppx/application.h
```diff
@@ -559,6 +564,18 @@ class Application
     // Initializes standard knobs
     void InitStandardKnobs();
 
+    // List gpus
+    void ListGPUs() const;
+
+    // Returns true to exit the loop
+    bool Mainloop();
```
Keenuts, line 571: The app already had a `IsRunning()` and `Quit()` method. Seems like this could return void, and the whole logic handled using those (see suggestions).

src/ppx/application.cpp
```diff
@@ -1273,6 +1213,262 @@ int Application::Run(int argc, char** argv)
...
+{
+    // Frame start
+    mFrameStartTime = static_cast<float>(mTimer.MillisSinceStart());
+
+#if defined(PPX_BUILD_XR)
+    if (mSettings.xr.enable) {
+        bool exitRenderLoop = false;
+        mXrComponent.PollEvents(exitRenderLoop);
+        if (exitRenderLoop) {
+            return true;
```
Keenuts, line 1274:
```suggestion
            Quit();
            return;
```

include/ppx/application.h
Keenuts, line 571: nit: this sounds to me like this will loop until it returns true/false.
What do you think about having 2 functions in the main loop, `Update()` and `Render()`, which splits the event/animation from the frame rendering?

Keenuts (commented):
Thanks! A few comments

include/ppx/application.h
wangra-google: thanks for pointing this out, it makes sense. I made the change as suggested.

src/ppx/application.cpp
wangra-google: changed as suggested

src/ppx/application.cpp
wangra-google: changed as suggested

src/ppx/application.cpp
```diff
@@ -1273,6 +1213,282 @@ int Application::Run(int argc, char** argv)
...
+        return false;
+    }
+
+    // Parse args.
+    if (Failed(mCommandLineParser.Parse(argc, const_cast<const char**>(argv)))) {
+        PPX_ASSERT_MSG(false, "Unable to parse command line arguments");
+        return EXIT_FAILURE;
+    }
+
+    // Call config.
```
Keenuts, line 1442: What about removing the DispatchConfig function now?

```cpp
// Call application's Config() method.
Config(mSettings);
```

src/ppx/application.cpp
```diff
@@ -1273,6 +1213,282 @@ int Application::Run(int argc, char** argv)
...
+    // Put this early because it might disable the display.
+    DispatchConfig();
+
+    // Knobs need to be set up before commandline parsing.
+    // This has dependency on DispatchConfig() where standard knob default values are set
+    DispatchInitKnobs();
```
Keenuts, line 1448: Same here?

```cpp
InitStandardKnobs();
// Let the application initialize their own knobs.
InitKnobs();
```

src/ppx/application.cpp
```diff
@@ -1273,6 +1213,282 @@ int Application::Run(int argc, char** argv)
...
+    if (Failed(mCommandLineParser.Parse(argc, const_cast<const char**>(argv)))) {
+        PPX_ASSERT_MSG(false, "Unable to parse command line arguments");
+        return EXIT_FAILURE;
+    }
+
+    // Call config.
+    // Put this early because it might disable the display.
+    DispatchConfig();
+
+    // Knobs need to be set up before commandline parsing.
```
Keenuts, line 1446: This comment is in contradiction with the code above:

```cpp
 if (Failed(mCommandLineParser.Parse(argc, const_cast<const char**>(argv)))) {
```

src/ppx/application.cpp
```diff
@@ -1142,55 +1142,14 @@ bool Application::IsRunning() const
...
-    if (!mKnobManager.IsEmpty()) {
-        auto options = mCommandLineParser.GetOptions();
-        mKnobManager.UpdateFromFlags(options);
-    }
-
-    // Asset directories based on settings in mSettings, mCommandLineParser and mKnobManager
-    // note that mKnobManager needs to be updated by options from mCommandLineParser before this call
-    UpdateAssetDirs();
-
     if (mSettings.appName.empty()) {
```
Keenuts, line 1147: The constructor already initializes this to kDefaultAppName, this can be removed no?

src/ppx/application.cpp
```diff
@@ -1142,55 +1142,14 @@ bool Application::IsRunning() const
...
         mSettings.appName = "PPX Application";
     }
     if (mSettings.window.title.empty()) {
         mSettings.window.title = mSettings.appName;
```
Keenuts, line 1151: Same here, could the default constructor also initialize this to kDefaultAppName and those lines removed?

include/ppx/application.h
```diff
@@ -548,6 +549,11 @@ class Application
     void   DestroyPlatformWindow();
     bool   IsRunning() const;
 
+    // DispatchConfig() allows derived classes to update settings
+    // This is the final step of updating the settings
+    // based on the values defined in DispatchConfig() and mStandardOpts
+    void FinalizeSettings();
```
Keenuts, line 555: I feel like this comment doesn't tell me what this function is useful for.
It seems `FinalizeSettings` is applying the knobs options to the app settings.
Maybe `UpdateStandardSettings()` would be a better name?
And for the comment, what about something like:

```cpp
// Updates the standard application settings to reflect the knob values. 
```

Keenuts (commented):
Thanks! A few improvements left and I feel like this cleanup will be good.

src/ppx/application.cpp
wangra-google: i'd prefer keep them in a function so that knob initializations are in the same func for clarity (also being consistent with other part of the code in this file.)

src/ppx/application.cpp
```diff
@@ -1273,6 +1213,282 @@ int Application::Run(int argc, char** argv)
...
+        }
+        if (!mWindow) {
+            PPX_ASSERT_MSG(false, "out of memory");
+            return ERROR_OUT_OF_MEMORY;
+        }
+    }
+    return SUCCESS;
+}
+
+void Application::FrameUpdate()
```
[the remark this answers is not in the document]
wangra-google: make sense, changed as suggested

src/ppx/application.cpp
```diff
@@ -1273,6 +1213,282 @@ int Application::Run(int argc, char** argv)
...
+            MoveWindow(ci.hwnd, wr.left + 1, wr.top, (wr.right - wr.left), (wr.bottom - wr.top), TRUE);
+            MoveWindow(ci.hwnd, wr.left, wr.top, (wr.right - wr.left), (wr.bottom - wr.top), TRUE);
+
+            mForceInvalidateClientArea = false;
+        }
+#endif
+    }
+}
+
+void Application::Mainloop()
```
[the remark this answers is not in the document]
wangra-google: changed as suggested

src/ppx/application.cpp
```diff
@@ -1273,6 +1213,282 @@ int Application::Run(int argc, char** argv)
...
+        else {
+            mFirstFrameTime = nowMs / 1000.f;
+        }
+    }
+    // If we reach the maximum number of frames allowed
+    if ((mStandardOpts.pFrameCount->GetValue() > 0 && mFrameCount >= mStandardOpts.pFrameCount->GetValue()) ||
+        (nowMs / 1000.f) > mRunTimeSeconds) {
+        Quit();
+    }
+    return;
```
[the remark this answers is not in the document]
wangra-google: removed

include/ppx/application.h
```diff
@@ -559,6 +565,20 @@ class Application
     // Initializes standard knobs
     void InitStandardKnobs();
 
+    // List gpus
+    void ListGPUs() const;
+
+    void FrameUpdate();
```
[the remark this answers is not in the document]
wangra-google: added comments

src/ppx/application.cpp
wangra-google: changed as suggested

src/ppx/application.cpp
wangra-google: oops, nice catch, fixed

src/ppx/application.cpp
wangra-google: yes, updated

src/ppx/application.cpp
wangra-google: updated

include/ppx/application.h
```diff
@@ -559,6 +563,23 @@ class Application
...
+    // List gpus
+    void ListGPUs() const;
+
+    // Process events which could change the Running status of the application
+    void ProcessEvents();
+
+    // Render the frame, handles both XR and non-XR cases
+    void FrameRender();
+
+    void Mainloop();
```
apazylbe, line 575: nit: MainLoop

src/ppx/application.cpp
```diff
@@ -1273,173 +1202,168 @@ int Application::Run(int argc, char** argv)
...
 
-    // Create surface
-    ppxres = InitializeGrfxSurface();
-    if (Failed(ppxres)) {
-        return EXIT_FAILURE;
-    }
+Result Application::InitializeWindow()
+{
+    if (mWindow)
+        return SUCCESS;
```
apazylbe, line 1239: nit: put {} around this statement

src/ppx/application.cpp
```diff
@@ -1273,173 +1202,168 @@ int Application::Run(int argc, char** argv)
...
-    // ---------------------------------------------------------------------------------------------
-
-    // Initialize and start timer
-    ppx::TimerResult tmres = mTimer.Start();
-    if (tmres != ppx::TIMER_RESULT_SUCCESS) {
-        PPX_ASSERT_MSG(false, "failed to start application timer");
-        return EXIT_FAILURE;
+        mWindow->ProcessEvent();
+        if (!IsRunning())
+            return;
```
apazylbe, line 1272: nit: put {} around this statement

src/ppx/application.cpp
```diff
@@ -1273,173 +1202,168 @@ int Application::Run(int argc, char** argv)
...
+        // Frame start
+        mFrameStartTime = static_cast<float>(mTimer.MillisSinceStart());
 
-                mForceInvalidateClientArea = false;
-            }
-#endif
-        }
+        ProcessEvents();
+        if (!IsRunning())
+            return;
```
apazylbe, line 1365: nit: put {} around this

include/ppx/application.h
```diff
@@ -559,6 +563,23 @@ class Application
...
     void InitStandardKnobs();
 
+    // List gpus
+    void ListGPUs() const;
+
+    // Process events which could change the Running status of the application
+    void ProcessEvents();
+
+    // Render the frame, handles both XR and non-XR cases
+    void FrameRender();
```
apazylbe, line 573: nit: can we rename this to RenderFrame()? imo makes it clearer to put verb first, but feel free to ignore if you disagree

src/ppx/application.cpp
```diff
@@ -1491,6 +1415,142 @@ int Application::Run(int argc, char** argv)
...
+        PPX_ASSERT_MSG(false, "Unable to parse command line arguments");
+        return EXIT_FAILURE;
+    }
+
+    // Call config.
+    // Put this early because it might disable the display.
+    Config(mSettings);
+
+    // Knobs need to be set up after commandline parsing.
+    // This has dependency on Config() where standard knob default values are set
```
apazylbe, line 1440: There seems to be a lot of dependencies with config, knobs and command line parser in this section which seems a bit brittle. In particular here, we have to call Config() since it sets defaults, then initialize knobs and then finalize settings since they are updated from the knobs. Have we considered settings defaults for knobs in the InitKnobs instead of in Config()? https://github.com/google/bigwheels/pull/362/files
Just a thought though, thinking of how to make this code clearer

include/ppx/application.h
wangra-google: renamed the function as suggested

src/ppx/application.cpp
wangra-google: fixed

src/ppx/application.cpp
wangra-google: fixed

src/ppx/application.cpp
wangra-google: fixed

include/ppx/application.h
wangra-google: renamed the function as suggested

src/ppx/application.cpp
wangra-google: Agree, it is really error prone like this. 
Thanks for sharing the link of the PR, looks like a better solution. Although I feel that the name `InitKnob` is not quite right: the function is allowing the derived class to have its own knob initializations, plus changing the default values of the standard knobs. Would it be clearer to have another separate virtual function purely for allowing the derived class to overwrite the default value of the standard knobs? Not sure if this is better, would like to hear your opinion on this. thanks :)

apazylbe: Hm, I don't think I have a strong opinion on that one. I can see an argument for either way, so I will be fine with either. I think there is other general weirdness with mSettings and knobs, but we can address that separately including the default options in Config() or somewhere else. Just want to explore that before its use solidified.

apazylbe (approved):

Keenuts (approved):
LGTM, thanks!
For the default settings being brittle, I'd say this can be a follow-up commit.

PR #361, opened 2023-11-27
Fix offscreen format

wangra-google:


hysw:
Ah, another of silent merge conflict :(
In `GraphicsBenchmarkApp::RenderFormat()`, `kFramebufferFormatTypes[pFramebufferFormat->GetIndex()]` should be `pFramebufferFormat->GetValue()` instead.

hysw (approved):

PR #362, opened 2023-11-28
Not for submission, some thought experiments

apazylbe:
- There is a bit of a weird relationship between settings and default values for things that repeat: resolution and XR UI resolution. If the defaults for those are provided then the values in the settings will be overwritten. Might not be a problem, but might not be immediately obvious.

- There is an ordering dependencies between some functions in the beginning of Run(). They seem to be documented, but comments don't seem to be the best way for this, as some comments already might cause confusion. This explores the option of providing defaults in the InitKnobs instead of as part of the settings.

- Removing the storage for the flags in command line parser, this seems to be only needed for `--help`, so it shouldn't be necessary to store that.

### DIFF
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
@@ -41,8 +41,11 @@ const grfx::Api kApi = grfx::API_DX_12_0;
 const grfx::Api kApi = grfx::API_VK_1_1;
 #endif
 
-void GraphicsBenchmarkApp::InitKnobs()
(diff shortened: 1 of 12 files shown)


hysw:
I'll add a bit more to the resolution mess

To obtain the resolution to render:

* `Application::GetWindow{Width,Height}()`,  use `mSettings.window.{width,height}`
* `GetSwapchain()->GetColorImage(i)->GetWidth()` (04_cube), uses image parameter
* `GetSwapchain()->GetWidth()` (22_image_filter), uses swapchain parameter

To get the viewport

* `Application::GetViewport()`, use `mSettings.window.{width,height}`
* `GetSwapchain()->GetRenderPass()->GetGetViewport()` (21_text_draw), uses renderpass parameter

PR #363, opened 2023-11-29
[do not submit, breaking into smaller pieces] Fixes and better usage message for StandardOptions default values

angela28chen:
Fixes:
* For ranged knobs, the max value was erroneously being used as default value
* Default metrics path was not working as described in the help message

New:
* Usage message will now show default values

Changes:
* `--assets-path` renamed to `--extra-assets-path` for clarity
* `--metrics-filename` renamed to `--metrics-path`
* Logic for `--metrics-path` and `--screenshot-path` is made similar for ease of use & helper function `ppx::fs::GetFullPath` is added
* Removed `InitializeAssetDirs()` and shifted what it does into `UpdateAssetDirs()` instead
* Using more concise `InitKnob()` syntax for creating standard options knobs

### DIFF
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -327,23 +327,23 @@ struct ApplicationSettings
     // Default values for standard knobs
     struct StandardKnobsDefaultValue
     {
-        std::vector<std::string> assetsPaths     = {};
-        std::vector<std::string> configJsonPaths = {};
-        bool                     deterministic   = false;
-        bool                     enableMetrics   = false;
-        uint64_t                 frameCount      = UINT64_MAX;
-        int                      gpuIndex        = INT_MAX;
+        const std::vector<std::string> assetsPaths     = {};
+        const std::vector<std::string> configJsonPaths = {};
+        bool                           deterministic   = false;
+        bool                           enableMetrics   = false;
+        uint64_t                       frameCount      = 0;
+        int                            gpuIndex        = 0;
 #if !defined(PPX_LINUX_HEADLESS)
         bool headless = false;
 #endif
         bool                listGpus              = false;
-        std::string         metricsFilename       = std::filesystem::current_path().u8string();
+        std::string         metricsFilename       = "report_@.json";
         bool                overwriteMetricsFile  = false;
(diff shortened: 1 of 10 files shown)


apazylbe:
Thanks for the changes! Could you split this into separate PRs? Even looking at the PR description it looks like it can be logically broken up. Personally, it would make it way easier to review and reduce cognitive load, that way when reviewing we wouldn't have to remember what the intent is, and if there are iterations we wouldn't have to re-read sections that are already good. 
For ex:
PR 1: fixes
PR 2: add defaults to help message
PR 3: changes to assets paths
PR 4: Using more concise InitKnobs syntax
PR 5: metrics path and screenshot path

apazylbe:
I only skimmed the PR so I don't have the full context, but I think we need to be careful with making "external" changes like renaming the flags, and consider the cost vs benefit. There are other places where BigWheels is used beyond this repo, so changing a flag name might mean they break visibly or silently. Perhaps we can consider other options like adding a new flag and "deprecating" the other one if the behaviour is different or living with a non-optimal name. Just some thoughts, probably better discussed on specific cases.

include/ppx/application.h
```diff
@@ -327,23 +327,23 @@ struct ApplicationSettings
...
+        bool                           deterministic   = false;
+        bool                           enableMetrics   = false;
+        uint64_t                       frameCount      = 0;
+        int                            gpuIndex        = 0;
 #if !defined(PPX_LINUX_HEADLESS)
         bool headless = false;
 #endif
         bool                listGpus              = false;
-        std::string         metricsFilename       = std::filesystem::current_path().u8string();
+        std::string         metricsFilename       = "report_@.json";
```
wangra-google, line 340: i have a crash in `std::filesystem::create_directories` with `report_@.json` as metricsFilename on windows

angela28chen:
OK, let me split this PR up and investigate crash

include/ppx/application.h
angela28chen: Can you tell me which dir you were executing this from?

wangra-google: i am running from vs2022, the `Working Directory` is set to `$(ProjectDir)`

PR #368, opened 2023-11-29
Benchmark: Debug View knob

elviscapiaq:


wangra-google (changes requested):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -417,11 +423,14 @@ void GraphicsBenchmarkApp::SetupSphereResources()
 
     // Uniform buffers for draw calls
     {
-        grfx::BufferCreateInfo bufferCreateInfo        = {};
-        bufferCreateInfo.size                          = PPX_MINIMUM_UNIFORM_BUFFER_SIZE;
-        bufferCreateInfo.usageFlags.bits.uniformBuffer = true;
-        bufferCreateInfo.memoryUsage                   = grfx::MEMORY_USAGE_CPU_TO_GPU;
-        PPX_CHECKED_CALL(GetDevice()->CreateBuffer(&bufferCreateInfo, &mSphere.uniformBuffer));
+        mSphere.uniformBuffersForDrawCalls.resize(kMaxSphereInstanceCount);
```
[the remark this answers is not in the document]
elviscapiaq: Thanks for pointing it. Updated to use `push_constants`.

assets/benchmarks/shaders/Benchmark_VsAluBound.hlsl
```diff
@@ -55,6 +55,7 @@ VSOutput vsmain(
   result.normalTS    = mul(Scene.ITModelMatrix, float4(normal, 0)).xyz;
   result.tangentTS   = mul(Scene.ITModelMatrix, float4(tangent, 0)).xyz;
   result.bitangentTS = cross(normal, tangent);
+  result.color       = Scene.Color;
```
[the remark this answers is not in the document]
elviscapiaq: You are right. Updated!

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -1426,6 +1442,26 @@ ppx::grfx::Format GraphicsBenchmarkApp::RenderFormat()
     return renderFormat;
 }
 
+void GraphicsBenchmarkApp::CreateColorsForDrawCalls()
+{
+    // Add the most distinguishable colors for humans.
+    mColorsForDrawCalls.push_back({0.0f, 1.0f, 0.0f});  // Green
```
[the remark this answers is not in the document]
elviscapiaq: Yes, it is easy to distinguish among the 10 colors. But the randomly generated colors also work, so I updated it for simplicity.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -55,6 +55,11 @@ void GraphicsBenchmarkApp::InitKnobs()
     pEnableSpheres->SetDisplayName("Enable Spheres");
     pEnableSpheres->SetFlagDescription("Enable the Spheres in the scene.");
 
+    GetKnobManager().InitKnob(&pDebugView, "debug-view", false);
```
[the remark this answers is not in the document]
wangra-google: ah, i see the latest commit now. thanks

assets/benchmarks/shaders/Benchmark.hlsli
```diff
@@ -20,16 +20,24 @@ struct SceneData
...
 };
 
-ConstantBuffer<SceneData> Scene : register(b0);
-Texture2D                 AlbedoTexture         : register(t1);
-SamplerState              AlbedoSampler         : register(s2);
-Texture2D                 NormalMap             : register(t3);
-SamplerState              NormalMapSampler      : register(s4);
-Texture2D                 MetalRoughness        : register(t5);
-SamplerState              MetalRoughnessSampler : register(s6);
+struct ColorParams
```
wangra-google, line 25: better to add `float unused;` at the end, the compiler will most likely add padding to make sure the element is 16bytes aligned anyways.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -25,13 +25,14 @@ static constexpr size_t SKYBOX_UNIFORM_BUFFER_REGISTER = 0;
...
 static constexpr size_t SKYBOX_SAMPLER_REGISTER        = 2;
 
-static constexpr size_t SPHERE_UNIFORM_BUFFER_REGISTER                = 0;
-static constexpr size_t SPHERE_ALBEDO_SAMPLED_IMAGE_REGISTER          = 1;
-static constexpr size_t SPHERE_ALBEDO_SAMPLER_REGISTER                = 2;
-static constexpr size_t SPHERE_NORMAL_SAMPLED_IMAGE_REGISTER          = 3;
-static constexpr size_t SPHERE_NORMAL_SAMPLER_REGISTER                = 4;
-static constexpr size_t SPHERE_METAL_ROUGHNESS_SAMPLED_IMAGE_REGISTER = 5;
-static constexpr size_t SPHERE_METAL_ROUGHNESS_SAMPLER_REGISTER       = 6;
+static constexpr size_t SPHERE_COLOR_BUFFER_REGISTER                  = 0;
```
wangra-google, line 28: the color buffer is also a uniform. so probably better with `SPHERE_COLOR_UNIFORM_BUFFER_REGISTER`

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -25,13 +25,14 @@ static constexpr size_t SKYBOX_UNIFORM_BUFFER_REGISTER = 0;
...
+static constexpr size_t SPHERE_UNIFORM_BUFFER_REGISTER                = 1;
```
wangra-google, line 29: the be consistent with the naming, probably better with SPHERE_SCENEDATA_UNIFORM_BUFFER_REGISTER

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -623,6 +612,10 @@ void GraphicsBenchmarkApp::SetupSpheresPipelines()
     piCreateInfo.setCount                          = 1;
     piCreateInfo.sets[0].set                       = 0;
     piCreateInfo.sets[0].pLayout                   = mSphere.descriptorSetLayout;
+    piCreateInfo.pushConstants.count               = sizeof(float3) / sizeof(uint32_t);
```
wangra-google, line 615: with the padding, you can change this to sizeof(float4)/sizeof(uint32_t)

wangra-google: maybe better to add a const like kDebugColorPushConstantCount = sizeof(float4)/sizeof(uint32_t) so that you can also use this kDebugColorPushConstantCount later when PushGraphicsConstants is called

wangra-google (changes requested):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -219,9 +225,9 @@ void GraphicsBenchmarkApp::Setup()
     // Descriptor Pool
     {
         grfx::DescriptorPoolCreateInfo createInfo = {};
-        createInfo.sampler                        = 5 * GetNumFramesInFlight(); // 1 for skybox, 3 for spheres, 1 for blit
-        createInfo.sampledImage                   = 6 * GetNumFramesInFlight(); // 1 for skybox, 3 for spheres, 1 for quads, 1 for blit
-        createInfo.uniformBuffer                  = 2 * GetNumFramesInFlight(); // 1 for skybox, 1 for spheres
+        createInfo.sampler                        = 5 * GetNumFramesInFlight();                           // 1 for skybox, 3 for spheres, 1 for blit
+        createInfo.sampledImage                   = 6 * GetNumFramesInFlight();                           // 1 for skybox, 3 for spheres, 1 for quads, 1 for blit
+        createInfo.uniformBuffer                  = 2 * GetNumFramesInFlight() + kMaxSphereInstanceCount; // 1 for skybox, 1 for spheres, `kMaxSphereInstanceCount` for draw calls
```
wangra-google, line 230: this is probably not needed with push constants?

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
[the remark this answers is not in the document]
wangra-google: if the randomly generated colors works fine, maybe code would be cleaner without those predefined colors? I could be wrong, but my feeling is that predefine 10 colors doesn't really change much especially we will have much more drawcalls than 10.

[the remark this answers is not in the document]
angela28chen: I think it's nice to have colours that are visually distinguishable so if there's a way to do that randomly (perhaps with HSV colours, varying the hue) that would be nice.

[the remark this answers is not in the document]
elviscapiaq: I attempted to use HSV colors; however, even with a slight variation in the hue, the colors don't appear distinguishable, even for a few spheres. Therefore, I believe randomly generated colors would be a better option. I have removed the predefined colors to generate all colors randomly.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
elviscapiaq: Done.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
elviscapiaq: Done.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
elviscapiaq: Good idea. Done.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -876,12 +878,13 @@ void GraphicsBenchmarkApp::ProcessKnobs()
...
     const bool depthTestWriteKnobChanged   = pDepthTestWrite->DigestUpdate();
     const bool enableSpheresKnobChanged    = pEnableSpheres->DigestUpdate();
+    const bool debugViewKnobChanged        = pDebugViews->DigestUpdate();
 
     const bool enableSpheres = pEnableSpheres->GetValue();
     // If the LOD is empty, assuming we skipped the setup for all sphere resources at start
     // So need to do the intial setup for all resources here
     const bool spheresAreSetUp               = mSpheresAreSetUp;
-    const bool updateSphereDescriptors       = spheresAreSetUp && allTexturesTo1x1KnobChanged;
+    const bool updateSphereDescriptors       = spheresAreSetUp && (allTexturesTo1x1KnobChanged || debugViewKnobChanged);
```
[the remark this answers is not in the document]
elviscapiaq: That is right. Removed.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
```diff
@@ -61,6 +92,16 @@ enum class SpherePS
     SPHERE_PS_MEM_BOUND
 };
 
+inline const int ToInt(SpherePS vs)
+{
```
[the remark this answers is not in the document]
elviscapiaq: Done.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
```diff
@@ -61,6 +92,16 @@ enum class SpherePS
...
+    switch (vs) {
+        case SpherePS::SPHERE_PS_SIMPLE: return 0;
+        case SpherePS::SPHERE_PS_ALU_BOUND: return 1;
+        case SpherePS::SPHERE_PS_MEM_BOUND: return 2;
+        default: return -1;
+    }
+}
+
 inline const char* ToString(SpherePS vs)
```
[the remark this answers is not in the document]
elviscapiaq: Yes, updated.

wangra-google (approved):

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -221,7 +227,7 @@ void GraphicsBenchmarkApp::Setup()
         grfx::DescriptorPoolCreateInfo createInfo = {};
         createInfo.sampler                        = 5 * GetNumFramesInFlight(); // 1 for skybox, 3 for spheres, 1 for blit
         createInfo.sampledImage                   = 6 * GetNumFramesInFlight(); // 1 for skybox, 3 for spheres, 1 for quads, 1 for blit
-        createInfo.uniformBuffer                  = 2 * GetNumFramesInFlight(); // 1 for skybox, 1 for spheres
+        createInfo.uniformBuffer                  = 2 * GetNumFramesInFlight(); // 1 for skybox, 1 for spheres, `kMaxSphereInstanceCount` for draw calls
```
angela28chen, line 230: nit: unmodify comment since there aren't additional uniform buffers being used

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -1426,6 +1434,15 @@ ppx::grfx::Format GraphicsBenchmarkApp::RenderFormat()
     return renderFormat;
 }
 
+void GraphicsBenchmarkApp::CreateColorsForDrawCalls()
+{
+    // Create colors randomly.
+    mColorsForDrawCalls.resize(kMaxSphereInstanceCount);
+    for (size_t i = 0; i < kMaxSphereInstanceCount; i++) {
+        mColorsForDrawCalls[i] = float4(mRandom.Float(), mRandom.Float(), mRandom.Float(), 0.5f);
```
angela28chen, line 1442: Should this random number generation be connected with the random seed `kSeed` somehow?

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -892,8 +894,14 @@ void GraphicsBenchmarkApp::ProcessKnobs()
         pDrawCallCount->SetValue(pSphereInstanceCount->GetValue());
     }
 
+    // If the debug view is enabled, then change the pixel shader to PsSimple.
```
Keenuts, line 897: Shall the knob be disabled be this is on?
(Instead of ignoring/reseting its value?)

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
Keenuts: The default seed is also fixed in BigWheels, so fine with both.

Keenuts, line 1442: set alpha to 1.0?

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
```diff
@@ -32,9 +33,40 @@ static constexpr uint32_t kMaxSphereInstanceCount  = 3000;
 static constexpr uint32_t kSeed                    = 89977;
 static constexpr uint32_t kMaxFullscreenQuadsCount = 1000;
 
+static constexpr float4   kDefaultDrawCallColor        = float4(1.0f, 0.175f, 0.365f, 0.5f);
```
Keenuts, line 36: why is the alpha to 0.5? (the original shader also had it to 0.5, but unsure if that's by design or an oversight)

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
```diff
@@ -32,9 +33,40 @@ static constexpr uint32_t kMaxSphereInstanceCount  = 3000;
...
+};
+
+inline const int ToInt(DebugView dv)
+{
+    switch (dv) {
+        case DebugView::DISABLED: return 0;
+        case DebugView::SHOW_DRAWCALLS: return 1;
+        default: return -1;
+    }
+}
```
Keenuts, line 55:
```suggestion
enum class DebugView
{
    DISABLED = 0,
    SHOW_DRAWCALLS
};
```

Keenuts: all subsequent elements values are last-element + 1. Then, at usage, `(int)DebugView::SHOW_DRAWCALLS`

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
wangra-google: it is by design so that when the alpha blending is enabled, there is no need to change the shader const. this value is not doing anything if alpha blend is disabled.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
wangra-google: 0.5 is fine, the alpha will be ignored if alpha blend is disabled. it is set to 0.5 so that when alpha blending is enabled, there is no need to change the shader const

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
Keenuts: Right, makes sense, thanks!

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
elviscapiaq: Good catch. Updated.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
elviscapiaq: That is true. Updated.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
elviscapiaq: Looks like a simpler way to do it. Thanks. Updated.

angela28chen (approved):

elviscapiaq:
I rebased this PR and also updated its description.

apazylbe (approved):

PR #370, opened 2023-12-01
Refactor knob creation to use InitKnob

angela28chen:
Preferred for the shorter syntax, and it suits the usual split between knob declaration in header file and initialization in the cpp.


hysw (approved):
LGTM.

Somewhat related, I'd prefer `InitKnobs(KnobManager& knobManager)` than `GetKnobManager()` everywhere, although that probably belong to a different PR.

apazylbe (approved):

Keenuts (approved):

PR #371, opened 2023-12-01
Changes to metrics report and screenshot filenames

angela28chen:
and the behavior of associated flags `--screenshot-path` and `--metrics-filename` to make them more consistent with each other.

### DIFF
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -348,7 +348,7 @@ struct ApplicationSettings
         std::pair<int, int> resolution            = std::make_pair(0, 0);
         uint32_t            runTimeMs             = 0;
         int                 screenshotFrameNumber = -1;
-        std::string         screenshotPath        = "screenshot_frame#.ppm";
+        std::string         screenshotPath        = "screenshot_frame_#.ppm";
         int                 statsFrameWindow      = -1;
         bool                useSoftwareRenderer   = false;
 #if defined(PPX_BUILD_XR)
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -343,12 +343,12 @@ struct ApplicationSettings
         bool headless = false;
 #endif
         bool                listGpus              = false;
-        std::string         metricsFilename       = std::filesystem::current_path().string();
+        std::string         metricsFilename       = "report_@.json";
         bool                overwriteMetricsFile  = false;
         std::pair<int, int> resolution            = std::make_pair(0, 0);
         uint32_t            runTimeMs             = 0;
         int                 screenshotFrameNumber = -1;
-        std::string         screenshotPath        = "";
+        std::string         screenshotPath        = "screenshot_frame#.ppm";
         int                 statsFrameWindow      = -1;
         bool                useSoftwareRenderer   = false;
 #if defined(PPX_BUILD_XR)
--- a/include/ppx/fs.h
+++ b/include/ppx/fs.h
@@ -143,9 +143,12 @@ std::filesystem::path GetInternalDataPath();
 std::filesystem::path GetExternalDataPath();
 #endif
 
+// Returns the default output directory where test artifacts will be written.
+std::filesystem::path GetDefaultOutputDirectory();
+
 // Constructs the full path from a possibly partial path
 // If `regexToReplace` is specified, replace all instances of that symbol in the filename with `replaceString`
-std::filesystem::path GetFullPath(const std::filesystem::path& partialPath, const std::filesystem::path& defaultFolder, const std::filesystem::path& ext, const std::string& regexToReplace = "", const std::string& replaceString = "");
+std::filesystem::path GetFullPath(const std::filesystem::path& partialPath, const std::filesystem::path& defaultFolder, const std::string& regexToReplace = "", const std::string& replaceString = "");
 
 } // namespace ppx::fs
 
--- a/include/ppx/fs.h
+++ b/include/ppx/fs.h
@@ -143,6 +143,10 @@ std::filesystem::path GetInternalDataPath();
 std::filesystem::path GetExternalDataPath();
 #endif
 
+// Constructs the full path from a possibly partial path
+// If `regexToReplace` is specified, replace all instances of that symbol in the filename with `replaceString`
+std::filesystem::path GetFullPath(const std::filesystem::path& partialPath, const std::filesystem::path& defaultFolder, const std::filesystem::path& ext, const std::string& regexToReplace = "", const std::string& replaceString = "");
+
 } // namespace ppx::fs
 
 #endif // ppx_fs_h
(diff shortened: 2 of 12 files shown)


angela28chen:
@wangra-google can you confirm if you still observe a crash in the create directories call when running the `vk_graphics_pipeline` app on windows with VS? I tried this and it ran successfully and produced a file 
```
Metrics report written to path ["C:\\src\\git\\bigwheels\\build\\benchmarks\\graphics_pipeline\\report_1701470606.json"]
```

angela28chen:
Also right now I have it set so that the default behaviour will write the screenshot and the metrics report to the current working folder, but is there a better default for this?

wangra-google:
> @wangra-google can you confirm if you still observe a crash in the create directories call when running the `vk_graphics_pipeline` app on windows with VS? I tried this and it ran successfully and produced a file
> 
> ```
> Metrics report written to path ["C:\\src\\git\\bigwheels\\build\\benchmarks\\graphics_pipeline\\report_1701470606.json"]
> ```

sure, do you want me to try just this PR or with some other changes as well?

wangra-google:
> > @wangra-google can you confirm if you still observe a crash in the create directories call when running the `vk_graphics_pipeline` app on windows with VS? I tried this and it ran successfully and produced a file
> > ```
> > Metrics report written to path ["C:\\src\\git\\bigwheels\\build\\benchmarks\\graphics_pipeline\\report_1701470606.json"]
> > ```
> 
> sure, do you want me to try just this PR or with some other changes as well?

Just tried with your branch, it is working fine! thanks!

src/ppx/fs.cpp
```diff
@@ -190,4 +191,28 @@ std::filesystem::path GetExternalDataPath()
...
+
+    std::filesystem::path fullPath = partialPath;
+
+    if (regexToReplace != "") {
+        std::string fileName    = fullPath.filename().string();
+        std::string newfileName = std::regex_replace(fileName, std::regex(regexToReplace), replaceString);
+        fullPath.replace_filename(std::filesystem::path(newfileName));
+    }
+
+    if (!fullPath.has_parent_path()) {
```
Keenuts, line 207: Should this be checking for root_path instead?
As-is, it prevents prepending any relative path to the basepath.
For desktop, shouldn't change anything, as prepending CWD doesn't change using the path as-is, but it does impact Android: no way to write to the external storage using a relative path.

angela28chen (commented):
Made some changes as suggested. 

Also I have a general question about output files on Android. When I run this code now it will save files to the external data path and show this in the output log:
```
I/PPX: Screenshot of frame 5 saved to: /storage/emulated/0/Android/data/com.google.bigwheels.project_graphics_pipeline.debug/files/screenshot_frame_5.ppm
I/PPX: Metrics report written to path ["/storage/emulated/0/Android/data/com.google.bigwheels.project_graphics_pipeline.debug/files/report_1670460980.json"]
```
How do we check these files? I tried looking on device, can't find it. I also tried looking in Android Studio Device File Explorer, permission denied at the location logged above. Looking at `/data/data/com.google.bigwheels.project_graphics_pipeline.debug/files` which seems to be another way I can access it, I can't see any files in here.

Keenuts:
> I/PPX: Screenshot of frame 5 saved to: /storage/emulated/0/Android/data/com.google.bigwheels.project_graphics_pipeline.debug/files/screenshot_frame_5.ppm


Isn't this folder accessible using `/sdcard/data/com.google...` or something similar?

Keenuts:
> > I/PPX: Screenshot of frame 5 saved to: /storage/emulated/0/Android/data/com.google.bigwheels.project_graphics_pipeline.debug/files/screenshot_frame_5.ppm
> 
> Isn't this folder accessible using `/sdcard/data/com.google...` or something similar?

Yep, working as intended, just accessible at `/sdcard/Android/data/com.google.bigwheels.project_01_triangle.debug/files` from ADB.

include/ppx/fs.h
```diff
@@ -143,6 +143,13 @@ std::filesystem::path GetInternalDataPath();
 std::filesystem::path GetExternalDataPath();
 #endif
 
+// Returns the default output directory where test artifacts will be written.
```
apazylbe, line 146: nit: maybe we should remove the word `test`? I think testing is one use case for outputting files, but I don't think it's the only one

include/ppx/fs.h
```diff
@@ -143,6 +143,13 @@ std::filesystem::path GetInternalDataPath();
...
+std::filesystem::path GetDefaultOutputDirectory();
+
+// Constructs the full path from a possibly partial path
+// If `regexToReplace` is specified, replace all instances of that symbol in the filename with `replaceString`
+std::filesystem::path GetFullPath(const std::filesystem::path& partialPath, const std::filesystem::path& defaultFolder, const std::string& regexToReplace = "", const std::string& replaceString = "");
```
apazylbe, line 151: Since we don't have an existing use case for passing defaultPath as anything other than GetDefaultOutputDirectory(), should we just call that inside the function instead of passing as a parameter?

include/ppx/application.h
```diff
@@ -343,12 +343,12 @@ struct ApplicationSettings
...
 #endif
         bool                listGpus              = false;
-        std::string         metricsFilename       = std::filesystem::current_path().string();
+        std::string         metricsFilename       = "report_@.json";
         bool                overwriteMetricsFile  = false;
         std::pair<int, int> resolution            = std::make_pair(0, 0);
         uint32_t            runTimeMs             = 0;
         int                 screenshotFrameNumber = -1;
-        std::string         screenshotPath        = "";
+        std::string         screenshotPath        = "screenshot_frame_#.ppm";
```
apazylbe, line 351: not necessarily for this PR, but should we add validation on these to make sure there are not empty?

apazylbe (commented):
Sorry, didn't realize I haven't reviewed this. Some small comments, otherwise lgtm

include/ppx/fs.h
angela28chen: Good point, I removed it.

include/ppx/fs.h
angela28chen: I think it makes it easier to unit test if this is separate so I'd prefer to keep it like this.

include/ppx/application.h
angela28chen: I believe passing an empty string for the screenshot path would cause an assert ("Partial path cannot be empty string") in the code when `GetFullPath(mStandardOpts.pScreenshotPath->GetValue()...)` is called in the screenshot code. Example in unit tests: `GetFullPath_PartialPathEmpty`

angela28chen: It wouldn't trigger until the screenshot frame though, I don't think this is checked during flag parsing or anything.

apazylbe (approved):

PR #373, opened 2023-12-05
build: make shaders depends on includes

Keenuts:
Before, a change in an included file did not trigger a shader rebuild. So if bindings were changed in the Benchmark.hlsli headers, the project required a fresh build.


assets/benchmarks/shaders/CMakeLists.txt
```diff
@@ -11,6 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+set(INCLUDE_FILES
+  "${PPX_DIR}/assets/benchmarks/shaders/Benchmark.hlsli"
```
apazylbe, line 15: Do we need to add "VsOutput.hlsli" here as well?

assets/benchmarks/shaders/CMakeLists.txt
```diff
@@ -11,6 +11,10 @@
...
+)
+
 generate_rules_for_shader("shader_benchmarks_compute_filter_3"
     SOURCE "${PPX_DIR}/assets/benchmarks/shaders/ComputeFilter3.hlsl"
     INCLUDES ${INCLUDE_FILES}
```
apazylbe, line 20: Should the include files be added only for the shaders that actually use them, or maybe that's not worth the extra complexity?

elviscapiaq:
I have been facing with this issue. Thanks @Keenuts for addressing it.

assets/benchmarks/shaders/CMakeLists.txt
Keenuts: It's a single line, fixed the dependencies to be the correct for each shader.

PR #375, opened 2023-12-06
Fix invalid timestamp computation for single-pass rendering

elviscapiaq:
The time computation for multi-pass rendering was incorrect. For both `Timestamp Queries` to the GPU, the PIPELINE_STAGE_TOP_OF_PIPE_BIT stage was erroneously used. To correct this, the first timestamp should use the PIPELINE_STAGE_TOP_BIT stage to record the timestamp as soon as all previous commands are processed by the command processor. The second timestamp should employ the PIPELINE_STAGE_BOTTOM_BIT stage to record the timestamp as soon as all previous commands retire.

### DIFF
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
@@ -1508,7 +1508,7 @@ void GraphicsBenchmarkApp::RecordCommandBuffer(PerFrame& frame, const RenderPass
 
     // Write end timestamp
     // Note the framebuffer is still in RENDER_TARGET state, although it should not really be a problem.
-    frame.cmd->WriteTimestamp(frame.timestampQuery, grfx::PIPELINE_STAGE_TOP_OF_PIPE_BIT, /* queryIndex = */ 1);
+    frame.cmd->WriteTimestamp(frame.timestampQuery, grfx::PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT, /* queryIndex = */ 1);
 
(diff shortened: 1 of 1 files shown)


wangra-google:
> The time computation for multi-pass rendering was incorrect. For both `Timestamp Queries` to the GPU, the PIPELINE_STAGE_TOP_OF_PIPE_BIT stage was erroneously used. To correct this, the first timestamp should use the PIPELINE_STAGE_TOP_BIT stage to record the timestamp as soon as all previous commands are processed by the command processor. The second timestamp should employ the PIPELINE_STAGE_BOTTOM_BIT stage to record the timestamp as soon as all previous commands retire.

the fix seems to be a valid change, although i still dont get why the original bug could happen because of this. The time difference between before and after this fix would be just the time duration of the last drawcall which should not cause the big error that Nathan was seeing. This also doesn't quite explain why the bug only happens in multi-pass mode.

wangra-google (approved):

wangra-google:
so the fix makes sense. 
in the multi-pass case, there are some syncs in-between render passes, so all the drawcalls are basically "serialized" without overlapping, whereas in the single-pass case, the gpu time between the 2 "VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT " are just the time duration in CP, not including the gpu execution time.

Keenuts (approved):
Thanks for this! The wording around TOP_OF_PIPE and BOTTOM_OF_PIPE is indeed very weird, but yes, definitely seems like this is the correct fix to make sure the first timestamp if before all the submitted work, and the second after all the previously submitted work. Thanks!

PR #378, opened 2023-12-06
Fix metric calculation for offscreen frame buffer.

hysw:


benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -319,10 +319,15 @@ void GraphicsBenchmarkApp::UpdateMetrics()
...
 
-    const float    gpuWorkDurationInSec = static_cast<float>(mGpuWorkDuration / static_cast<double>(frequency));
-    const uint32_t width                = GetSwapchain()->GetWidth();
-    const uint32_t height               = GetSwapchain()->GetHeight();
-    const uint32_t quadCount            = pFullscreenQuadsCount->GetValue();
+    const float        gpuWorkDurationInSec = static_cast<float>(mGpuWorkDuration / static_cast<double>(frequency));
+    const grfx::Format swapchainColorFormat = GetSwapchain()->GetColorFormat();
+    const uint32_t     swapchainWidth       = GetSwapchain()->GetWidth();
+    const uint32_t     swapchainHeight      = GetSwapchain()->GetHeight();
+    bool               isOffscreen          = pRenderOffscreen->GetValue();
```
wangra-google, line 326: nit: `const` so that it is consistent with other vars

wangra-google (approved):
LGT

Issue #349, opened 2023-11-15
SPV shaders from previous builds in root directory overwrite compiled shaders in a custom directory


apazylbe:
Our Linux Building documentation outlines the process of building from the root directory (https://github.com/google/bigwheels/blob/main/docs/building_and_running.md#linux). If someone does that and then builds using a custom directory (say build), their generated shaders from root directory will be overwriting the generated shaders in the custom directory.
This is fairly niche issue, but can be pretty sneaky. Where if you modify the shaders after, your modifications will be overwritten with previously compiled shaders.

Here's where assets are copied: https://github.com/google/bigwheels/blob/main/CMakeLists.txt#L171

We can try to approach it in a few ways:
1) see if we can filter out the spv files from assets copy
2) see if we can copy assets in the beginning of the build and the shaders are written on top 
3) detect if shaders were already present there and somehow warn

hysw:
Android Studio (gradle?) behave this way, and will interfere with standard cmake build, i.e. it likes spv output to reside in `/assets/*/shaders/spv` instead of `/build/assets/*/shaders/spv`

PR #380, opened 2023-12-08
Bug: Validation errors?

hysw:
Here's a few lines of stuff that I need to add to make {dx,vk}_graphics_pipeline run without throwing out validation error


apazylbe:
Which vulkan sdk version are you using?

hysw:
Looks like it is partially #349 related :(

PR #381, opened 2023-12-11
Fix more type warnings.

hysw:


benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -937,7 +937,8 @@ void GraphicsBenchmarkApp::ProcessKnobs()
             SetupSpheres();
         }
         else {
-            const bool requireMoreSpheres = sphereInstanceCountKnobChanged && (pSphereInstanceCount->GetValue() > mInitializedSpheres);
+            const uint32_t initializedCount   = static_cast<uint32_t>(pSphereInstanceCount->GetValue());
+            const bool     requireMoreSpheres = sphereInstanceCountKnobChanged && (initializedCount > mInitializedSpheres);
```
apazylbe, line 941: I wonder, do we need to check `sphereInstanceCountKnobChanged`?

apazylbe (approved):
One question, although not directly related to this PR, otherwise lgtm

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
hysw: sphereInstanceCountKnobChanged is not the best name for it. It is "user has released the slider and values are final". I'll add this weirdness to the knob design update doc.

apazylbe: ah, cool, thanks

elviscapiaq (approved):

PR #384, opened 2023-12-13
Place android build shaders into build and disallow in-source builds

apazylbe:
One alternative solution to https://github.com/google/bigwheels/issues/349


Keenuts:
I'm fine with this.
I tried playing around with the gradle/cmake to avoid hard-coding the build directory, but there is one step I cannot solve: the gradle appends a "random" hash to the staging directory path. This is outside of the control of gradle, and before cmake...
This absolute path is suboptimal, but I haven't found a better solution either, so I'm fine with this 😊

Keenuts (approved):

hysw:
Can we have it in `build_android` / `android_build`? Local cmake/tools usually defaults to `/build`

hysw (approved):

apazylbe:
> Can we have it in `build_android` / `android_build`? Local cmake/tools usually defaults to `/build`

We can have any path so both build_android, android_build are fine too. The build artifacts are currently being written to `build/android`, and with this change the assets will be written to `build/assets` which will then use the same folder for assets for both android and desktop (provided you use build directory), and I thought it might be nice to have the same assets folder for both? Do you think there might be problems with that? We can definitely switch to Android using build_android for both assets and build outputs

hysw:
> it might be nice to have the same assets folder for both

Android build might pick up different configuration than desktop build. That might cause problem when debugging

p.s. I guess `build/android/assets` should also works fine.

apazylbe:
> > it might be nice to have the same assets folder for both
> 
> Android build might pick up different configuration than desktop build. That might cause problem when debugging
> 
> p.s. I guess `build/android/assets` should also works fine.

Changed to build_android, probably cleaner to use a separate folder

Keenuts (approved):

PR #385, opened 2023-12-13
Add validation to github CI and upgrade vulkan sdk to latest

apazylbe:
default to debug for debug non-android builds
add graphics pipeline to github CI builds since it is under active development.

### DIFF
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -297,7 +297,13 @@ struct ApplicationSettings
     struct
     {
         grfx::Api api               = grfx::API_UNDEFINED;
-        bool      enableDebug       = false;
+
+#if !defined(NDEBUG) and !defined(PPX_ANDROID)
+        bool enableDebug = true;
+#else
+        bool enableDebug = false;
+#endif
+
         uint32_t  numFramesInFlight = 1;
         uint32_t  pacedFrameRate    = 60;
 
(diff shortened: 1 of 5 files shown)


hysw (commented):
LGTM.

Somewhat related: it looks like we always have PPX_BUILD_XR enabled, which make part of the code path unchecked.

wangra-google (approved):

include/ppx/application.h
```diff
@@ -297,7 +297,13 @@ struct ApplicationSettings
     struct
     {
         grfx::Api api               = grfx::API_UNDEFINED;
-        bool      enableDebug       = false;
+
+#if !defined(NDEBUG) and !defined(PPX_ANDROID)
```
Keenuts, line 301: nit: maybe add the comment explaining debug is not enabled by default on Android because validation layers on old version can be a mess to setup.

Keenuts (approved):

include/ppx/application.h
apazylbe: done.

Keenuts (approved):

apazylbe:
> Somewhat related: it looks like we always have PPX_BUILD_XR enabled, which make part of the code path unchecked.

Yeah, we can add that to one or more of the existing CI checks. I can address that separately, and also probably add a flag to enable debug so we don't have to recompile

PR #388, opened 2023-12-19
XR simulated mouse input

hysw:
Simulate a mouse using right controller.

Uses "Khronos Simple Controller Profile", which should be available on most devices.

Controller input is interpreted as follows:

- `/user/hand/right/input/select/click` as left button 
- `/user/hand/right/input/aim/pose` as cursor using ray cast

### DIFF
--- a/include/ppx/imgui_impl.h
+++ b/include/ppx/imgui_impl.h
@@ -55,6 +55,9 @@ public:
     virtual void Shutdown(ppx::Application* pApp) override;
     virtual void Render(grfx::CommandBuffer* pCommandBuffer) override;
     virtual void ProcessEvent() override;
+#if defined(PPX_BUILD_XR)
+    void ProcessXrInput();
+#endif
 
 protected:
     virtual Result InitApiObjects(ppx::Application* pApp) override;
--- a/include/ppx/imgui_impl.h
+++ b/include/ppx/imgui_impl.h
@@ -37,7 +37,7 @@ public:
     virtual void   Shutdown(ppx::Application* pApp) = 0;
     virtual void   NewFrame();
     virtual void   Render(grfx::CommandBuffer* pCommandBuffer) = 0;
-    virtual void   ProcessEvent() {}
+    virtual void   ProcessEvents() {}
 
 protected:
     virtual Result InitApiObjects(ppx::Application* pApp) = 0;
@@ -54,7 +54,7 @@ public:
 
     virtual void Shutdown(ppx::Application* pApp) override;
     virtual void Render(grfx::CommandBuffer* pCommandBuffer) override;
-    virtual void ProcessEvent() override;
+    virtual void ProcessEvents() override;
 #if defined(PPX_BUILD_XR)
     void ProcessXrInput();
 #endif
--- a/include/ppx/imgui_impl.h
+++ b/include/ppx/imgui_impl.h
@@ -37,6 +37,7 @@ public:
     virtual void   Shutdown(ppx::Application* pApp) = 0;
     virtual void   NewFrame();
     virtual void   Render(grfx::CommandBuffer* pCommandBuffer) = 0;
+    virtual void   ProcessEvent() {}
 
 protected:
     virtual Result InitApiObjects(ppx::Application* pApp) = 0;
@@ -53,6 +54,7 @@ public:
 
     virtual void Shutdown(ppx::Application* pApp) override;
     virtual void Render(grfx::CommandBuffer* pCommandBuffer) override;
+    virtual void ProcessEvent() override;
 
 protected:
     virtual Result InitApiObjects(ppx::Application* pApp) override;
@@ -60,6 +62,9 @@ protected:
 
 private:
     grfx::DescriptorPoolPtr mPool;
+#if defined(PPX_BUILD_XR)
+    bool mSimulatedMouseDown = false;
+#endif
 };
 
 #if defined(PPX_D3D12)
--- a/include/ppx/xr_component.h
+++ b/include/ppx/xr_component.h
@@ -177,6 +177,12 @@ public:
     glm::mat4 GetViewMatrixForCurrentView() const;
     XrPosef   GetPoseForCurrentView() const;
 
+    std::optional<XrPosef> GetUIPointingState() const { return mImguiPointingState; }
+    std::optional<bool>    GetUIClickState() const { return mImguiClickState; }
+    // Return cursor location on the UI plane, from center in unit of meters
+    // Note, the current UI swapchain covers a region of [-0.5, +0.5] x [-0.5, +0.5]
+    std::optional<XrVector2f> GetUICursor() const;
+
     bool IsSessionRunning() const { return mIsSessionRunning; }
     bool ShouldRender() const { return mShouldRender; }
 
@@ -226,6 +232,16 @@ private:
     XrDebugUtilsMessengerEXT mDebugUtilMessenger = XR_NULL_HANDLE;
     bool                     mIsSessionRunning   = false;
     bool                     mShouldRender       = false;
+
+    // KHR controller input profile
+    XrActionSet mImguiInput          = XR_NULL_HANDLE;
+    XrSpace     mImguiPointingSpace  = XR_NULL_HANDLE;
+    XrAction    mImguiClickAction    = XR_NULL_HANDLE;
+    XrAction    mImguiPointingAction = XR_NULL_HANDLE;
+    XrTime      mImguiActionTime     = {};
+
+    std::optional<XrPosef> mImguiPointingState = {};
+    std::optional<bool>    mImguiClickState    = {};
 
     std::optional<float> mNearPlaneForFrame     = std::nullopt;
     std::optional<float> mFarPlaneForFrame      = std::nullopt;
--- a/include/ppx/xr_component.h
+++ b/include/ppx/xr_component.h
@@ -129,10 +129,16 @@ class XrComponent
 public:
     virtual ~XrComponent() = default;
 
-    void     InitializeBeforeGrfxDeviceInit(const XrComponentCreateInfo& createInfo);
-    void     InitializeAfterGrfxDeviceInit(const grfx::InstancePtr pGrfxInstance);
-    XrResult InitializeInteractionProfile();
-    void     Destroy();
+    void InitializeBeforeGrfxDeviceInit(const XrComponentCreateInfo& createInfo);
+    void InitializeAfterGrfxDeviceInit(const grfx::InstancePtr pGrfxInstance);
+    void Destroy();
+
+    // Initialize interaction profiles.
+    // Currently supported interaction profile:
+    //  - khr/simple_controller
+    //
+    // The error returned by this function can be safely ignored.
+    XrResult InitializeInteractionProfiles();
 
     void     PollEvents(bool& exitRenderLoop);
     XrResult PollActions();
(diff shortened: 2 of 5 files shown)


src/ppx/imgui_impl.cpp
```diff
@@ -374,6 +374,39 @@ void ImGuiImplVk::Render(grfx::CommandBuffer* pCommandBuffer)
...
+        bool isMouseDown = xrComponent.GetUIClickState().value_or(false);
+        if (isMouseDown != mSimulatedMouseDown) {
+            mSimulatedMouseDown = isMouseDown;
+            io.AddMouseButtonEvent(0, isMouseDown);
+        }
+
+        std::optional<XrVector2f> cursor = xrComponent.GetUICursor();
+        if (cursor.has_value()) {
+            ImGui::SetMouseCursor(ImGuiMouseCursor_Arrow);
+            io.MousePos = ImVec2{
```
wangra-google, line 395: could you add some comments on why there is a (x+1)/2? (i guess it is to bring the range from [-1,1] to [0,1]?)

src/ppx/xr_component.cpp
```diff
@@ -48,6 +48,43 @@ static XrBool32 XrDebugUtilsMessengerCallback(XrDebugUtilsMessageSeverityFlagsEX
...
+{
+    XrVector3f pos = controller.position;
+    XrVector3f fwd = ForwardVector(controller);
+
+    float zDistance = z - pos.z;
+    // Controller is not pointing forward (or very sideway)
+    if (Sign(zDistance) * fwd.z < 0.001) {
+        return std::nullopt;
+    }
+    float scalar = zDistance / fwd.z;
```
wangra-google, line 84: is it possible that fwd.z is 0?

src/ppx/xr_component.cpp
```diff
@@ -48,6 +48,43 @@ static XrBool32 XrDebugUtilsMessengerCallback(XrDebugUtilsMessageSeverityFlagsEX
     return XR_FALSE; /* OpenXR spec suggests always returning false. */
 }
 
+template <int N>
+void CharArrayCopy(char (&dst)[N], const char* src)
+{
+    strncpy(dst, src, N);
+    // Force the last byte to be '\0'
+    dst[N - 1] = '\0';
```
wangra-google, line 56: would it be better to have src as string, and assert `src.length() <= N`?

src/ppx/xr_component.cpp
```diff
@@ -417,6 +504,45 @@ void XrComponent::PollEvents(bool& exitRenderLoop)
...
+        }
+    }
+}
+
+std::optional<XrVector2f> XrComponent::GetUICursor() const
+{
+    if (!mImguiPointingState.has_value()) {
+        return std::nullopt;
+    }
+    return ProjectCursor(mImguiPointingState.value(), -0.5f);
```
wangra-google, line 545: out of curiosity, why is the z -0.5f?

src/ppx/xr_component.cpp
hysw: `Sign(zDistance) * fwd.z == 0` if `fwd.z == 0`

src/ppx/xr_component.cpp
hysw: `assert(strlen() < N)` should also work.
There's not much reason to use assert here, a truncated app name or action name is not fatal (and probably won't happen anyway).

src/ppx/xr_component.cpp
hysw: That's where the layer is, see `ConditionallyPopulateImGuiLayer`. Added a constant for it.

src/ppx/imgui_impl.cpp
hysw: Ah, I were wondering why the movement feels a bit odd, it's because the mapping is wrong :(
It should be a mapping from `[-0.5m, +0.5m]` to `[0, swapchain.width]`,

Added a comment

src/ppx/xr_component.cpp
hysw: Updated to string_view, and added assert.

wangra-google (approved):

src/ppx/imgui_impl.cpp
```diff
@@ -374,6 +374,42 @@ void ImGuiImplVk::Render(grfx::CommandBuffer* pCommandBuffer)
...
+
+        bool isMouseDown = xrComponent.GetUIClickState().value_or(false);
+        if (isMouseDown != mSimulatedMouseDown) {
+            mSimulatedMouseDown = isMouseDown;
+            io.AddMouseButtonEvent(0, isMouseDown);
+        }
+
+        std::optional<XrVector2f> cursor = xrComponent.GetUICursor();
+        if (cursor.has_value()) {
+            // Mappting cursor location from meters to Imgui screen coordinate
```
Keenuts, line 394: nit: mappting -> Mapping

src/ppx/xr_component.cpp
```diff
@@ -48,6 +50,44 @@ static XrBool32 XrDebugUtilsMessengerCallback(XrDebugUtilsMessageSeverityFlagsEX
...
+    dst[length] = '\0';
+}
+
+float Sign(float x)
+{
+    return (x < 0 ? -1 : (x > 0 ? 1 : 0));
+}
+
+XrVector3f ForwardVector(const XrPosef& controller)
+{
```
Keenuts, line 68: Could you reuse GLM's quaternion operations for this?
It seems less error prone to reuse quat rot instead of rewriting it with the known factors like this?

```cpp
glm::vec3 unit_forward(0, 0, -1);
glm::quat rotation(q.x, q.y, q.z, q.w);
auto result = rotation * unit_forward;
return XrVector3f(result.x, result.y, result.z);
```

src/ppx/xr_component.cpp
```diff
@@ -162,7 +202,7 @@ void XrComponent::InitializeBeforeGrfxDeviceInit(const XrComponentCreateInfo& cr
     instanceCreateInfo.next = &instanceCreateInfoAndroid;
 #endif // defined(PPX_ANDROID)
 
-    strncpy(instanceCreateInfo.applicationInfo.applicationName, createInfo.appName.c_str(), sizeof(instanceCreateInfo.applicationInfo.applicationName));
+    CharArrayStrCpy(instanceCreateInfo.applicationInfo.applicationName, createInfo.appName.c_str());
```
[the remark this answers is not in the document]
hysw: `strlcpy` is non-standard :(

src/ppx/imgui_impl.cpp
```diff
@@ -135,9 +135,9 @@ Result ImGuiImpl::Init(ppx::Application* pApp)
 
 void ImGuiImpl::SetColorStyle()
 {
-    //ImGui::StyleColorsClassic();
+    // ImGui::StyleColorsClassic();
```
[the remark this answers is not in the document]
hysw: reverted.
I'd rather have standalone change for clang format related stuff, this is applied by IntelliSense not clang-format-12

src/ppx/xr_component.cpp
hysw: Done, although using explicit `glm::rotate` instead of multiplication, the glm definition of multiplication is a bit at odd with math

src/ppx/imgui_impl.cpp
hysw: Done.

src/ppx/xr_component.cpp
```diff
@@ -48,6 +50,48 @@ static XrBool32 XrDebugUtilsMessengerCallback(XrDebugUtilsMessageSeverityFlagsEX
...
+}
+
+std::optional<XrVector2f> ProjectCursor(XrPosef controller, float z)
+{
+    XrVector3f pos = controller.position;
+    XrVector3f fwd = ForwardVector(controller);
+
+    float zDistance = z - pos.z;
+    // Controller is not pointing forward (or very sideway)
+    if (Sign(zDistance) * fwd.z < 0.001) {
```
wangra-google, line 88: maybe better to add the comment here to explain that it is doing the dot product here dot(normalize(vec3(0,0,z-pos.z)), fwd)?

hysw: Done, it's now an explicit dot product. Might take a few more cpu cycle.

include/ppx/imgui_impl.h
```diff
@@ -37,6 +37,7 @@ class ImGuiImpl
     virtual void   Shutdown(ppx::Application* pApp) = 0;
     virtual void   NewFrame();
     virtual void   Render(grfx::CommandBuffer* pCommandBuffer) = 0;
+    virtual void   ProcessEvent() {}
```
apazylbe, line 40: nit: Should this be `ProcessEvents`?

include/ppx/xr_component.h
```diff
@@ -118,11 +129,13 @@ class XrComponent
 public:
     virtual ~XrComponent() = default;
 
-    void InitializeBeforeGrfxDeviceInit(const XrComponentCreateInfo& createInfo);
-    void InitializeAfterGrfxDeviceInit(const grfx::InstancePtr pGrfxInstance);
-    void Destroy();
+    void     InitializeBeforeGrfxDeviceInit(const XrComponentCreateInfo& createInfo);
+    void     InitializeAfterGrfxDeviceInit(const grfx::InstancePtr pGrfxInstance);
+    XrResult InitializeInteractionProfile();
```
apazylbe, line 134: Could you add some comments for the new functions/variables? 
For example, here mention what you have in the function body: "// Controller support is not essential, so if anything fail, we just return."

apazylbe (approved):
Lgtm, just some nits. Also, would you be able to add some comments in the header files for some of the new functions/variables so that you can understand what they do without reading the code?

apazylbe:
Thank you Peter!

src/ppx/imgui_impl.cpp
```diff
@@ -374,6 +374,51 @@ void ImGuiImplVk::Render(grfx::CommandBuffer* pCommandBuffer)
...
+        return;
+    }
+
+    ImGuiIO&     io          = ImGui::GetIO();
+    XrComponent& xrComponent = pApp->GetXrComponent();
+
+    bool isMouseDown = xrComponent.GetUIClickState().value_or(false);
+    if (isMouseDown != mSimulatedMouseDown) {
+        mSimulatedMouseDown = isMouseDown;
+        io.AddMouseButtonEvent(0, isMouseDown);
```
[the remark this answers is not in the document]
hysw: Apparently `ImGuiMouseButton_Left` exist, although imgui don't use it in their source.

src/ppx/xr_component.cpp
```diff
@@ -258,10 +300,85 @@ void XrComponent::InitializeAfterGrfxDeviceInit(const grfx::InstancePtr pGrfxIns
...
+{
+    // Controller support is not essential, so if anything fail, we just return.
+
+    XrActionSetCreateInfo actionSetInfo{XR_TYPE_ACTION_SET_CREATE_INFO};
+    actionSetInfo.priority = 0;
+    CharArrayStrCpy(actionSetInfo.actionSetName, "imgui");
+    CharArrayStrCpy(actionSetInfo.localizedActionSetName, "ImGui");
+    CHECK_XR_CALL_RETURN_ON_FAIL(xrCreateActionSet(mInstance, &actionSetInfo, &mImguiInput));
+
+    XrActionCreateInfo clickActioninfo{XR_TYPE_ACTION_CREATE_INFO};
```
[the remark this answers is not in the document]
hysw: Done.

include/ppx/xr_component.h
hysw: The function does have a status return. So the comment is at call site.

src/ppx/imgui_impl.cpp
```diff
@@ -374,6 +374,42 @@ void ImGuiImplVk::Render(grfx::CommandBuffer* pCommandBuffer)
     ImGui::Render();
     ImGui_ImplVulkan_RenderDrawData(ImGui::GetDrawData(), grfx::vk::ToApi(pCommandBuffer)->GetVkCommandBuffer());
 }
+
+void ImGuiImplVk::ProcessEvent()
+{
+#if defined(PPX_BUILD_XR)
+    ImGuiIO&     io   = ImGui::GetIO();
+    Application* pApp = Application::Get();
+    if (pApp->GetSettings()->xr.enable) {
```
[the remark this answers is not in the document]
hysw: Changed the structure a bit so this is no longer relevent.

apazylbe (approved):

PR #389, opened 2023-12-20
Use if/else instead of switch

apazylbe:
Previous switch produced an error with certain compilation flags about unhandled enum values. We could either add a default or use if/else, this change uses if/else.


hysw:
imo, it's better to have a compile error than forgetting to deal with the new enum, i.e. add the new enum and put the warning/error into presubmit

angela28chen:
How about: https://github.com/google/bigwheels/pull/390

Keenuts:
Would be in favor of #390 too. (Or this with an else being an error to handle newly added enums)

apazylbe:
Fair enough, I was thinking that this function does not have to have all cases necessarily, but probably better to use switch and make sure new enums are at least thought about.
Closing in favour of https://github.com/google/bigwheels/pull/390

PR #390, opened 2023-12-20
Add all cases in switch statement for graphics pipeline fullscreen quads type

angela28chen:


benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -1696,6 +1696,12 @@ void GraphicsBenchmarkApp::RecordCommandBufferFullscreenQuad(PerFrame& frame, si
             frame.cmd->PushGraphicsConstants(mQuadsPipelineInterfaces[1], 3, &colorValues);
             break;
         }
+        case FullscreenQuadsType::FULLSCREEN_QUADS_TYPE_TEXTURE: {
```
hysw, line 1699: fyi, the reason for adding bracket (i.e. `case ... : { ... }`) to case statement is to limit variable scope, it's not necessary to add it to a break statement. It's also okay to leave it there if you prefer it

Keenuts (approved):
Nothing to add outside of the nit around brackets. Personal preference is no brackets if not required, but do as you prefer.

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
angela28chen: Thanks, I didn't know this! I took it out

angela28chen:
ready for merging

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -1696,6 +1696,11 @@ void GraphicsBenchmarkApp::RecordCommandBufferFullscreenQuad(PerFrame& frame, si
             frame.cmd->PushGraphicsConstants(mQuadsPipelineInterfaces[1], 3, &colorValues);
             break;
         }
+        case FullscreenQuadsType::FULLSCREEN_QUADS_TYPE_TEXTURE:
+            break;
+        default: {
+            PPX_ASSERT_MSG(true, "unsupported FullscreenQuadsType: " << static_cast<int>(pFullscreenQuadsType->GetValue()));
```
apazylbe, line 1702: Could you add a `break` in `default`? and I don't think we need `{}` there either.
The break wouldn't matter now, but if we ever added a case in the end or move the default to top we might forget to add break then

angela28chen: OK, done!

hysw (approved):

apazylbe (approved):

PR #393, opened 2024-01-03
Proof of Concept: XR Camera Movement

hysw:
Move camera using controller.

This is a proof of concept implementation
- controller direction for movement direction
- use the select button for moving (which conflict with the current ImGui integration)
- fixed 1m/s movment speed

I'd like to add joystick input, however that's not a part of khr/simple_controller. It will need to add device specific config, e.g. `/interaction_profiles/oculus/touch_controller`, which is not addressed here.

### DIFF
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
@@ -594,6 +594,8 @@ private:
     // UTILITY
     // =====================================================================
 
+    const Camera& GetCamera();
+
     // Loads shader at shaderBaseDir/fileName and creates it at ppShaderModule
     void SetupShader(const std::filesystem::path& fileName, grfx::ShaderModule** ppShaderModule);
     void SetupShader(const char* baseDir, const std::filesystem::path& fileName, grfx::ShaderModule** ppShaderModule);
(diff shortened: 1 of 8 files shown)


benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
```diff
@@ -594,6 +594,8 @@ class GraphicsBenchmarkApp
     // UTILITY
     // =====================================================================
 
+    const Camera& GetCamera();
```
wangra-google, line 597: const Camera& GetCamera() const;?

projects/fishtornado_xr/FishTornado.h
```diff
@@ -51,8 +51,9 @@ class FishTornadoApp
     static FishTornadoApp* GetThisApp();
     float                  GetTime() const { return mTime; }
     float                  GetDt() const { return mDt; }
-    const PerspCamera*     GetCamera() const { return &mCamera; }
-    const Shark*           GetShark() const { return &mShark; }
+
+    const Camera* GetCamera() const;
```
wangra-google, line 55: not related to your change, but wondering why pointers are returned instead of const ref? :p

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
hysw: Updated, in #391 since the commit was from that PR.

projects/fishtornado_xr/FishTornado.h
hysw: I'm not sure. That was from the initial version of fish tornado from long ago.

PR #396, opened 2024-01-08
Re-factor compute and rendering support

### LINKED ISSUE #26: Fluid simulation - Use resource pool for compute shaders
From https://github.com/google/bigwheels/pull/20#discussion_r1041170266

dnovillo:
This simplifies the logic handling scheduling and dispatch of compute shaders and textures.  

Main changes:

- Class Texture renamed to SimulationGrid for clarity. The supporting textures are created from Bitmaps. Each grid has its own descriptor set and vertex buffer.

- Simplified class hierarchy for shaders.  This removes the GraphicsShader class and moves the rendering code to SimulationGrid::Draw().

- Removed the concept of dispatch records. Instead, each compute shader has a Dispatch() method that builds up (or re-uses) the descriptor set with all the arguments and queues into the frame's command buffer.  This fixes #26

- Removed structs `ComputeResources` and `GraphicsResources`.  Moved members closer to where they're needed inside `FluidSimulationApp` or `SimulationGrid`.

- Renamed `FluidSimulation` to `FluidSimulationApp` and made it inherit from `ppx::Application`.  Moved all the application code to `sim.cpp`. This removes the need to have a simulation object.

- Compute dispatches are all grouped together before the render pass.

- Data required to be used in each dispatch operation is grouped inside the PerFrame structure. This is implemented as a pool that contains descriptors and uniform buffers to use for compute dispatch operations. On every frame, every entry in the pool is marked available. When scheduling a new compute operation, if there are not enough entries in the pool, a new one is added (in PerFrame::GetDispatchData).

- Removed unused code to generate and display a checkerboard grid (remnant from original port).

A subsequent cleanup will simplify the compute shader hierarchy.  It's now largely unnecessary, but this change was becoming too large already.


projects/fluid_simulation/sim.h
```diff
@@ -62,208 +85,128 @@ struct Bouncer
...
+
+    // Grids used for filtering.
+    std::vector<std::unique_ptr<SimulationGrid>> mBloomGrids;
+    std::unique_ptr<SimulationGrid>              mBloomGrid      = nullptr;
+    std::unique_ptr<SimulationGrid>              mCurlGrid       = nullptr;
+    std::unique_ptr<SimulationGrid>              mDisplayGrid    = nullptr;
+    std::unique_ptr<SimulationGrid>              mDitheringGrid  = nullptr;
+    std::unique_ptr<SimulationGrid>              mDivergenceGrid = nullptr;
+    std::unique_ptr<SimulationGrid>              mDrawColorGrid  = nullptr;
+    std::unique_ptr<SimulationGrid>              mDyeGrid[2];
```
chaoticbob, line 150: Might want to replace the 2 with a named constant of your choice in case someone else is reading this in the future knows what it means.

projects/fluid_simulation/sim.h
```diff
@@ -62,208 +85,128 @@ struct Bouncer
...
-    // also clears the execution schedule.
-    //
-    // TODO(https://github.com/google/bigwheels/issues/26): Implement resource pool.
-    void FreeComputeShaderResources();
+    void GenerateInitialSplat(PerFrame* frame);
 
-    // brief Execute all the scheduled graphics shaders in sequence.
-    void DispatchGraphicsShaders(const PerFrame& frame);
+    // Update the state of the simulation for the given frame.
+    void Update(PerFrame* frame);
```
chaoticbob, line 115: nit: should be `*pFrame*

projects/fluid_simulation/shaders.h
```diff
@@ -108,7 +104,7 @@ class Texture
 // This must match the CSInputs structure in assets/fluid_simulation/shaders/config.hlsli.
 struct alignas(16) ScalarInput
 {
-    ScalarInput(Texture* output)
+    ScalarInput(SimulationGrid* output)
```
chaoticbob, line 107: nit: should be `pOutput`

chaoticbob (approved):
Just a few minor nits. The big changes make it hard to the impact of overall changes to the program so I'll have to wait and see some of these in context.

dnovillo:
> Just a few minor nits. The big changes make it hard to the impact of overall changes to the program so I'll have to wait and see some of these in context.

Yeah, I tried hard to keep it small but removing the concept of dispatch records had a string of side effects that kept accumulating.  Once all the refactoring changes are in, I'll gladly fix any other inconsistencies in follow-up PRs.

projects/fluid_simulation/shaders.h
dnovillo: Done.  I left some un modified because they are disappearing in the next PR.

projects/fluid_simulation/sim.h
dnovillo: Done.

projects/fluid_simulation/sim.h
dnovillo: Done.

PR #398, opened 2024-01-10
Make fluid simulation window re-sizable

### LINKED ISSUE #229: [fluid simulation] Simulator should respond to swapchain resizing events
It's currently not possible to change the size of the window running the simulation.

dnovillo:
**NOTE TO REVIEWERS: This is a very simple change but relies on the new code structure from #323, #396 and #397.  Please review those first. Thanks.**

This adds a call back for `Resize` so that the simulation grids can be re-allocated to match the new window size.

Fixes #229.

### DIFF
--- a/projects/fluid_simulation/sim.cpp
+++ b/projects/fluid_simulation/sim.cpp
@@ -158,6 +158,7 @@ void FluidSimulationApp::Config(ppx::ApplicationSettings& settings)
     settings.grfx.api              = kApi;
     settings.grfx.enableDebug      = clOptions.HasExtraOption("enable-debug");
     settings.allowThirdPartyAssets = true;
+    settings.window.resizable      = true;
 }
 
 void FluidSimulationApp::Setup()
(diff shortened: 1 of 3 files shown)


projects/fluid_simulation/sim.cpp
```diff
@@ -309,6 +310,11 @@ void FluidSimulationApp::SetupRenderingPipeline()
     PPX_CHECKED_CALL(GetDevice()->CreateGraphicsPipeline(&gpci, &mGraphicsPipeline));
 }
 
+void FluidSimulationApp::Resize(uint32_t width, uint32_t height)
+{
+    SetupGrids();
```
chaoticbob, line 315: I assume this destroys and then recreates?

dnovillo: Yes.  They're all unique pointers and std containers that `SetupGrids` empties.

dnovillo:
@chaoticbob I've rebased the PR so it should be now simpler to review.  PTAL. Thanks.

apazylbe (approved):

PR #403, opened 2024-01-15
Add new knob to display CFD grids

dnovillo:
This adds a new knob to enable/disable the rendering of the intermediate grids using during simulation.  This used to be enabled by `--enable-debug`, but this makes it easier to split API verification from showing these grids.


dnovillo:
@apazylbe your fix to the build system helped.  Thanks!  PTAL.

apazylbe (approved):

Issue #97, opened 2023-04-13
CMake rules create duplicate implicit targets on Windows


vettoreldaniele:
There is a problem on how we generate CMake rules for shader targets, specifically on Windows.

We use `MAIN_DEPENDENCY` to specify the input shader file (`<shader_name>.hlsl`) when we create custom commands to compiler shaders with DXC (`add_custom_command`). We use `MAIN_DEPENDENCY` instead of `DEPENDS` because the former allows Visual Studio to avoid trying to compile the HLSL file with its built-in compiler (FXC). However, due to Visual Studio limitations, it is not possible to have multiple `add_custom_command` rules with the same `MAIN_DEPENDENCY` file. We do that when we compile the same shader for multiple stages. This does work in the sense that CMake generation succeeds, but it creates implicit multiple targets that cause Visual Studio to try to compile the same file multiple times in parallel causing sporadic build failures (in case the rules run in parallel).

To reproduce, try generating a VS solution and building a shader target with multiple stages.

I will add more specific information soon.

vettoreldaniele:
Reproduce:
1) Generate a VS solution with CMake. To keep it simple, only enable Vulkan (`-DPPX_D3D12=OFF -DPPX_VULKAN=ON`).
1) Open the solution (`BigWheels.sln`).
1) Pick a shader target. I chose `shader_skybox` (under the `shaders` folder in the Solution View).
![image](https://user-images.githubusercontent.com/3148446/232060877-4e7b1b85-2ab1-4c9b-98ae-80d7398fea3d.png)
1) Right click the vertex shader target (`shader_skybox_vs`) and choose `Rebuild` (we're using Rebuild instead of Build to force it to clean and build from scratch so that it's reproducible). You will have the following output showing it compiled the vertex shader:
```
Rebuild started...
1>------ Rebuild All started: Project: vk_shader_skybox_vs, Configuration: Debug x64 ------
1>------ Compiling VS Shader [SPV_6_6] ------
1>[SPV_6_6] Compiling VS C:/src/bigwheels_fork/assets/basic/shaders/SkyBox.hlsl to C:/src/bigwheels_fork/build-windows/assets/basic/shaders/spv/SkyBox.vs.spv
1>Building Custom Rule C:/src/bigwheels_fork/assets/basic/shaders/CMakeLists.txt
========== Rebuild All: 1 succeeded, 0 failed, 0 skipped ==========
```
[...]

PR #404, opened 2024-01-15
Add the same workaround for PPX_BUILD_XR=0

apazylbe:
as the one for PPX_BUILD_XR=1 mentioned in
to work around https://github.com/google/bigwheels/issues/97


dnovillo (approved):
Thanks for the quick fix!

PR #405, opened 2024-01-16
LiveMetric: realtime statistics for debug UI.

hysw:
This PR is a fork of #377 .

Adding real time statistics into the metric framework so it is easier to read.

### DIFF
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
@@ -306,9 +306,9 @@ void GraphicsBenchmarkApp::SetupMetrics()
     mMetricsData.metrics[MetricsData::kTypeBandwidth]         = AllocateMetricID();
     mMetricsData.metrics[MetricsData::kTypeGPUWorkDuration]   = AllocateMetricID();
 
-    AddLiveMetric(mMetricsData.metrics[MetricsData::kTypeCPUSubmissionTime]);
-    AddLiveMetric(mMetricsData.metrics[MetricsData::kTypeBandwidth]);
-    AddLiveMetric(mMetricsData.metrics[MetricsData::kTypeGPUWorkDuration]);
+    BindLiveMetric(mMetricsData.metrics[MetricsData::kTypeCPUSubmissionTime]);
+    BindLiveMetric(mMetricsData.metrics[MetricsData::kTypeBandwidth]);
+    BindLiveMetric(mMetricsData.metrics[MetricsData::kTypeGPUWorkDuration]);
 }
 
 void GraphicsBenchmarkApp::SetupMetricsRun()
@@ -321,18 +321,18 @@ void GraphicsBenchmarkApp::SetupMetricsRun()
             "ms",
             ppx::metrics::MetricInterpretation::LOWER_IS_BETTER,
             {0.f, 10000.f}};
-        ppx::metrics::MetricID res = AddMetric(metadata, mMetricsData.metrics[MetricsData::kTypeCPUSubmissionTime]);
-        PPX_ASSERT_MSG(res != ppx::metrics::kInvalidMetricID, "Failed to add CPU Submission Time metric");
+        bool bindres = BindMetric(mMetricsData.metrics[MetricsData::kTypeCPUSubmissionTime], metadata);
+        PPX_ASSERT_MSG(bindres, "Failed to bind CPU Submission Time metric");
     }
     {
         ppx::metrics::MetricMetadata metadata = {
             ppx::metrics::MetricType::GAUGE,
             "Bandwidth",
-            "GB/s",
+            "GiB/s",
             ppx::metrics::MetricInterpretation::HIGHER_IS_BETTER,
             {0.f, 10000.f}};
-        ppx::metrics::MetricID res = AddMetric(metadata, mMetricsData.metrics[MetricsData::kTypeBandwidth]);
-        PPX_ASSERT_MSG(res != ppx::metrics::kInvalidMetricID, "Failed to add Bandwidth metric");
+        bool bindres = BindMetric(mMetricsData.metrics[MetricsData::kTypeBandwidth], metadata);
+        PPX_ASSERT_MSG(bindres, "Failed to bind Bandwidth metric");
     }
 }
 
@@ -1333,7 +1333,7 @@ void GraphicsBenchmarkApp::DrawExtraInfo()
         ImGui::Text("Write Bandwidth");
         ImGui::NextColumn();
         ImGui::Text(
-            "%.2f GB/s min=%.2f max=%.2f\n%.2f%s%.2f GB/s",
+            "%.2f GiB/s min=%.2f max=%.2f\n%.2f%s%.2f GiB/s",
             bandwidthLive.Latest(),
             bandwidthLive.Min(),
             bandwidthLive.Max(),
@@ -1346,17 +1346,17 @@ void GraphicsBenchmarkApp::DrawExtraInfo()
             const auto bandwidth = GetGaugeBasicStatistics(mMetricsData.metrics[MetricsData::kTypeBandwidth]);
             ImGui::Text("Average Write Bandwidth");
             ImGui::NextColumn();
-            ImGui::Text("%.2f GB/s", bandwidth.average);
+            ImGui::Text("%.2f GiB/s", bandwidth.average);
             ImGui::NextColumn();
 
             ImGui::Text("Min Write Bandwidth");
             ImGui::NextColumn();
-            ImGui::Text("%.2f GB/s", bandwidth.min);
+            ImGui::Text("%.2f GiB/s", bandwidth.min);
             ImGui::NextColumn();
 
             ImGui::Text("Max Write Bandwidth");
             ImGui::NextColumn();
-            ImGui::Text("%.2f GB/s", bandwidth.max);
+            ImGui::Text("%.2f GiB/s", bandwidth.max);
             ImGui::NextColumn();
         }
     }
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
@@ -15,6 +15,7 @@
 #include "GraphicsBenchmarkApp.h"
 #include "SphereMesh.h"
 
+#include "ppx/application.h"
 #include "ppx/graphics_util.h"
 #include "ppx/grfx/grfx_format.h"
 #include "ppx/timer.h"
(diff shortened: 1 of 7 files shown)


mdagois:
Could you please add a quick summary of what problem the PR solves?
I am not extremely familiar with the benchmark, and I have not touched BigWheels for a few months.

hysw:
> Could you please add a quick summary of what problem the PR solves?
> I am not extremely familiar with the benchmark, and I have not touched BigWheels for a few months.

This gives a readable statistics for cpu submission time / gpu work duration / etc. i.e. the leading digit does not jump around every frame. This version is due to suggestion of @Keenuts 's suggestion on have it in the metric framework. The stand alone version is #377 .

src/ppx/metrics.cpp
```diff
@@ -265,20 +342,39 @@ MetricID Manager::AddMetric(const MetricMetadata& metadata)
...
+    bool hasLiveMetric  = (liveMetricIter != mLiveMetrics.end());
+    if (hasLiveMetric) {
+        liveMetricIter->second.RecordEntry(data);
+    }
     if (mActiveRun == nullptr) {
-        PPX_LOG_WARN("Attempted to record a metric entry with no active run.");
+        if (!hasLiveMetric) {
+            PPX_LOG_WARN("Attempted to record a metric entry with no active run.");
+        }
         return false;
```
Keenuts, line 371: Seems that the function returns true if the metric data was recorded. If the live metric exits, but there is no active run, shouldn't the function also return true since the live metric got recorded?

src/ppx/metrics.cpp
```diff
@@ -255,7 +324,15 @@ bool Manager::HasActiveRun() const
     return (mActiveRun != nullptr);
 }
 
-MetricID Manager::AddMetric(const MetricMetadata& metadata)
+MetricID Manager::EnsureAllocateID(MetricID reuseID)
```
Keenuts, line 327: This doesn't ensure the ID is allocated, it only ensures the ID is not kInvalidMetricID.
- doesn't prevent collisions
- doesn't actually check if the number is allocated, or even the possible range `[0; mNextMetricID[`

include/ppx/metrics.h
```diff
@@ -285,9 +386,24 @@ class Manager final
...
+    // Allocate a MetricID.
+    MetricID AllocateID() { return EnsureAllocateID(kInvalidMetricID); }
+
     // Adds a metric to the current run. A run must be started to add a metric.
     // Failure to add a metric returns kInvalidMetricID.
-    MetricID AddMetric(const MetricMetadata& metadata);
+    // Optionally bind the metric to an existing metricID.
+    MetricID AddMetric(
+        const MetricMetadata& metadata,
+        MetricID              metricID = kInvalidMetricID);
```
Keenuts, line 397: What about `std::optional<MetricID> metricId = std::nullopt`?

include/ppx/metrics.h
```diff
@@ -285,9 +386,24 @@ class Manager final
     // Returns whether a run is active.
     bool HasActiveRun() const;
 
+    // Allocate a MetricID.
+    MetricID AllocateID() { return EnsureAllocateID(kInvalidMetricID); }
```
Keenuts, line 390: Is that needed?
Could this be removed, and only `AddMetric` and `AddLiveMetric` used instead?
To bind 2 metrics, the user would call one function first, and used the returned ID to create the second one.
This way, there is no way for an ID to bound to no metrics.

include/ppx/metrics.h
```diff
@@ -285,9 +386,24 @@ class Manager final
...
+        MetricID              metricID = kInvalidMetricID);
+
+    // Adds a live metric.
+    // Optionally bind the metric to an existing metricID.
+    MetricID AddLiveMetric(
+        double   halfLife = LiveMetric::kDefaultHalfLife,
+        MetricID metricID = kInvalidMetricID);
+
+    // Adds a live metric to current run.
+    void BindMetric(MetricID liveMetricID, const MetricMetadata& metadata);
```
Keenuts, line 406: Is this defined somewhere?

include/ppx/metrics.h
```diff
@@ -299,6 +415,12 @@ class Manager final
     // Get Gauge Basic Statistics, only works for type GAUGE
     GaugeBasicStatistics GetGaugeBasicStatistics(MetricID id) const;
 
+    // Get Live Statistics for a LiveMetric, only works for type GAUGE
+    LiveStatistics GetLiveStatistics(MetricID id) const;
```
Keenuts, line 419: Shall this be renamed to `GetGaugeLiveStatistics` to match the function above?

Keenuts (changes requested):
Thanks for putting this together. A few comments 😊

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -1239,6 +1329,19 @@ void GraphicsBenchmarkApp::DrawExtraInfo()
         ImGui::Text("%.2f GB", dataWriteInGb);
         ImGui::NextColumn();
 
+        const auto bandwidthLive = GetLiveStatistics(mMetricsData.metrics[MetricsData::kTypeBandwidth]);
+        ImGui::Text("Write Bandwidth");
+        ImGui::NextColumn();
+        ImGui::Text(
+            "%.2f GB/s min=%.2f max=%.2f\n%.2f%s%.2f GB/s",
```
Keenuts, line 1336: super super nit: `GiB`? (Not even sure, the wikipedia article is confusing as GB seems to be either 1000^3 or 1024^3 depending on context)

include/ppx/metrics.h
hysw: My bad, I should have be more careful removing temporary code.

include/ppx/metrics.h
hysw: `std::optional<MetricID>` have two null value, i.e. `std::nullopt` and `kInvalidMetricID`.
In which case I'd rather have `MetricID AddMetric(MetricMetadata)` and `bool BindMetric(MetricID,MetricMetadata)`

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
hysw: In memory bandwidth context, `GB` means $2^{30}$. The places that use $10^9$ are network transfer and storage drives, because ... marketing...
Yes I guess I'll change this and the other existing parts to `GiB` to be extra clear

mdagois:
> > Could you please add a quick summary of what problem the PR solves?
> > I am not extremely familiar with the benchmark, and I have not touched BigWheels for a few months.
> 
> This gives a readable statistics for cpu submission time / gpu work duration / etc. i.e. the leading digit does not jump around every frame. This version is due to suggestion of @Keenuts 's suggestion on have it in the metric framework. The stand alone version is #377 .

It is a kind of tracker of statistics for an underlying gauge metrics, and you reset it (by clearing history) whenever you want. Is that correct?

Keenuts:
> > > Could you please add a quick summary of what problem the PR solves?
> > > I am not extremely familiar with the benchmark, and I have not touched BigWheels for a few months.
> > 
> > 
> > This gives a readable statistics for cpu submission time / gpu work duration / etc. i.e. the leading digit does not jump around every frame. This version is due to suggestion of @Keenuts 's suggestion on have it in the metric framework. The stand alone version is #377 .
> 
> It is a kind of tracker of statistics for an underlying gauge metrics, and you reset it (by clearing history) whenever you want. Is that correct?

The initial idea is to have a layer on top of the raw metrics that averages it over X samples.
Precision is not important since it's only used for UI display. What's more important is to be able to eyeball the metrics quickly (hence the window average).
Because this is for quick eyeballing investigation, it's a "nice to have" to be able to disable recording.

That's for the goal/idea I have in mind.

include/ppx/metrics.h
Keenuts: Yeah.. I agree wince we have 2 null values, optional is not a good idea. Don't like those magic numbers for null when we have alternatives, but not for this PR I suppose.
Let's disregard my comment!

include/ppx/application.h
```diff
@@ -523,13 +524,28 @@ class Application
...
+
     // Adds a metric to the current run. If no run is active, returns metrics::kInvalidMetricID.
     // See StartMetricsRun for why this wrapper is necessary.
-    virtual metrics::MetricID AddMetric(const metrics::MetricMetadata& metadata);
+    metrics::MetricID AddMetric(const metrics::MetricMetadata& metadata);
+
+    // Bind a metric to the current run, Return false if no run is active.
+    bool BindMetric(metrics::MetricID metricID, const metrics::MetricMetadata& metadata);
+
+    // Add a live metric, the returned MetricID can also be used for recorded metric.
```
Keenuts, line 537: the returned metricID? outdated comment?

src/test/metrics_test.cpp
```diff
@@ -522,4 +546,54 @@ TEST_F(MetricsTestFixture, MetricsGaugeIgnoresSameSeconds)
     EXPECT_EQ(gauge["time_series"][1][1], 11.0);
 }
 
+TEST_F(MetricsTestFixture, MetricsLiveStatistics)
```
Keenuts, line 549: Shall the following tests be added:
- bind live + normal metrics, start/stop run, check binding state (is live metric still recording after the run?)
- bind live metric and normal metric, record data, clear history, check reported date on both, validating clear only clears the live data.

Keenuts (approved):
Thanks for the changes! Added comments, but LGTM for the direction. Addind second owner.

include/ppx/metrics.h
hysw: Correct myself, it used to be that you need to re-AddMetric, which creates new metricID per run.

hysw:
> It is a kind of tracker of statistics for an underlying gauge metrics, and you reset it (by clearing history) whenever you want. Is that correct?

Addition to Nathan's comment. The use case is to tracks a set of gauge metrics that we want to see on screen, not necessary the ones we want to record to a profiling run (e.g. cpu time spent on input processing). The samples are $2^{-t_{elapsed}}$ weighted to make sure we are showing recent information (also this formulation allows constant time update). Please note this is not the last X samples since interval between sample can vary from $<0.1ms$ to $100^+ ms$ . The clearing of history is somewhat unnecessary, since old data's weight will decay to almost zero in a few seconds.

include/ppx/metrics.h
hysw: It does not report `MetricGauge`'s live statistics, so I'll leave Gauge out of the name.

hysw:
Adding Angela & Ran since they have reviewed #377 .

mdagois:
> > > > Could you please add a quick summary of what problem the PR solves?
> > > > I am not extremely familiar with the benchmark, and I have not touched BigWheels for a few months.
> > > 
> > > 
> > > This gives a readable statistics for cpu submission time / gpu work duration / etc. i.e. the leading digit does not jump around every frame. This version is due to suggestion of @Keenuts 's suggestion on have it in the metric framework. The stand alone version is #377 .
> > 
> > 
> > It is a kind of tracker of statistics for an underlying gauge metrics, and you reset it (by clearing history) whenever you want. Is that correct?
> 
> The initial idea is to have a layer on top of the raw metrics that averages it over X samples. Precision is not important since it's only used for UI display. What's more important is to be able to eyeball the metrics quickly (hence the window average). Because this is for quick eyeballing investigation, it's a "nice to have" to be able to disable recording.
> 
> That's for the goal/idea I have in mind.

Ok, thx. Makes sense.

mdagois:
LGTM from a metrics perspective. I'll defer approval to people that own the project.

include/ppx/application.h
```diff
@@ -523,13 +524,28 @@ class Application
...
     // Adds a metric to the current run. If no run is active, returns metrics::kInvalidMetricID.
     // See StartMetricsRun for why this wrapper is necessary.
-    virtual metrics::MetricID AddMetric(const metrics::MetricMetadata& metadata);
+    metrics::MetricID AddMetric(const metrics::MetricMetadata& metadata);
+
+    // Bind a metric to the current run, return false if no run is active.
+    bool BindMetric(metrics::MetricID metricID, const metrics::MetricMetadata& metadata);
+
+    // Bind a live metric, return true on success.
+    bool BindLiveMetric(metrics::MetricID metricID);
```
apazylbe, line 538: Could you elaborate here? What does "binding" a metric mean? Why do we need to call that? What happens if we don't? 
I would ideally like to know these things without reading the code.

include/ppx/application.h
```diff
@@ -523,13 +524,28 @@ class Application
     // See StartMetricsRun for why this wrapper is necessary.
     virtual bool HasActiveMetricsRun() const;
 
+    // Allocate a metric id to be used for a combined live/recorded metric.
```
apazylbe, line 527: Do we need to have all of these metric related functions in Application?

apazylbe (commented):
Sorry for the late review, only left a couple of comments cause I need to take another look fully.
Personally though, I find the metrics framework more confusing with these changes, and not sure if the complexity is justified. But if I am in minority, I am also fine with this.
Also I would like for Application to contain less code that doesn't need to be there, I understand we have some wrappers there to add additional metrics, but do the others functions need to be there?

hysw:
> Personally though, I find the metrics framework more confusing with these changes, and not sure if the complexity is justified. 

My preference is keeping it simple and separate, however Nathan have some good point, so I also made this version to see if integration makes sense.

PR #409, opened 2024-01-16
Update clang format

anishmgoyal:
Note: some slight changes vs. original change (#145)

The formatting results are a bit different, and the version of the lint action has changed. Also, this will retain the branch checkout action.

Also: seems like clang-format version 17 is more similar to gLinux than 16. If we see more differences, we might have to update this to clang-format version 18, but this works for our entire repo atm.


hysw:
I somewhat prefer `clang-format-16` since that seems to be the latest version supported by Visual Studio (at least from my VS2022 installation), doing custom formatter on MSVC is somewhat a pain.
On glinux, you can always `sudo apt install clang-format-16`.

hysw:
While you are at it, please
- add the `clang-format` version to `/docs/ci_testing.md` since that affects formatting
- maybe update `/docs/building_and_running.md` to recommend VS2022 if you decide to use `clang-format-16`

anishmgoyal:
> I somewhat prefer `clang-format-16` since that seems to be the latest version supported by Visual Studio (at least from my VS2022 installation), doing custom formatter on MSVC is a pain. Please let me know if Visual Studio updated their formatter. On glinux, you can always `sudo apt install clang-format-16`.

Fair enough. I'll change to 16, and maybe mark the one line of base_application.cpp as ignored, since the formatter fails miserably on that line. That's really the only diff I see between 16/17 for this codebase.

I'll mark this in the docs linked above.

hysw:
> since the formatter fails miserably on that line

The "fix" for that line:
```diff
@@ -54,9 +54,9 @@ std::filesystem::path BaseApplication::GetApplicationPath() const
     readlink("/proc/self/exe", buf, PATH_MAX);
     path = std::filesystem::path(buf);
 #elif defined(PPX_MSW)
-    HMODULE this_win32_module = GetModuleHandleA(nullptr);
-    char    buf[MAX_PATH];
+    char buf[MAX_PATH];
     std::memset(buf, 0, MAX_PATH);
+    HMODULE this_win32_module = GetModuleHandleA(nullptr);
     GetModuleFileNameA(this_win32_module, buf, MAX_PATH);
     path = std::filesystem::path(buf);
 #else
```

Yes, sometime `clang-format` is magical.

anishmgoyal:
Actually, it was in base_application.cpp, in a preprocessor directive. I didn’t see a clean fix for that, so I disabled the formatter in the last PR. I’ll play around some more though, and see if there’s a way to fix it without disabling the line.

hysw:
> Actually, it was in base_application.cpp, in a preprocessor directive. I didn’t see a clean fix for that, so I disabled the formatter in the last commit. I’ll play around some more though, and see if there’s a way to fix it without disabling the line.

That's the processor block after that line, and flipping the order of `buf` and `this_win32_module` stops `pid`'s assignment operator from trying to align it there.

anishmgoyal:
Ah! Ok - testing the recommended change.

apazylbe (approved):

hysw (approved):

PR #411, opened 2024-01-17
Rename projects to remove digit prefixes

angela28chen:
For internal testing, project folders cannot start with digits.


hysw:
We may still want a number for some of them to indicate they are basic feature demo and have incremental complexity, for the first 10 or so samples.

i.e. `sample_01_triangle`

angela28chen:
> We may still want a number for some of them to indicate they are basic feature demo and have incremental complexity, for the first 10 or so samples.

Why specifically the first 10 or so?

angela28chen:
I think if we want to divide the projects into groups to indicate that some are simple and some are more complex projects we could consider changing the grouping (including the applications under benchmark folder too) rather than all of them under projects folder and named with digits?

hysw:
> Why specifically the first 10 or so?

It gives some indication to where to start looking and understanding the structure. After 10 or so it's more feature specific.

> I think if we want to divide the projects into groups to indicate that some are simple and some are more complex projects we could consider changing the grouping (including the applications under benchmark folder too) rather than all of them under projects folder and named with digits?

Some of them are more like "tutorial", which the numbering does indicate the step.

apazylbe:
Agree with Peter, I think there is value in some numbering for incrementally complex samples, to indicate where you can start and how to progress. I like the `sample_n_name` scheme.

angela28chen:
> It gives some indication to where to start looking and understanding the structure. After 10 or so it's more feature specific.

I see, that makes sense. I took a look at the projects to try and understand which ones are useful to include in the samples. I'm thinking `sample_n_name` for samples 00-13 and then use just the name without # for 14 (input) and onward, how does that sound? (sample 13 is normal map)

apazylbe:
that lgtm, although I am wondering if 04_cube_xr should just be cube_xr instead of sample_04_cube_xr, since that is a now a fairly standalone XR project.

apazylbe:
@chaoticbob Any thoughts on the new naming scheme?

Keenuts (approved):

apazylbe (approved):

PR #413, opened 2024-01-22
Update time for zeroth frame

GrantComm:
Flocking, Shark, and Ocean all use time, which is updated in UpdateTime(). Call UpdateTime for the zeroth frame only, so that they all use the same time value.

### DIFF
--- a/projects/fishtornado_xr/Flocking.cpp
+++ b/projects/fishtornado_xr/Flocking.cpp
@@ -318,7 +318,7 @@ void Flocking::Shutdown()
     mMaterialConstants.Destroy();
 }
 
-void Flocking::Render(uint32_t frameIndex)
+void Flocking::Update(uint32_t frameIndex)
 {
     FishTornadoApp* pApp  = FishTornadoApp::GetThisApp();
     const float     t     = pApp->GetTime();
--- a/projects/fishtornado_xr/Flocking.cpp
+++ b/projects/fishtornado_xr/Flocking.cpp
@@ -318,7 +318,7 @@ void Flocking::Shutdown()
     mMaterialConstants.Destroy();
 }
 
-void Flocking::Update(uint32_t frameIndex)
+void Flocking::Render(uint32_t frameIndex)
 {
     FishTornadoApp* pApp  = FishTornadoApp::GetThisApp();
     const float     t     = pApp->GetTime();
--- a/projects/fishtornado_xr/Shark.cpp
+++ b/projects/fishtornado_xr/Shark.cpp
@@ -78,7 +78,7 @@ void Shark::Shutdown()
     mMaterialConstants.Destroy();
 }
 
-void Shark::Update(uint32_t frameIndex)
+void Shark::Update()
 {
     const float t = FishTornadoApp::GetThisApp()->GetTime();
 
@@ -94,7 +94,10 @@ void Shark::Update(uint32_t frameIndex)
 
     // Find direction of travel
     mDir = normalize(mVel);
+}
 
+void Shark::Render(uint32_t frameIndex)
+{
     // Calculate rotation matrix for orientation
     quat     q           = glm::rotation(float3(0, 0, 1), mDir);
     float4x4 rotMat      = glm::toMat4(q);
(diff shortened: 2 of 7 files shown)


projects/fishtornado_xr/Flocking.cpp
```diff
@@ -318,7 +318,7 @@ void Flocking::Shutdown()
     mMaterialConstants.Destroy();
 }
 
-void Flocking::Update(uint32_t frameIndex)
+void Flocking::Render(uint32_t frameIndex)
```
wangra-google, line 321: it would be better to keep the original name `Update` instead of renaming it to `Render`. Normally, `Render` would contain Drawcalls.

wangra-google (changes requested):

projects/fishtornado_xr/Flocking.cpp
wangra-google: maybe we could pass in `currentViewIndex` into the `Update()` function and update the perFrame and perEye data accordingly?

apazylbe: Agree that we should keep the name as `Update`.
I think we don't currently have the perEye/perView data, it is just reusing the same perFrame data, right? and it is indirectly getting the currentViewIndex info from GetCamera() and GetPoseForCurrentView() functions below. 
We can change the whole project to have separate per view data, but probably out scope for this change.

projects/fishtornado_xr/Shark.cpp
```diff
@@ -78,7 +78,7 @@ void Shark::Shutdown()
     mMaterialConstants.Destroy();
 }
 
-void Shark::Update(uint32_t frameIndex)
+void Shark::Update()
```
apazylbe, line 81: I think we can keep this as a single `Update` function and just call in the existing place.

Keenuts (commented):
Agreeing with the comments made. Delegating approval to Aliya.

projects/fishtornado_xr/Flocking.cpp
GrantComm: Reverted back to the original function names and added the currentViewIndex logic to the Shark's update function. Is the current change what you were looking for @wangra-google @apazylbe ?

wangra-google (approved):

projects/fishtornado_xr/Flocking.cpp
apazylbe: Flocking, Shark, and Ocean all use time, which is updated in `UpdateTime()`. Perhaps we should be calling `UpdateTime` for view = 0 only, and leave the other `Update` functions the same, so that they use the same time value for all views?

GrantComm: That makes sense. Wrapped the UpdateTime call in a conditional

projects/fishtornado_xr/Shark.cpp
```diff
@@ -78,22 +78,25 @@ void Shark::Shutdown()
     mMaterialConstants.Destroy();
 }
 
-void Shark::Update(uint32_t frameIndex)
+void Shark::Update(uint32_t frameIndex, uint32_t currentViewIndex)
```
wangra-google, line 81: As Aliya suggested, if we only Call UpdateTime once, there is no need to check the view index in the Update() function anymore

projects/fishtornado_xr/Flocking.cpp
wangra-google: @apazylbe I am wondering if we should have a single virtual FrameUpdate() function that is called once per frame not per eye (in application.cpp before the per eye dispatch render), and put any per frame update (like UpdateTime) in there. I think we could have more and more cases that requires update once per frame.

apazylbe: @wangra-google Perhaps. I actually think application.cpp shouldn't be calling Render() twice (still need to make that change though), then individual projects can structure Render() as they want. Although "PreRender/FrameUpdate" might be a pretty common method that we might want to add in Application.
No change for this PR though I think.

apazylbe:
Grant, can you update the PR description, that will be used as a commit. Thanks!

Issue #408, opened 2024-01-16
ClearRenderTarget/ClearDepthStencil don't work with dynamic render passes


apazylbe:
This is a follow up from https://github.com/google/bigwheels/pull/356

ClearRenderTarget/ClearDepthStencil (https://github.com/google/bigwheels/blob/b0bded04e4fc17048443ee4ef74507a2e6a942f2/src/ppx/grfx/vk/vk_command.cpp#L289) accept an image and clear an image while inside a render pass. The functions use current render pass variable to figure out the attachment index and and render area. When we have a dynamic render pass we can look up the attachment index by the image instead of the image view and save the render area on BeginRendering and BeginRenderpass. We would also have to check that there is an active render pass here https://github.com/google/bigwheels/blob/b0bded04e4fc17048443ee4ef74507a2e6a942f2/src/ppx/grfx/vk/vk_command.cpp#L294 with HasActiveRenderPass

alizahlalani:
pull #414 does not address DX12

As a follow up - before this issue can be closed, changes also need to be made for DX12 - https://github.com/google/bigwheels/blob/main/src/ppx/grfx/dx12/dx12_command.cpp#L416

PR #414, opened 2024-01-22
make ClearRenderTarget and ClearDepthStencil work with dynamic rendering

alizahlalani:
For issue #408  

ClearRenderTarget and ClearDepthStencil now work within a dynamic render pass. RenderArea is now saved directly in BeginRendering.

DynamicRendering sample is updated to call both these functions. These functions still work in the Cube sample as well (for a non dynamic rendering example).


src/ppx/grfx/grfx_command.cpp
```diff
@@ -71,6 +71,10 @@ void CommandBuffer::BeginRendering(const grfx::RenderingInfo* pRenderingInfo)
 
     BeginRenderingImpl(pRenderingInfo);
     mDynamicRenderPassActive = true;
+    mRenderArea              = pRenderingInfo->renderArea;
+    mRenderTargetView        = pRenderingInfo->pRenderTargetViews;
+    mRenderTargetCount       = pRenderingInfo->renderTargetCount;
+    mDepthStencilView        = pRenderingInfo->pDepthStencilView;
```
apazylbe, line 77: Could we reset these values in `EndRendering`?

src/ppx/grfx/grfx_command.cpp
```diff
@@ -71,6 +71,10 @@ void CommandBuffer::BeginRendering(const grfx::RenderingInfo* pRenderingInfo)
 
     BeginRenderingImpl(pRenderingInfo);
     mDynamicRenderPassActive = true;
+    mRenderArea              = pRenderingInfo->renderArea;
+    mRenderTargetView        = pRenderingInfo->pRenderTargetViews;
```
apazylbe, line 75: I think this is a bug to assign it like this, since pRenderingInfo might go out of scope by the time we use this array. But if we switch to vector this doesn't matter.

include/ppx/grfx/grfx_command.h
```diff
@@ -582,6 +582,15 @@ class CommandBuffer
     //
     void Draw(const grfx::FullscreenQuad* pQuad, uint32_t setCount, const grfx::DescriptorSet* const* ppSets);
 
+    bool HasActiveRenderPass() const;
+
+protected:
+    bool                           mDynamicRenderPassActive = false;
+    grfx::Rect                     mRenderArea              = {};
+    grfx::RenderTargetView* const* mRenderTargetView        = {};
+    uint32_t                       mRenderTargetCount       = 0;
```
apazylbe, line 591: Would you be able to either: 
- document that these are values for dynamic render passes only
- or save those on BeginRenderPass as well, which would potentially allow unifying code branches in ClearRenderTarget

src/ppx/grfx/vk/vk_command.cpp
```diff
@@ -290,36 +290,67 @@ void CommandBuffer::ClearRenderTarget(
...
+    if (!HasActiveRenderPass()) {
         return;
     }
 
-    // Make sure pImage is a render target in current render pass
-    const uint32_t renderTargetIndex = pCurrentRenderPass->GetRenderTargetImageIndex(pImage);
-    if (renderTargetIndex == UINT32_MAX) {
-        return;
+    grfx::Rect renderArea;
+    uint32_t   colorAttachment;
```
apazylbe, line 298: Perhaps we can initialize colorAttachment to UINT32_MAX and if it is still UINT32_MAX after the following if/else we can log a warning. Then we wouldn't need imageFound.

include/ppx/grfx/grfx_command.h
```diff
@@ -582,6 +582,15 @@ class CommandBuffer
     //
     void Draw(const grfx::FullscreenQuad* pQuad, uint32_t setCount, const grfx::DescriptorSet* const* ppSets);
 
+    bool HasActiveRenderPass() const;
```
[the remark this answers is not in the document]
alizahlalani: Done. Moved to protected.

include/ppx/grfx/grfx_command.h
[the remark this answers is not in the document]
alizahlalani: Done.

include/ppx/grfx/grfx_command.h
```diff
@@ -582,6 +582,15 @@ class CommandBuffer
...
+
+protected:
+    bool                           mDynamicRenderPassActive = false;
+    grfx::Rect                     mRenderArea              = {};
+    grfx::RenderTargetView* const* mRenderTargetView        = {};
+    uint32_t                       mRenderTargetCount       = 0;
+    grfx::DepthStencilView*        mDepthStencilView        = nullptr;
```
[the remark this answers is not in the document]
alizahlalani: Done.

projects/29_dynamic_rendering/DynamicRendering.cpp
```diff
@@ -126,19 +126,23 @@ void DynamicRenderingApp::Setup()
             grfx::RenderingInfo renderingInfo   = {};
             renderingInfo.flags.bits.suspending = true;
             renderingInfo.renderArea            = {0, 0, swapchain->GetWidth(), swapchain->GetHeight()};
-            renderingInfo.RTVClearValues[0]     = {{0.7f, 0.7f, 0.7f, 1.0f}};
-            renderingInfo.DSVClearValue         = {1.0f, 0xFF};
             renderingInfo.renderTargetCount     = 1;
-            renderingInfo.pRenderTargetViews[0] = swapchain->GetRenderTargetView(imageIndex, grfx::AttachmentLoadOp::ATTACHMENT_LOAD_OP_CLEAR);
+            renderingInfo.pRenderTargetViews[0] = swapchain->GetRenderTargetView(imageIndex, grfx::AttachmentLoadOp::ATTACHMENT_LOAD_OP_LOAD);
```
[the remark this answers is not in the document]
alizahlalani: Done.

projects/29_dynamic_rendering/DynamicRendering.cpp
```diff
@@ -126,19 +126,23 @@ void DynamicRenderingApp::Setup()
...
             renderingInfo.pDepthStencilView     = swapchain->GetDepthStencilView(imageIndex);
 
             float4x4 P   = glm::perspective(glm::radians(60.0f), GetWindowAspect(), 0.001f, 10000.0f);
             float4x4 V   = glm::lookAt(float3(0, 0, 5), float3(0, 0, 0), float3(0, 1, 0));
             float4x4 M   = glm::translate(float3(0.0, 0.0, -2.0)) * glm::scale(float3(2.0, 2.0, 2.0));
             float4x4 mat = P * V * M;
 
+            // Clear RTV to greyish blue
+            grfx::RenderTargetClearValue rtvClearValue = {0.23f, 0.23f, 0.33f, 0};
+            grfx::DepthStencilClearValue dsvClearValue = {1.0f, 0xFF};
```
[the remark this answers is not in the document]
alizahlalani: Done.

src/ppx/grfx/grfx_command.cpp
alizahlalani: Done.

src/ppx/grfx/grfx_command.cpp
alizahlalani: Done.

include/ppx/grfx/grfx_command.h
alizahlalani: Done.

src/ppx/grfx/vk/vk_command.cpp
alizahlalani: Would we want to log a warning or just return out?

apazylbe: I think we can log a warning and return

src/ppx/grfx/vk/vk_command.cpp
```diff
@@ -335,18 +362,44 @@ void CommandBuffer::ClearDepthStencil(
...
+    grfx::Rect renderArea;
+    uint32_t   baseArrayLayer;
+
+    // Dynamic render pass
+    if (mDynamicRenderPassActive) {
+        // Clear rect
+        renderArea = mRenderArea;
+
+        // Make sure pImage is depth stencil in depth stencil view
+        if (mDepthStencilView->GetImage().Get() != pImage) {
```
Keenuts, line 378: Not sure I follow, in both cases, `pImage` is compared to the attribute, and if unequal, nothing is done. Is this parameter useful? What is the scenario for that?

src/ppx/grfx/vk/vk_command.cpp
```diff
@@ -335,18 +362,44 @@ void CommandBuffer::ClearDepthStencil(
...
+        auto pCurrentRenderPass = GetCurrentRenderPass();
+
+        // Make sure pImage is depth stencil in current render pass
+        if (pCurrentRenderPass->GetDepthStencilImage().Get() != pImage) {
+            return;
+        }
 
-    // Get view
-    auto view = pCurrentRenderPass->GetDepthStencilView();
+        // Get view
```
Keenuts, line 394: Google style usually discourage adding comments that simply reformulates the associated line:
```
// get array layer
baseArrayLayer = view->GetArrayLayer()
```
The comment just repeats the function name, without adding value.
A few comments in this PR could be removed.

src/ppx/grfx/vk/vk_command.cpp
```diff
@@ -290,36 +290,63 @@ void CommandBuffer::ClearRenderTarget(
...
 
-    // Get view at renderTargetIndex
-    auto view = pCurrentRenderPass->GetRenderTargetView(renderTargetIndex);
+        // Get view at renderTargetIndex
+        auto view      = pCurrentRenderPass->GetRenderTargetView(renderTargetIndex);
+        baseArrayLayer = view->GetArrayLayer();
+    }
+
+    if (colorAttachment == UINT32_MAX) {
+        PPX_LOG_WARN("pImage is not a render target.");
```
Keenuts, line 331: Shall this be a warn or an assert?
Could that happen without it being an actual bug?

projects/29_dynamic_rendering/DynamicRendering.cpp
```diff
@@ -139,6 +138,11 @@ void DynamicRenderingApp::Setup()
 
             preRecordedCmd->BeginRendering(&renderingInfo);
             {
+                // Clear RTV to greyish blue
+                grfx::RenderTargetClearValue rtvClearValue = {0.23f, 0.23f, 0.33f, 0};
```
Keenuts, line 142: nit: any reasons to change the output?
(Internal tests should be fine if you do, but if changing the output is not required, might want to avoid)

```suggestion
                grfx::RenderTargetClearValue rtvClearValue = {0.7f, 0.7f, 0.7f, 1.f};
```

Keenuts (commented):
Sorry for the review delay!

include/ppx/grfx/grfx_command.h
```diff
@@ -582,6 +582,16 @@ class CommandBuffer
...
 
+protected:
+    // Returns true when inside a render pass (dynamic or regular)
+    bool HasActiveRenderPass() const;
+
+    // Below values are for dynamic render passes only.
+    bool                                   mDynamicRenderPassActive = false;
+    grfx::Rect                             mRenderArea              = {};
+    std::vector<grfx::RenderTargetViewPtr> mRenderTargetViews       = {};
+    grfx::DepthStencilViewPtr              mDepthStencilView        = nullptr;
```
apazylbe, line 593: perhaps we can either rename these or put them in a struct like `DynamicRenderPassInfo mCurrentDynamicRenderPassInfo` to make it clear that these are valid for an active dynamic render pass only

src/ppx/grfx/vk/vk_command.cpp
```diff
@@ -290,36 +290,63 @@ void CommandBuffer::ClearRenderTarget(
     grfx::Image*                        pImage,
     const grfx::RenderTargetClearValue& clearValue)
 {
-    auto pCurrentRenderPass = GetCurrentRenderPass();
-    if (IsNull(pCurrentRenderPass)) {
+    if (!HasActiveRenderPass()) {
```
apazylbe, line 293: I know it was like this before, but perhaps we should add a warning here, just in case someone expects this to work outside a render pass but it silently does not

src/ppx/grfx/vk/vk_command.cpp
apazylbe: I think technically it should be an assert. It would happen if someone passes an image that is not part of the renderpass, in which case it wouldn't work, so it won't do the clear. I don't see a reason why anyone would want to do that and not know that the clear did not work. So probably an assert is warranted.

src/ppx/grfx/vk/vk_command.cpp
apazylbe: I think it probably should do an assert if the image is not found. If someone passes an image that is not actually a DS attachment, the clear won't work and we probably should inform user about that

projects/29_dynamic_rendering/DynamicRendering.cpp
alizahlalani: changed to suggestion

src/ppx/grfx/vk/vk_command.cpp
alizahlalani: I've left it in and added the assert if image not found. If you prefer I remove this check altogether - let me know

include/ppx/grfx/grfx_command.h
```diff
@@ -191,6 +191,13 @@ struct RenderingInfo
     grfx::DepthStencilClearValue DSVClearValue                          = {1.0f, 0xFF};
 };
 
+struct DynamicRenderPassInfo
+{
+    grfx::Rect                             mRenderArea        = {};
+    std::vector<grfx::RenderTargetViewPtr> mRenderTargetViews = {};
+    grfx::DepthStencilViewPtr              mDepthStencilView  = nullptr;
+};
```
apazylbe, line 199: Could you move this to line 593 to be a nested class. I don't think there is a good reason to expose this in a public interface, and DynamicRenderPassInfo keyword can be used for other things.

src/ppx/grfx/grfx_command.cpp
```diff
@@ -79,7 +86,10 @@ void CommandBuffer::EndRendering()
     PPX_ASSERT_MSG(IsNull(mCurrentRenderPass), "Non-dynamic render pass active, use EndRendering instead");
 
     EndRenderingImpl();
-    mDynamicRenderPassActive = false;
+    mDynamicRenderPassActive                  = false;
+    mDynamicRenderPassInfo.mRenderArea        = {};
```
apazylbe, line 90: I think you can set `mDynamicRenderPassInfo = {};`

src/ppx/grfx/vk/vk_command.cpp
apazylbe: As Nathan mentioned, I think we can remove a few of these: on line 394, 397, 401

apazylbe (approved):
A few nits, otherwise lgtm, thanks!

src/ppx/grfx/vk/vk_command.cpp
alizahlalani: Removed parameter check for both dynamic and non dynamic renderpasses

Keenuts (approved):

PR #418, opened 2024-02-06
Add subsampledFormat for Image

bjoeris:
Add subsampledIformat for images and related classes.

### DIFF
--- a/include/ppx/grfx/grfx_draw_pass.h
+++ b/include/ppx/grfx/grfx_draw_pass.h
@@ -43,6 +43,7 @@ struct DrawPassCreateInfo
     grfx::RenderTargetClearValue renderTargetClearValues[PPX_MAX_RENDER_TARGETS]   = {};
     grfx::DepthStencilClearValue depthStencilClearValue                            = {};
     grfx::ShadingRatePattern*    pShadingRatePattern                               = nullptr;
+    bool                         subsampledFormat                                  = false;
 };
 
 //! @struct DrawPassCreateInfo2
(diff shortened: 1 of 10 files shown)


include/ppx/grfx/grfx_draw_pass.h
```diff
@@ -43,6 +43,7 @@ struct DrawPassCreateInfo
     grfx::RenderTargetClearValue renderTargetClearValues[PPX_MAX_RENDER_TARGETS]   = {};
     grfx::DepthStencilClearValue depthStencilClearValue                            = {};
     grfx::ShadingRatePattern*    pShadingRatePattern                               = nullptr;
+    bool                         subsampledFormat                                  = false;
```
apazylbe, line 46: Do you think instead of boolean we should add new struct to represent create flags (grfx::ImageCreateFlags) similar to grfx::ImageUsageFlags (https://github.com/google/bigwheels/blob/19dcf467601550420c6a9d32029f0c3137e989ee/include/ppx/grfx/grfx_helper.h#L210), then we can add anything else easily without adding and copying additional boolean values everywhere

Keenuts (approved):
LGTM. Creating this struct could help, but fine with both for now if that's too much changes.

include/ppx/grfx/grfx_draw_pass.h
bjoeris: Done. I added both ImageCreateFlags and SamplerCreateFlags. For the moment I only added the subsampledFormat flag. The other flags can be added as necessary.

apazylbe (approved):
Thank you!

PR #419, opened 2024-02-09
Adds XR Multiview Rendering Support

GrantComm:
Adds multiview implementation through XrComponent. Follow up prs will integrate multiview into Cube_XR and FishTornado_XR


src/ppx/grfx/vk/vk_device.cpp
```diff
@@ -256,6 +274,8 @@ Result Device::ConfigureFeatures(const grfx::DeviceCreateInfo* pCreateInfo, VkPh
     features.shaderStorageImageMultisample        = foundFeatures.shaderStorageImageMultisample;
     features.samplerAnisotropy                    = foundFeatures.samplerAnisotropy;
 
+    mHasMultiView = pCreateInfo->multiView && foundMultiViewFeatures.multiview;
```
apazylbe, line 277: It feels a little strange that `pCreateInfo->multiview` is only used here, I think we should either:
1) Leave the field there and only enable extension and feature if pCreateInfo->multiview is true (so if it is requested)
or
2) Just not have a field in DeviceCreateInfo, and enable multiview if it is supported.

src/ppx/grfx/vk/vk_device.cpp
```diff
@@ -538,6 +558,20 @@ Result Device::CreateApiObjects(const grfx::DeviceCreateInfo* pCreateInfo)
...
+        VkPhysicalDeviceMultiviewFeatures physicalDeviceMultiviewFeatures =
+            {
+                VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MULTIVIEW_FEATURES,
+                nullptr,
+                VK_TRUE,  // multiview
+                VK_FALSE, // multiviewGeometryShader
+                VK_FALSE  // multiviewTessellationShader
+            };
+
+        extensionStructs.push_back(reinterpret_cast<VkBaseOutStructure*>(&physicalDeviceMultiviewFeatures));
```
apazylbe, line 572: This is adding a pointer to a variable that will be out of scope when it is used later. `physicalDeviceMultiviewFeatures` should be defined above the if statement

include/ppx/grfx/grfx_pipeline.h
```diff
@@ -172,6 +178,7 @@ struct GraphicsPipelineCreateInfo
     grfx::ColorBlendState          colorBlendState    = {};
     grfx::OutputState              outputState        = {};
     grfx::ShadingRateMode          shadingRateMode    = grfx::SHADING_RATE_NONE;
+    grfx::MultiViewState           multiViewMask      = {};
```
apazylbe, line 181: Since it contains both view mask and correlation mask it should probably be named multiViewState or multiViewInfo

src/ppx/grfx/vk/vk_render_pass.cpp
```diff
@@ -133,6 +134,16 @@ Result RenderPass::CreateRenderPass(const grfx::internal::RenderPassCreateInfo*
...
     vkci.pDependencies          = &subpassDependencies;
 
+    if (hasMultiView && pCreateInfo->multiViewMask > 0) {
+        VkRenderPassMultiviewCreateInfo multiviewInfo = {VK_STRUCTURE_TYPE_RENDER_PASS_MULTIVIEW_CREATE_INFO};
+        multiviewInfo.subpassCount                    = 1;
+        multiviewInfo.pViewMasks                      = &pCreateInfo->multiViewMask;
+        multiviewInfo.correlationMaskCount            = 1;
+        multiviewInfo.pCorrelationMasks               = &pCreateInfo->multiCorrelationMask;
+
+        vkci.pNext = &multiviewInfo;
```
apazylbe, line 144: This is pointing to a variable that will be out of scope when used. `multiviewInfo` should be outside of the if statement.

src/ppx/grfx/vk/vk_render_pass.cpp
```diff
@@ -26,6 +26,7 @@ namespace vk {
 
 Result RenderPass::CreateRenderPass(const grfx::internal::RenderPassCreateInfo* pCreateInfo)
 {
+    bool hasMultiView   = ToApi(GetDevice())->HasMultiView() ? true : false;
```
apazylbe, line 29: `HasMultiView()` already returns a boolean, right? Is there a reason to add the ternary operator?

src/ppx/xr_component.cpp
```diff
@@ -727,36 +727,46 @@ void XrComponent::EndFrame(const std::vector<grfx::SwapchainPtr>& swapchains, ui
         layers.data(),                        // layers
     };
 
-    CHECK_XR_CALL(xrEndFrame(mSession, &frameEndInfo));
+    auto a = xrEndFrame(mSession, &frameEndInfo);
```
apazylbe, line 730: can you give it a descriptive name? like `res` or `ret` or similar? Or revert to the previous behaviour

include/ppx/grfx/grfx_swapchain.h
```diff
@@ -105,6 +105,7 @@ struct SwapchainCreateInfo
     grfx::Format              colorFormat         = grfx::FORMAT_UNDEFINED;
     grfx::Format              depthFormat         = grfx::FORMAT_UNDEFINED;
     uint32_t                  imageCount          = 0;
+    uint32_t                  arrayLayerCount     = 1;
```
apazylbe, line 108: Can you add a comment that this is for XR swapchains only?

include/ppx/grfx/grfx_pipeline.h
```diff
@@ -153,6 +153,12 @@ struct OutputState
     grfx::Format depthStencilFormat                          = grfx::FORMAT_UNDEFINED;
 };
 
+struct MultiViewState
```
apazylbe, line 156: Could you define this somewhere in a common header so we can use it for grfx_pipeline and grfx_renderpass? (maybe here include/ppx/grfx/grfx_helper.h)?

src/ppx/xr_component.cpp
```diff
@@ -733,30 +733,39 @@ void XrComponent::EndFrame(const std::vector<grfx::SwapchainPtr>& swapchains, ui
...
     // Projection and (optional) depth info layer from color+depth swapchains.
     for (size_t i = 0; i < viewCount; ++i) {
+        const uint32_t swapchain_index = startIndex + (indexToSwapchainMultiplier * i);
+
         XrCompositionLayerProjectionView view = {XR_TYPE_COMPOSITION_LAYER_PROJECTION_VIEW};
         view.pose                             = mViews[i].pose;
         view.fov                              = mViews[i].fov;
-        view.subImage.swapchain               = swapchains[startIndex + i]->GetXrColorSwapchain();
+        view.subImage.swapchain               = swapchains[swapchain_index]->GetXrColorSwapchain();
+        view.subImage.imageArrayIndex         = i * indexToArrayIndexMultiplier;
```
Keenuts, line 755: personal preference:
I'd be in favor or not using `indexToSwapchainMultipler` or similar, but something like:

```cpp
const bool isMultiView = isMultiView();
for (...)
    swapchain_index = isMultiView ? startIndex : startIndex + i;
    [...]
    view.subImage.imageArrayIndex = isMultiView ? i : 0;
```

src/ppx/grfx/grfx_swapchain.cpp
```diff
@@ -248,6 +252,11 @@ Result Swapchain::CreateRenderPasses()
         rpCreateInfo.depthStencilClearValue     = {1.0f, 0xFF};
         rpCreateInfo.ownership                  = grfx::OWNERSHIP_RESTRICTED;
         rpCreateInfo.pShadingRatePattern        = mCreateInfo.pShadingRatePattern;
+        rpCreateInfo.arrayLayerCount            = mCreateInfo.arrayLayerCount;
+#if defined(PPX_BUILD_XR)
+        rpCreateInfo.multiViewState.viewMask        = (mCreateInfo.pXrComponent && mCreateInfo.arrayLayerCount > 1 ? mCreateInfo.pXrComponent->GetDefaultViewMask() : 0);
```
Keenuts, line 257: super nit/personal preference (same below):

```suggestion
        if (mCreateInfo.pXrComponent && mCreateInfo.arrayLayerCount > 1)
            rpCreateInfo.multiViewState.viewMask = mCreateInfo.pXrComponent->GetDefaultViewMask();
```

Keenuts (approved):

src/ppx/grfx/vk/vk_device.cpp
```diff
@@ -259,6 +276,8 @@ Result Device::ConfigureFeatures(const grfx::DeviceCreateInfo* pCreateInfo, VkPh
     features.shaderStorageImageMultisample        = foundFeatures.shaderStorageImageMultisample;
     features.samplerAnisotropy                    = foundFeatures.samplerAnisotropy;
 
+    mHasMultiView = pCreateInfo->multiView;
```
apazylbe, line 279: I believe this should take into account whether extension is available (the check on line 207 above and foundMultiviewFeatures.multiview)

src/ppx/grfx/vk/vk_render_pass.cpp
```diff
@@ -26,6 +26,7 @@ namespace vk {
 
 Result RenderPass::CreateRenderPass(const grfx::internal::RenderPassCreateInfo* pCreateInfo)
 {
+    bool hasMultiView   = ToApi(GetDevice())->HasMultiView();
```
apazylbe, line 29: This field is used 100 lines below, can we move the declaration there?

src/ppx/grfx/vk/vk_render_pass.cpp
```diff
@@ -133,6 +134,16 @@ Result RenderPass::CreateRenderPass(const grfx::internal::RenderPassCreateInfo*
     vkci.dependencyCount        = 1;
     vkci.pDependencies          = &subpassDependencies;
 
+    VkRenderPassMultiviewCreateInfo multiviewInfo = {VK_STRUCTURE_TYPE_RENDER_PASS_MULTIVIEW_CREATE_INFO};
+    if (hasMultiView && pCreateInfo->multiViewState.viewMask > 0) {
```
apazylbe, line 138: I think we should assert that hasMultiview is true if multiviewState was provided. Otherwise it will silently ignore multiview state?

apazylbe (changes requested):

src/ppx/grfx/vk/vk_device.cpp
```diff
@@ -236,8 +241,20 @@ Result Device::ConfigureFeatures(const grfx::DeviceCreateInfo* pCreateInfo, VkPh
 {
     vk::Gpu* pGpu = ToApi(pCreateInfo->pGpu);
 
-    VkPhysicalDeviceFeatures foundFeatures = {};
-    vkGetPhysicalDeviceFeatures(pGpu->GetVkGpu(), &foundFeatures);
+    VkPhysicalDeviceFeatures          foundFeatures          = {};
+    VkPhysicalDeviceMultiviewFeatures foundMultiViewFeatures = {VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MULTIVIEW_FEATURES};
```
apazylbe, line 245: This is not used anywhere. I think in theory we need to check this in line 616 because in theory this can be false.

PR #420, opened 2024-02-12
Removing PPX_LINUX_HEADLESS which is overlapping with --headless.

angela28chen:


apazylbe (approved):
I don't think it is fully redundant, there could be some use for this since it removes linking an xcb library (added in https://github.com/google/bigwheels/pull/49). However, I don't know of existing uses for that, so I am fine with removing it and we can add it back if it turns out it was needed.

angela28chen:
Renamed PR & PPX_LINUX_HEADLESS has been removed from internal testing. Ready for merging.

Keenuts (approved):

PR #426, opened 2024-02-13
Add pipeline args util class

anishmgoyal:
Utility for creating graphics pipelines

### DIFF
--- a/include/ppx/scene/scene_pipeline_args.h
+++ b/include/ppx/scene/scene_pipeline_args.h
@@ -1,0 +1,219 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ppx_scene_pipeline_args_h
+#define ppx_scene_pipeline_args_h
+
+#include "ppx/scene/scene_config.h"
+#include "ppx/scene/scene_material.h"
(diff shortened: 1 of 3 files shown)


include/ppx/scene/scene_pipeline_args.h
```diff
@@ -0,0 +1,219 @@
...
+#include <unordered_map>
+
+namespace ppx {
+namespace scene {
+
+class MaterialPipelineArgs;
+
+// -------------------------------------------------------------------------------------------------
+
+// size = 8
```
Keenuts, line 32: You might want to use `static_assert(sizeof(X) == N)` and maybe `OFFSET_OF` with static_assert if the layout is important.

include/ppx/scene/scene_pipeline_args.h
```diff
@@ -0,0 +1,219 @@
...
+
+// size = 24
+struct MaterialTextureParams
+{
+    uint     textureIndex;      // offset = 0
+    uint     samplerIndex;      // offset = 4
+    float2x2 texCoordTransform; // offset = 8
+};
+
+// size = 184
```
Keenuts, line 64: same for all those structs for which the size & offsets are important.

Keenuts:
I think this should be tested in some way before we merge it.
Would it be possible to use this in one of the samples (or copy a sample and convert is to use this) so we can at least have some "coverage"?

apazylbe:
> I think this should be tested in some way before we merge it. Would it be possible to use this in one of the samples (or copy a sample and convert is to use this) so we can at least have some "coverage"?

@footballhead Does your sample use the changes in this PR?

include/ppx/scene/scene_pipeline_args.h
apazylbe: There is static assert in scene_pipeline_args.cpp for the constants, perhaps that is sufficient?

footballhead:
> > I think this should be tested in some way before we merge it. Would it be possible to use this in one of the samples (or copy a sample and convert is to use this) so we can at least have some "coverage"?
> 
> @footballhead Does your sample use the changes in this PR?

Yes. It's also used in #428 so that might be worth reviewing and checking in.

apazylbe:
Any objections to merging this?

PR #427, opened 2024-02-13
Add scene renderer class

anishmgoyal:


assets/CMakeLists.txt
```diff
@@ -20,3 +20,4 @@ add_subdirectory(fluid_simulation/shaders)
 add_subdirectory(gbuffer/shaders)
 add_subdirectory(materials/shaders)
 add_subdirectory(oit_demo/shaders)
+add_subdirectory(scene_renderer/shaders)
```
footballhead, line 23: Can we move this into it's own PR? These shaders are already checked in and this is needed to use them

footballhead: Done in #470. This line can be removed from this PR.

PR #434, opened 2024-02-16
[knob-framework] Knob Framework V2

angela28chen:
Adding a newer version of the knob framework and a demo that utilizes it.

Added in this PR:
- `ApplicationSettings.useKnobManagerNew` : A flag for applications to use the new knob framework on top of the old one
- `OptionsNew`: An intermediate format along with some supporting classes (`JsonConverterNew` and `CommandLineParserNew`) that are intended to replace `CliOptions` and `CommandLineParser`
- `KnobManagerNew` and `KnobNew`
- `knob_demo/` : Example project that features the newer knob framework
- Minor changes to `string_util`, adding Split function and removing whitespace padding from compound strings produced by `ToString()` functions

Some notable features of the newer version of knob framework:
- Decoupled knob function and display
- Runtime adjustable knob parameters (e.g. min, max, available subset of options)
- Runtime Load/Save support with JSON config files
- Can now identify if non-knob options are specified on the command line and terminate the application early

Future plans:
- Switch all projects and benchmarks to use the new knob framework.
- Change `ppx::Application` to use only the new knob framework and `mStandardOptsNew`
- Remove code for old knob framework and `CommandLineParser` + `CliOptions`
- Rename the replacements to remove "New" suffix

### DIFF
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -20,6 +20,7 @@
 #include "ppx/fs.h"
 #include "ppx/imgui_impl.h"
 #include "ppx/knob.h"
+#include "ppx/knob_new.h"
 #include "ppx/math_config.h"
 #include "ppx/metrics.h"
 #include "ppx/timer.h"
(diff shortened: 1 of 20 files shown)


angela28chen:
Closing because this is old, will re-open if I get the time to work on it again.

PR #435, opened 2024-02-16
[knob-framework] Add intermediate format `OptionsNew`

angela28chen:
Intermediate format `OptionsNew` is intended to help interface between command line arguments, JSON config files, and (in the future) knobs.


angela28chen:
This is the larger context for the knob framework changes: #434 
I am trying to break it into reviewable smaller pieces

PR #436, opened 2024-02-26
Update software_rendering.md

Keenuts:
Add a note to say the DLL shall be "unlocked".
By default Windows marks some files from remote locations as "locked". In the DLL case, this will prevent the DLL from being loaded, the the samples will fail to initialize.

### DIFF
--- a/docs/software_rendering.md
+++ b/docs/software_rendering.md
@@ -13,7 +13,7 @@ bin/bigwheels_application.exe --use-software-renderer
 
 **Note that while WARP already comes with Windows, we use DX12 features that require the latest version of WARP, which is distributed indipendently.
 You can download the latest pre-built WARP DLL [here](https://nuget.info/packages/Microsoft.Direct3D.WARP) (simply double-click `d3d10warp.dll` under `build/native/amd64` to download).
-Place this DLL in the same folder of your executables before running.**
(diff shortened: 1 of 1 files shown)


docs/software_rendering.md
```diff
@@ -13,7 +13,7 @@ bin/bigwheels_application.exe --use-software-renderer
 
 **Note that while WARP already comes with Windows, we use DX12 features that require the latest version of WARP, which is distributed indipendently.
 You can download the latest pre-built WARP DLL [here](https://nuget.info/packages/Microsoft.Direct3D.WARP) (simply double-click `d3d10warp.dll` under `build/native/amd64` to download).
-Place this DLL in the same folder of your executables before running.**
+Place this DLL in the same folder of your executables before running. WARNING: You have to tick "unlock" on the DLL's properties.**
```
apazylbe, line 16: I don't remember needing to do anything, and I don't see an option to "unlock". How do you get to that option?

Keenuts: When I download the DLL from the link, it comes from an "untrusted source".
This means there is a bit Windows sets. If you do right-click, properties, you should see a "unlock" button at the bottom.

Keenuts: What happened on my fresh windows install is the DLL would be loaded, then unloaded and the system wide DLL would get loaded instead.
Unlocking the DLL solved the issue.

apazylbe (approved):

PR #441, opened 2024-03-05
Fixes for Vulkan Validation errors

angela28chen:
* VUID-VkShaderModuleCreateInfo-pCode-08742 : Needs to be checked by messageIdName and not messageIdNumber
* VUID-VkPipelineRasterizationStateCreateInfo-pNext-pNext : Need extension VK_EXT_depth_clip_enable rather than VK_EXT_depth_range_unrestricted

### DIFF
--- a/src/ppx/grfx/vk/vk_instance.cpp
+++ b/src/ppx/grfx/vk/vk_instance.cpp
@@ -34,7 +34,7 @@ static VkBool32 VKAPI_PTR DebugUtilsMessengerCallback(
     // Ignore these messages because they're nonsense
     // clang-format off
     if (
-        (pCallbackData->messageIdNumber == 0x3d492883) // vkCreateShaderModule(): The SPIR-V Extension (SPV_GOOGLE_hlsl_functionality1) was declared, but none of the requirements were met to use it. The Vulkan spec states: If pCode declares any of the SPIR-V extensions listed in the SPIR-V Environment appendix, one of the corresponding requirements must be satisfied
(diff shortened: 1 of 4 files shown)


src/ppx/grfx/vk/vk_instance.cpp
```diff
@@ -34,7 +34,7 @@ static VkBool32 VKAPI_PTR DebugUtilsMessengerCallback(
     // Ignore these messages because they're nonsense
     // clang-format off
     if (
-        (pCallbackData->messageIdNumber == 0x3d492883) // vkCreateShaderModule(): The SPIR-V Extension (SPV_GOOGLE_hlsl_functionality1) was declared, but none of the requirements were met to use it. The Vulkan spec states: If pCode declares any of the SPIR-V extensions listed in the SPIR-V Environment appendix, one of the corresponding requirements must be satisfied
+        (std::string(pCallbackData->pMessageIdName) == "VUID-VkShaderModuleCreateInfo-pCode-08742") // vkCreateShaderModule(): The SPIR-V Extension (SPV_GOOGLE_hlsl_functionality1) was declared, but none of the requirements were met to use it. The Vulkan spec states: If pCode declares any of the SPIR-V extensions listed in the SPIR-V Environment appendix, one of the corresponding requirements must be satisfied
```
wangra-google, line 37: I did a quick search in the validation layer code, it is strange that we have different messageIdNumber if the pMessageIdName remains the same:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/layers/error_message/logging.cpp#L144

Unless we updated the validation layer recently so that the `hash_util::VuidHash` is different? @apazylbe

angela28chen: Maybe it has to do with this + different versions of Vulkan 1.1 being used? https://www.reddit.com/r/vulkan/comments/b41ev1/messageidnumber_reported_by_vulkan_validation/

In general it just seems like the `messageIdNumber` shouldn't be used. Official documentation also suggests `messageIdName` https://docs.vulkan.org/guide/latest/validation_overview.html#_special_usage_tags

wangra-google: i see, thanks for the links

wangra-google (approved):

apazylbe (approved):
Thanks!

PR #442, opened 2024-03-10
Add initial Wayland support

footballhead:
Screenshot from 2024-03-07 21-49-31

The window lacks the expected decoration in GNOME since we use glfw 3.3.6 and libdecor support was added 3.3.9.

This does not work well with high-DPI monitors. The window is made larger than the framebuffer.


Keenuts (approved):

Keenuts:
Thanks!
Just tested on Sway in pure Wayland, works fine, except the 2 issues I opened related to scaling.

PR #445, opened 2024-03-11
Wait for render complete before image acquisition

footballhead:
Recent Vulkan validation layers flag that imageAcquiredSemaphore may still be signaled during AcquireNextImage:

```
** VULKAN VALIDATION ERROR MESSAGE ***
Severity : ERROR
Type     : [VALIDATION]
Message  : Validation Error: [ VUID-vkAcquireNextImageKHR-semaphore-01779 ] Object 0: handle = 0x27a9e40000000112, type = VK_OBJECT_TYPE_SEMAPHORE; | MessageID = 0x5717e75b | vkAcquireNextImageKHR():  Semaphore must not have any pending operations. The Vulkan spec states: If semaphore is not VK_NULL_HANDLE it must not have any uncompleted signal or wait operations pending (https://vulkan.lunarg.com/doc/view/1.3.275.0/windows/1.3-extensions/vkspec.html#VUID-vkAcquireNextImageKHR-semaphore-01779)
```

Apart from Application frame rate pacing, there is no explicit synchronization between Present for the previous frame and AcquireNextImage for the next frame. As such, applications with no frame pacing or those that cannot hit the target FPS may not have finished rendering the previous frame when it goes to render the next frame.

The fix is to wait for renderCompleteFence before AcquireNextImage. This guarantees that imageAcquiredSemaphore is unsignaled.

### DIFF
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
@@ -1082,6 +1082,9 @@ void GraphicsBenchmarkApp::Render()
     uint32_t           imageIndex = UINT32_MAX;
     grfx::SwapchainPtr swapchain  = GetSwapchain(currentViewIndex);
 
+    // Wait for and reset render complete fence
+    PPX_CHECKED_CALL(frame.renderCompleteFence->WaitAndReset());
+
 #if defined(PPX_BUILD_XR)
     if (IsXrEnabled()) {
         PPX_ASSERT_MSG(swapchain->ShouldSkipExternalSynchronization(), "XRComponent should not be nullptr when XR is enabled!");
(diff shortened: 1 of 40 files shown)


benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -1092,6 +1092,9 @@ void GraphicsBenchmarkApp::Render()
     }
     else
 #endif
+        // Wait for and reset render complete fence
```
footballhead, line 1095: Reminder to self: this isn't going to work, consider:

- is this needed for XR? If so, can be moved up before the `
#if defined(PPX_BUILD_XR)` around line 1084?
- If this is not needed for XR then move it inside the `else` block at line 1098

footballhead: I spoke with @apazylbe and we believe that this wait is needed in both XR and non-XR cases. So I've moved it farther up to around line 1084

apazylbe (approved):
thanks!

Keenuts (approved):

footballhead:
The failing Linux build looks like a flake but I don't have the buttons to rerun it

PR #446, opened 2024-03-12
Add support for partial bindings (vk only)

anishmgoyal:
DX12 support may be added later. Adds supports for partially bound descriptors, which allows us to create arrays for textures where only a subset of the allotted slots are used.

### DIFF
--- a/include/ppx/grfx/grfx_descriptor.h
+++ b/include/ppx/grfx/grfx_descriptor.h
@@ -31,22 +31,25 @@ namespace grfx {
 //!
 struct DescriptorBinding
 {
-    uint32_t              binding         = PPX_VALUE_IGNORED;               //
-    grfx::DescriptorType  type            = grfx::DESCRIPTOR_TYPE_UNDEFINED; //
-    uint32_t              arrayCount      = 1;                               // WARNING: Not VkDescriptorSetLayoutBinding::descriptorCount
-    grfx::ShaderStageBits shaderVisiblity = grfx::SHADER_STAGE_ALL;          // Single value not set of flags (see note above)
+    uint32_t               binding         = PPX_VALUE_IGNORED;               //
+    grfx::DescriptorType   type            = grfx::DESCRIPTOR_TYPE_UNDEFINED; //
+    uint32_t               arrayCount      = 1;                               // WARNING: Not VkDescriptorSetLayoutBinding::descriptorCount
+    grfx::ShaderStageBits  shaderVisiblity = grfx::SHADER_STAGE_ALL;          // Single value not set of flags (see note above)
+    DescriptorBindingFlags flags;
 
     DescriptorBinding() {}
 
     DescriptorBinding(
-        uint32_t              binding_,
-        grfx::DescriptorType  type_,
-        uint32_t              arrayCount_      = 1,
-        grfx::ShaderStageBits shaderVisiblity_ = grfx::SHADER_STAGE_ALL)
+        uint32_t               binding_,
+        grfx::DescriptorType   type_,
+        uint32_t               arrayCount_      = 1,
+        grfx::ShaderStageBits  shaderVisiblity_ = grfx::SHADER_STAGE_ALL,
+        DescriptorBindingFlags flags_           = 0)
         : binding(binding_),
           type(type_),
           arrayCount(arrayCount_),
-          shaderVisiblity(shaderVisiblity_) {}
+          shaderVisiblity(shaderVisiblity_),
+          flags(flags_) {}
 };
 
 struct WriteDescriptor
--- a/src/ppx/grfx/vk/vk_descriptor.cpp
+++ b/src/ppx/grfx/vk/vk_descriptor.cpp
@@ -266,6 +266,7 @@ Result DescriptorSetLayout::CreateApiObjects(const grfx::DescriptorSetLayoutCrea
 
     std::vector<VkDescriptorSetLayoutBinding> vkBindings;
     std::vector<VkDescriptorBindingFlags>     vkBindingFlags;
+    bool                                      hasBindingFlags = false;
     for (size_t i = 0; i < pCreateInfo->bindings.size(); ++i) {
         const grfx::DescriptorBinding& baseBinding = pCreateInfo->bindings[i];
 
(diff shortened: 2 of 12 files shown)


src/ppx/grfx/vk/vk_descriptor.cpp
```diff
@@ -275,11 +276,20 @@ Result DescriptorSetLayout::CreateApiObjects(const grfx::DescriptorSetLayoutCrea
...
         vkBindings.push_back(vkBinding);
+
+        VkDescriptorBindingFlags vkBindingFlag = ToVkDescriptorBindingFlags(baseBinding.flags);
+        vkBindingFlags.push_back(vkBindingFlag);
     }
 
+    VkDescriptorSetLayoutBindingFlagsCreateInfoEXT vkBindingWrapper = {VK_STRUCTURE_TYPE_DESCRIPTOR_SET_LAYOUT_BINDING_FLAGS_CREATE_INFO_EXT};
+    vkBindingWrapper.bindingCount                                   = CountU32(vkBindingFlags);
+    vkBindingWrapper.pBindingFlags                                  = DataPtr(vkBindingFlags);
+    vkBindingWrapper.pNext                                          = nullptr;
```
apazylbe, line 287: Do we add this even if none of the flags are set? 
It seems like it comes from VK_EXT_descriptor_indexing extension, which is optional currently https://github.com/google/bigwheels/blob/a99d5bfbd4127f41677fe7113447aae587e90d4f/src/ppx/grfx/vk/vk_device.cpp#L154 so it's possible this extension won't be enabled. 
Perhaps we should only set this if the flags are actually not 0? And assert that the required features are supported?

anishmgoyal: Right - my mistake, I did mean to call the function and check for support. Will add this

src/ppx/grfx/vk/vk_device.cpp
```diff
@@ -297,6 +297,63 @@ Result Device::ConfigureFeatures(const grfx::DeviceCreateInfo* pCreateInfo, VkPh
...
+    if (!missingFeatures.empty()) {
+        std::stringstream ss;
+        ss << "Device does not support required features:" << PPX_LOG_ENDL;
+        for (const auto& elem : missingFeatures) {
+            ss << " " << elem << PPX_LOG_ENDL;
+        }
+        PPX_ASSERT_MSG(false, ss.str());
+        return ppx::ERROR_REQUIRED_FEATURE_UNAVAILABLE;
+    }
+
```
Keenuts, line 353: Why is this not:
```cpp
PPX_ASSERT_MSG(foundDiFeatures.runtimeDescriptorArray, "RuntimeDescriptorArray feature missing.");
return foundDiFeatures.runtimeDescriptorArray ? PPX::SUCCESS : PPX::ERROR_REQUIRED_FEATURE_UNAVAILABLE;
```

src/ppx/grfx/vk/vk_descriptor.cpp
```diff
@@ -275,12 +276,25 @@ Result DescriptorSetLayout::CreateApiObjects(const grfx::DescriptorSetLayoutCrea
         vkBinding.stageFlags                   = ToVkShaderStageFlags(baseBinding.shaderVisiblity);
         vkBinding.pImmutableSamplers           = nullptr;
         vkBindings.push_back(vkBinding);
+
+        if (baseBinding.flags != 0) {
+            ValidateDescriptorBindingFlags(baseBinding.flags);
+            VkDescriptorBindingFlags vkBindingFlag = ToVkDescriptorBindingFlags(baseBinding.flags);
+            vkBindingFlags.push_back(vkBindingFlag);
+        }
```
apazylbe, line 284: From the Vulkan spec:
[VkDescriptorSetLayoutCreateInfo](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkDescriptorSetLayoutCreateInfo.html)::pBindings[i] uses the flags in pBindingFlags[i]
I think it is possible for some flags to be 0 and for some flags to be not 0? So we have to insert 0 flags regardless in that case to preserve which flags apply to which binding

src/ppx/grfx/vk/vk_descriptor.cpp
```diff
@@ -275,12 +276,25 @@ Result DescriptorSetLayout::CreateApiObjects(const grfx::DescriptorSetLayoutCrea
         vkBinding.stageFlags                   = ToVkShaderStageFlags(baseBinding.shaderVisiblity);
         vkBinding.pImmutableSamplers           = nullptr;
         vkBindings.push_back(vkBinding);
+
+        if (baseBinding.flags != 0) {
+            ValidateDescriptorBindingFlags(baseBinding.flags);
```
apazylbe, line 281: should we check the return value?

src/ppx/grfx/vk/vk_descriptor.cpp
anishmgoyal: oh right - my mistake. i'll add that change

src/ppx/grfx/vk/vk_device.cpp
anishmgoyal: I wanted to keep it like this so that we can add more checks in the future

src/ppx/grfx/vk/vk_descriptor.cpp
anishmgoyal: yes - my mistake. updating this

src/ppx/grfx/vk/vk_device.cpp
```diff
@@ -987,6 +1013,12 @@ bool Device::FragmentStoresAndAtomicsSupported() const
     return mDeviceFeatures.fragmentStoresAndAtomics == VK_TRUE;
 }
 
+bool Device::PartialDescriptorBindingsSupported() const
+{
+    return mDescriptorIndexingFeatures.descriptorBindingPartiallyBound &&
+           mDescriptorIndexingFeatures.runtimeDescriptorArray;
```
apazylbe, line 1019: I am curious, is runtimeDescriptorArray feature required for partial descriptor bindings?

include/ppx/grfx/grfx_descriptor.h
```diff
@@ -31,22 +31,25 @@ namespace grfx {
...
 {
-    uint32_t              binding         = PPX_VALUE_IGNORED;               //
-    grfx::DescriptorType  type            = grfx::DESCRIPTOR_TYPE_UNDEFINED; //
-    uint32_t              arrayCount      = 1;                               // WARNING: Not VkDescriptorSetLayoutBinding::descriptorCount
-    grfx::ShaderStageBits shaderVisiblity = grfx::SHADER_STAGE_ALL;          // Single value not set of flags (see note above)
+    uint32_t               binding         = PPX_VALUE_IGNORED;               //
+    grfx::DescriptorType   type            = grfx::DESCRIPTOR_TYPE_UNDEFINED; //
+    uint32_t               arrayCount      = 1;                               // WARNING: Not VkDescriptorSetLayoutBinding::descriptorCount
+    grfx::ShaderStageBits  shaderVisiblity = grfx::SHADER_STAGE_ALL;          // Single value not set of flags (see note above)
+    DescriptorBindingFlags flags;
```
apazylbe, line 38: I think we should check the flags in the DX12 Create and emit a warning or error if these are set? Although I don't know how DX12 equivalent of descriptors behave, perhaps partial bindings are always supported.

apazylbe (approved):
A couple of comments, otherwise lgtm

src/ppx/grfx/vk/vk_device.cpp
anishmgoyal: Actually no - will remove this.

include/ppx/grfx/grfx_descriptor.h
anishmgoyal: Ahhh - my mistake, I did add a function for querying for support, seems I forgot to apply it. Will follow-up on this.

PR #459, opened 2024-04-29
Adds multiview to CubeXr

GrantComm:
Adds multiview into Cube_XR. This is a followup to pr#419 where multiview was added to the repo.

### DIFF
--- a/include/ppx/application.h
+++ b/include/ppx/application.h
@@ -89,6 +89,7 @@ struct StandardOptions
 #if defined(PPX_BUILD_XR)
     std::shared_ptr<KnobFlag<std::pair<int, int>>>      pXrUiResolution;
     std::shared_ptr<KnobFlag<std::vector<std::string>>> pXrRequiredExtensions;
+    std::shared_ptr<KnobFlag<bool>>                     pXrEnableMultiview;
 #endif
 
     std::shared_ptr<KnobFlag<std::vector<std::string>>> pAssetsPaths;
@@ -211,6 +212,7 @@ struct ApplicationSettings
 #if defined(PPX_BUILD_XR)
         std::pair<int, int>      xrUiResolution       = std::make_pair(0, 0);
         std::vector<std::string> xrRequiredExtensions = {};
+        bool                     xrEnableMultiview    = false;
 #endif
     } standardKnobsDefaultValue;
 };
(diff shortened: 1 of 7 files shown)


include/ppx/application.h
```diff
@@ -394,6 +394,21 @@ class Application
...
+        return viewProjectionMatrix;
+    }
+
+    uint32_t GetCurrentViewIndex() const
+    {
+        uint32_t current_view_index = 0;
+        current_view_index          = (IsXrEnabled() ? mXrComponent.GetCurrentViewIndex() : 0);
+
+        return current_view_index;
+    }
```
apazylbe, line 411: Instead of adding these, can we incorporate them into and go through XrCamera here: https://github.com/google/bigwheels/blob/7d2f93755c57f3dc0c24f805a7552af678017ab0/src/ppx/xr_component.cpp#L871
Otherwise I think we are reverting some changes made here https://github.com/google/bigwheels/pull/391

projects/cube_xr/CubeXr.cpp
```diff
@@ -32,14 +32,18 @@ void CubeXrApp::Config(ppx::ApplicationSettings& settings)
     settings.grfx.pacedFrameRate        = 0;
     settings.xr.enable                  = true;
     settings.xr.enableDebugCapture      = false;
+    settings.xr.enableMultiView         = true;
```
apazylbe, line 35: Ideally we set this from a flag instead

apazylbe (changes requested):

projects/cube_xr/CubeXr.cpp
```diff
@@ -32,14 +32,18 @@ void CubeXrApp::Config(ppx::ApplicationSettings& settings)
...
+#if PPX_ANDROID
+    settings.xr.enable = true;
+#endif
```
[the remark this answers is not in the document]
GrantComm: Removed

include/ppx/application.h
GrantComm: Removed these functions and just utilized the xrComponent.

projects/cube_xr/CubeXr.cpp
apazylbe: Not quite, perhaps you can add a KnobFlag<bool> here https://github.com/google/bigwheels/blob/main/include/ppx/application.h#L91 and initialize it here https://github.com/google/bigwheels/blob/main/src/ppx/application.cpp#L670

apazylbe (approved):

PR #479, opened 2024-06-06
Add calculations for YUV / planar images

anishmgoyal:
Also, create format definitions for YUV images.


src/ppx/graphics_util.cpp
```diff
@@ -31,6 +32,143 @@
 namespace ppx {
 namespace grfx_util {
 
+namespace {
+
+const uint32_t kYuvWidth  = 3152; // 3000;
+const uint32_t kYuvHeight = 3840; // 3000;
+
+// Start planar image helper functions
```
anishmgoyal, line 40: These methods aren't used in this PR, but will be used in a subsequent PR for methods to load planar textures.

include/ppx/graphics_util.h
```diff
@@ -149,6 +149,14 @@ class TextureOptions
         const std::filesystem::path& path,
         grfx::Texture**              ppTexture,
         const TextureOptions&        options);
+
+    friend Result LoadFramesFromRawVideo(
```
anishmgoyal, line 153: This method isn't used in this PR, but will be used later during the bandwidth tests (when productionized).

anishmgoyal, line 153: This isn't used in this PR, but will be used in subsequent PRs.

apazylbe (changes requested):

third_party/OpenXR-SDK-Source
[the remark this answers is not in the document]
anishmgoyal: Ahh, no - I'm not sure how that got in there, I'll remove that

src/ppx/graphics_util.cpp
```diff
@@ -31,6 +32,143 @@
 namespace ppx {
 namespace grfx_util {
 
+namespace {
+
+const uint32_t kYuvWidth  = 3152; // 3000;
+const uint32_t kYuvHeight = 3840; // 3000;
```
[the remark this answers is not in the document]
anishmgoyal: My mistake - another couple lines that must've carried over by mistake. I attempted to apply a git patch, and I think more of it got applied than intended.

src/ppx/graphics_util.cpp
```diff
@@ -31,6 +32,145 @@
...
+    bool hasColSubsampling = subsampling == grfx::FORMAT_CHROMA_SUBSAMPLING_420;
+    bool hasChromaValue    = false;
+    bool hasLumaValue      = false;
+    for (auto it = plane.members.begin(); it != plane.members.end(); ++it) {
+        const grfx::FormatPlaneDesc::Member& member = *it;
+        if (member.type == grfx::FORMAT_PLANE_CHROMA_TYPE_CHROMA) {
+            hasChromaValue = true;
+        }
+        else {
+            hasLumaValue = true;
```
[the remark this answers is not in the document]
anishmgoyal: Actually - this should at least have an else-if. I'll add an assert warning for the undefined case though.

include/ppx/grfx/grfx_format.h
```diff
@@ -255,11 +267,56 @@ struct FormatDesc
...
+    FORMAT_PLANE_CHROMA_TYPE_UNDEFINED,
+    FORMAT_PLANE_CHROMA_TYPE_LUMA,
+    FORMAT_PLANE_CHROMA_TYPE_CHROMA,
+};
+
+struct FormatPlaneDesc
+{
+    struct Member
+    {
+        // Note: it's expected that only one bit would be set here. That being
```
wangra-google, line 292: maybe add an assert to make sure there is only 1 bit set when we use this var?

src/ppx/graphics_util.cpp
```diff
@@ -31,6 +32,178 @@
...
+
+// Gets the height of a single plane, in terms of number of pixels represented.
+// This doesn't directly correlate to the number of bits / bytes for the plane's
+// height. The value returned can be used in a copy-image-to-buffer command.
+// plane: The plane to get the height for (containing information about the
+//        color components represented in the plane).
+// subsampling: The type of subsampling applied to chroma values for the image
+//              (e.g. 444, 422, 420).
+// imageHeight: The height of the image, in pixels, with no subsampling applied.
+uint32_t GetPlaneHeightForCopy(
```
wangra-google, line 47: maybe better without `ForCopy` since this could be used for other purpose than copy-image-to-buffer?

src/ppx/graphics_util.cpp
```diff
@@ -31,6 +32,178 @@
...
+}
+
+// Gets the width of a single plane, in terms of number of pixels represented.
+// This doesn't directly correlate to the number of bits / bytes for the plane's
+// height. The value returned can be used in a copy-image-to-buffer command.
+// plane: The plane to get the width for (containing information about the
+//        color components represented in the plane).
+// subsampling: The type of subsampling applied to chroma values for the image
+//              (e.g. 444, 422, 420).
+// imageHeight: The width of the image, in pixels, with no subsampling applied.
```
wangra-google, line 94: typo: imageWidth

include/ppx/grfx/grfx_format.h
anishmgoyal: Swapped it out for a new enum

src/ppx/graphics_util.cpp
anishmgoyal: Fixed

src/ppx/graphics_util.cpp
anishmgoyal: Replaced with 'InPixels', since it's the pixel height / width of a plane after subsampling. I'm not sure if there are uses besides copy where we'd use this, but just in case, this is at least more descriptive.

src/ppx/graphics_util.cpp
```diff
@@ -31,6 +32,178 @@
...
+    bool hasRowSubsampling = (subsampling == grfx::FORMAT_CHROMA_SUBSAMPLING_420) ||
+                             (subsampling == grfx::FORMAT_CHROMA_SUBSAMPLING_422);
+    bool hasChromaValue = false;
+    for (auto it = plane.members.begin(); it != plane.members.end(); ++it) {
+        const grfx::FormatPlaneDesc::Member& member = *it;
+        if (member.type == grfx::FORMAT_PLANE_CHROMA_TYPE_CHROMA) {
+            hasChromaValue = true;
+            break;
+        }
+    }
```
wangra-google, line 109: not really a bit deal, but could simply the code a bit with `std::any_of`, something like:
```
bool hasChromaValue = std::any_of(plane.members.begin(), plane.members.end(), 
                       [](const grfx::Member& m) { 
                           return m.type == grfx::FORMAT_PLANE_CHROMA_TYPE_CHROMA; 
                       });
```

src/ppx/graphics_util.cpp
```diff
@@ -1510,5 +1683,58 @@ Result CreateMeshFromFile(
...
+    if (size % frameSizeInBytes != 0) {
+        PPX_LOG_WARN("There may be partial frames in the raw video file at " << path << "; loading as much as possible.");
+    }
+
+    std::vector<char> buffer(frameSizeInBytes);
+    size_t            totalRead = 0;
+    while (totalRead + frameSizeInBytes <= size) {
+        const size_t readSize = file.Read(buffer.data(), size);
+        if (readSize != size) {
+            PPX_ASSERT_MSG(false, "Cannot read the content of the file!");
```
Keenuts, line 1729: The correct message should probably be something like `unable to load video frames: expected X bytes for format F, could only read Y bytes.`

src/ppx/graphics_util.cpp
```diff
@@ -1510,5 +1683,58 @@ Result CreateMeshFromFile(
...
+    }
+    const size_t size = file.GetLength();
+
+    if (size % frameSizeInBytes != 0) {
+        PPX_LOG_WARN("There may be partial frames in the raw video file at " << path << "; loading as much as possible.");
+    }
+
+    std::vector<char> buffer(frameSizeInBytes);
+    size_t            totalRead = 0;
+    while (totalRead + frameSizeInBytes <= size) {
```
Keenuts, line 1726: shall the last partial frame be loaded and unavailable bits be set to zero?
Or shall this be a hard error?
In what cases do we expect this to happen?

src/ppx/graphics_util.cpp
```diff
@@ -31,6 +32,145 @@
 namespace ppx {
 namespace grfx_util {
 
+namespace {
+
+const uint32_t kYuvWidth  = 3152; // 3000;
+const uint32_t kYuvHeight = 3840; // 3000;
```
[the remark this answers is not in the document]
anishmgoyal: Removed

third_party/OpenXR-SDK-Source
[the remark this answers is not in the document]
anishmgoyal: Removed

src/ppx/grfx/grfx_format.cpp
```diff
@@ -41,7 +43,24 @@ namespace grfx {
...
             FORMAT_COMPONENT_##Component,                                   \
-            OFFSETS_UNDEFINED                                               \
+            OFFSETS_UNDEFINED,                                              \
+            FORMAT_CHROMA_SUBSAMPLING_UNDEFINED,                            \
+            /*isPlanar=*/false                                              \
+    }
+
+#define UNCOMP_PLANAR_FORMAT(Name, Type, Aspect, BytesPerTexel, Layout, Component, ChromaSubsampling) \
+    {                                                                                                 \
+        "" #Name,                                                                                     \
```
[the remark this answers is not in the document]
anishmgoyal: It's not mandatory - I mostly kept it for uniformity, but I'll remove it from the macros.

src/ppx/grfx/grfx_format.cpp
```diff
@@ -41,7 +43,24 @@ namespace grfx {
...
+            FORMAT_DATA_TYPE_##Type,                                                                  \
+            FORMAT_ASPECT_##Aspect,                                                                   \
+            BytesPerTexel,                                                                            \
+            /*blockWidth=*/1,                                                                         \
+            -1,                                                                                       \
```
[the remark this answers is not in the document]
anishmgoyal: Done

src/ppx/grfx/grfx_format.cpp
```diff
@@ -192,6 +233,33 @@ const FormatDesc* GetFormatDescription(grfx::Format format)
...
+            // statement. This is for the compiler.
+            break;
+    }
+
+    const FormatDesc* description = GetFormatDescription(format);
+    if (description == nullptr || !description->isPlanar) {
+        PPX_LOG_WARN("Attempted to get planes for format " << format << ", which is non-planar");
+    }
+    else {
+        PPX_LOG_ERROR("Could not find planes for planar format " << format << "; returning null.");
```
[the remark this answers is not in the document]
anishmgoyal: I feel like it's ok for someone to look up planes in a non-planar format, so long as they respect the nullopt result. It's more of an issue (and necessarily something that a programmer should fix) if there are no planes defined for a planar format.

src/ppx/graphics_util.cpp
```diff
@@ -1510,5 +1650,57 @@ Result CreateMeshFromFile(
...
+        PPX_LOG_ERROR("Failed to fetch information for texture format " << format);
+        return ppx::ERROR_FAILED;
+    }
+
+    std::optional<grfx::FormatPlaneDesc> formatPlanes = grfx::GetFormatPlaneDescription(format);
+
+    uint32_t frameSizeInBytes = 0;
+    if (formatPlanes.has_value()) {
+        frameSizeInBytes = GetPlanarImageSizeInBytes(*formatDesc, *formatPlanes, width, height);
+    }
```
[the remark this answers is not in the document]
anishmgoyal: Fair point. Updated, and added an assertion that formatPlanes has a value.

src/ppx/graphics_util.cpp
```diff
@@ -31,6 +32,178 @@
...
+        if (member.type == grfx::FORMAT_PLANE_CHROMA_TYPE_CHROMA) {
+            hasChromaValue = true;
+        }
+        else if (member.type == grfx::FORMAT_PLANE_CHROMA_TYPE_LUMA) {
+            hasLumaValue = true;
+        }
+        else {
+            PPX_LOG_WARN("Member " << member.component << "has unknown chroma type.");
+        }
+    }
```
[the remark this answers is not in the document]
anishmgoyal: Updated in all occurences

src/ppx/graphics_util.cpp
```diff
@@ -1510,5 +1683,58 @@ Result CreateMeshFromFile(
...
+    const size_t size = file.GetLength();
+
+    if (size % frameSizeInBytes != 0) {
+        PPX_LOG_WARN("There may be partial frames in the raw video file at " << path << "; loading as much as possible.");
+    }
+
+    std::vector<char> buffer(frameSizeInBytes);
+    size_t            totalRead = 0;
+    while (totalRead + frameSizeInBytes <= size) {
+        const size_t readSize = file.Read(buffer.data(), size);
```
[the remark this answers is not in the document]
anishmgoyal: Fixed

src/ppx/graphics_util.cpp
anishmgoyal: I really don't expect it in any case, for someone to have an incomplete video file. I'm more open to the first option than the second, though, as I wouldn't want to prevent someone from using such a file in a benchmark. In almost all cases, I expect it to be a single frame "video" that's being loaded, as a snapshot of a camera feed.

src/ppx/graphics_util.cpp
anishmgoyal: Updated

src/ppx/graphics_util.cpp
anishmgoyal: Is functional-style preferred in bigwheels? I can update all relevant loops if so!

wangra-google: good question, I am not sure if there is a specific style we should follow, @apazylbe: is there any preference on this in BW? thanks

Keenuts: There are no clear preferences. However, be mindful of the readability:
Here, it's quite clear: any_of and a simple equal. IMO as readable as the equivalent range-based for if not more.
But if you start to convert all range-based for into `std::for_each`, then it can become messier than the simpler alternative.

My rule of thumb is: if the lambda is a one-liner, and I can use a `std::accumulate`, `std::any` or this kind of wrapper, I use it. If it's just to replace a for with a std::for_each, I don't.

src/ppx/grfx/grfx_format.cpp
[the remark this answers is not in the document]
Keenuts: OK, fair

src/ppx/graphics_util.cpp
```diff
@@ -31,6 +35,175 @@
...
+//        color components represented in the plane).
+// subsampling: The type of subsampling applied to chroma values for the image
+//              (e.g. 444, 422, 420).
+// imageHeight: The height of the image, in pixels, with no subsampling applied.
+uint32_t GetPlaneHeightInPixels(
+    const grfx::FormatPlaneDesc::Plane& plane,
+    grfx::FormatChromaSubsampling       subsampling,
+    uint32_t                            imageHeight)
+{
+    bool hasColSubsampling = (subsampling == grfx::FORMAT_CHROMA_SUBSAMPLING_420);
```
Keenuts, line 55: Shouldn't the condition be `subsampling != grfs::FORMAT_CHROMA_SUBSAMPLING_UNDEFINED`?

src/ppx/graphics_util.cpp
```diff
@@ -31,6 +35,175 @@
...
+    }
+
+    if (hasRowSubsampling && hasChromaValue) {
+        // Note: even if the layer has a luma value, generally, in the case of
+        // buffer copies, the width is treated as a half width, if we're
+        // subsampling at 4:2:0 or 4:2:2, and are looking at a plane with chroma
+        // values.
+        return imageWidth / 2;
+    }
+    return imageWidth;
```
Keenuts, line 119: Could you add a link here, or near the subsampling format definition to explain how the subsampling format affects the color/raw subsampling? So one with no knowledge about this type of frame (e.g. me 🙃) can understand why some formats have color subsampling, others height /2 and so on.

src/ppx/graphics_util.cpp
Keenuts: Ah got it.

src/ppx/graphics_util.cpp
anishmgoyal: For sure - I was referring to the other loops of this form in this class

src/ppx/graphics_util.cpp
anishmgoyal: Sure - I'll add some info in a comment

src/ppx/graphics_util.cpp
Keenuts: Yeah, the syntax is a bit messy. If you feel like it's less readable, stick with classic loops 😊

src/ppx/graphics_util.cpp
```diff
@@ -31,6 +35,175 @@
...
+            PPX_LOG_WARN(
+                "Frame size will be inaccurate, there is vertical subsampling "
+                "with both chroma and luma values present on a single plane, "
+                "which is not supported!");
+        }
+
+        // If we're subsampling at 4:2:0, the image will have half its height.
+        return imageHeight / 2;
+    }
+    return imageHeight;
```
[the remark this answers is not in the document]
Keenuts: Thanks for the details. Makes sense now that I know a single plane can carry Y and UV.

Keenuts (approved):

wangra-google (approved):

src/ppx/graphics_util.cpp
```diff
@@ -1510,5 +1706,65 @@ Result CreateMeshFromFile(
...
+    if (!file.Open(path)) {
+        PPX_ASSERT_MSG(false, "Cannot open the file!");
+        return ppx::ERROR_FAILED;
+    }
+    const size_t fileSize = file.GetLength();
+
+    std::vector<char> buffer(frameSize);
+    size_t            totalRead = 0;
+    while (totalRead < fileSize) {
+        const size_t bytesRead = file.Read(buffer.data(), frameSize);
```
apazylbe, line 1746: Is this safe to do after a std::move operation? Would buffer have frameSize capacity?

Issue #464, opened 2024-05-17
Consider removing debug capture swapchain for XR


apazylbe:
Added in https://github.com/google/bigwheels/pull/11 

Given we can capture with tools (for ex RenderDoc) directly, I am not sure if we still need this. It also makes code more complicated. So if this is not needed anymore, we should remove this.

apazylbe:
@wangra-google Thoughts? Do you think we still need this?

wangra-google:
oh, sorry i totally missed this conversation. Yeah, I don't think this is needed anymore. Feel free to remove this.

PR #481, opened 2024-06-07
Remove debug capture setting

apazylbe:
Remove debug capture option, tools like RenderDoc can capture directly, and it makes code more complicated.
Also use IsXrEnabled instead of settings.xr.enable, that way we can switch how that is controled in the function.

Fixes https://github.com/google/bigwheels/issues/464


Keenuts (approved):

Keenuts:
Thanks!

PR #482, opened 2024-06-11
[glTF] Load scenes with UINT8 indices

### LINKED ISSUE #453: [glTF] Support accessors with UNSIGNED_BYTE 5121 componentType
Using `scene::GltfLoader`, some models in [glTF-Sample-Assets](https://github.com/KhronosGroup/glTF-Sample-Assets) fail to load because they use index buffers with unsigned byte components (`componentType=5121`).

Refer to [5.24.2](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#_mesh_primitive_indices):

> When defined, the accessor MUST have SCALAR type and an unsigned integer component type.

Filtering [5.1.3](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#_accessor_componenttype) for that criteria leaves 5121 UNSIGNED_BYTE (unimplemented), 5123 UNSIGNED_SHORT (implemented), and 5125 UNSIGNED_INT (implemented)

This would allow BigWheels to load more models like:

- [AlphaBlendModeTest](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/AlphaBlendModeTest)
- [InterpolationTest](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/InterpolationTest)
- [MultiUVTest](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/MultiUVTest)
- [OrientationTest](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/OrientationTest)
- [SpecGlossVsMetalRough](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/SpecGlossVsMetalRough)
- [TextureCoordinateTest](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/TextureCoordinateTest)
- [VertexColorTest](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/VertexColorTest)

footballhead:
Load scenes that contain 8-bit index buffers. How the index buffer ends up being packed depends on the capabilities of the device:

- On Vulkan with `VK_KHR_index_type_uint8` extension, the UINT8 indices can be packed into a `INDEX_TYPE_UINT8` buffer. This is the optimal case as it doesn't require additional memory.
- Otherwise, the UINT8 indices need to be repacked into UINT16 indices. This comes at the cost of a larger GPU buffer.

This allows us to load more Khronos [glTF-Sample-Assets](https://github.com/KhronosGroup/glTF-Sample-Assets) like [TextureCoordinateTest](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/TextureCoordinateTest).

image

Fixes #453


footballhead:
Sg I'll look at adding UINT8 support to Geometry instead

footballhead:
Actually I'm looking at https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_index_type_uint8.html which I believe is required for VK_INDEX_TYPE_UINT8_KHR. That's what I think I need for UINT8 support in Geometry. I don't see that it's been promoted

Keenuts:
Ahh right, index buffer, not storage buffer, sorry I missed that!
Not promoted, but seems like it's widely supported (>70% on Sasha willems' DB) since it was a VK_EXT_index_type_int8 before.

apazylbe:
I tried on three devices I have (1 software lavapipe, and 2 android) and they all support VK_EXT_index_type_uint8. According to https://vulkan.gpuinfo.org/listextensions.php support for the extension is at 62%. Would it make sense to support UINT8 through the extension, or perhaps fall back to repacking if it is not available?

footballhead:
I like the idea of supporting the extension with this as a fallback. I'll look into how to make that happen.

footballhead:
I've incorporated #515 and #486 into this PR so it should be ready for review

src/ppx/scene/scene_gltf_loader.cpp
```diff
@@ -1644,7 +1672,8 @@ ppx::Result GltfLoader::LoadMeshData(
     for (uint32_t batchIdx = 0; batchIdx < CountU32(batchInfos); ++batchIdx) {
         const auto& batch = batchInfos[batchIdx];
 
-        const grfx::IndexType indexType       = (batch.indexFormat == grfx::FORMAT_R32_UINT) ? grfx::INDEX_TYPE_UINT32 : grfx::INDEX_TYPE_UINT16;
+        const grfx::IndexType indexType       = (batch.repackedIndexFormat == grfx::FORMAT_R32_UINT) ? grfx::INDEX_TYPE_UINT32 : (batch.repackedIndexFormat == grfx::FORMAT_R8_UINT) ? grfx::INDEX_TYPE_UINT8
```
apazylbe, line 1675: This is a little hard to read, perhaps we can add a switch instead?

apazylbe (approved):

apazylbe (approved):

footballhead:
@apazylbe merge conflicts are solved

PR #487, opened 2024-07-02
Add BlitImage (Vulkan Only)

jorgeag-google:
The second of a series of CL that integrates the commits from:

https://github.com/bjoeris/bigwheels/tree/experimental-foveation

into BW main's branch.

The series will have 4 CL, this is the second one that integrates the following commit

Add BlitImage (Vulkan Only)


include/ppx/grfx/grfx_command.h
```diff
@@ -162,6 +162,40 @@ struct ImageToImageCopyInfo
     } extent;
 };
 
+//! @struct ImageBlitInfo
+//!
+//!
+struct ImageBlitInfo
+{
+    struct
```
wangra-google, line 170: srcImage and dskImage use the same struct so that they can be only declared once instead of twice?

wangra-google (approved):

src/ppx/grfx/vk/vk_command.cpp
```diff
@@ -1035,6 +1035,65 @@ void CommandBuffer::CopyImageToImage(
...
 }
 
+void CommandBuffer::BlitImage(
+    const grfx::ImageBlitInfo* pCopyInfo,
+    grfx::Image*               pSrcImage,
+    grfx::Image*               pDstImage)
+{
+    bool isSourceDepthStencil = grfx::GetFormatDescription(pSrcImage->GetFormat())->aspect == grfx::FORMAT_ASPECT_DEPTH_STENCIL;
+    bool isDestDepthStencil   = grfx::GetFormatDescription(pDstImage->GetFormat())->aspect == grfx::FORMAT_ASPECT_DEPTH_STENCIL;
+    PPX_ASSERT_MSG(isSourceDepthStencil == isDestDepthStencil, "both images in an image copy must be depth-stencil if one is depth-stencil");
```
apazylbe, line 1045: Is it fine for them to differ if they are not both depth stencil? For example, one depth, one color

apazylbe (changes requested):

src/ppx/grfx/vk/vk_command.cpp
jorgeag-google: I belive so, the spec only says that  if one has depth or stencil the other one should also have it. No other restiction on the format

apazylbe: I see from the spec "If either of srcImage or dstImage was created with a depth/stencil format, the other must have exactly the same format", which is not what we are checking now. Also it seems like if formats are one with just depth and one with just stencil the assert will not trigger?

jorgeag-google: Yes, you are right I see your point. LI guess the correct check will be that both (dstImg and srcImg) have the same value in both fields (depth and stencil)

jorgeag-google: Actually, I switched it again to match what I believe the spec says.

include/ppx/grfx/dx12/dx12_command.h
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2024 Google LLC
```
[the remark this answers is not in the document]
apazylbe: Could you revert the copyright year modification?

src/ppx/grfx/vk/vk_command.cpp
apazylbe: I think it should be depth or stencil, maybe smth like this `isSourceDepthOrStencil = grfx::GetFormatDescription(pSrcImage->GetFormat())->aspect & grfx::FORMAT_ASPECT_DEPTH_STENCIL;` so it is true if it has either depth or stencil or both
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/5268c4004524149b6bb7230969038e834914d652/layers/core_checks/cc_copy_blit_resolve.cpp#L2937
Now that I am looking at the other code, I am not sure if this was copied from the other function. By the way, there are a bunch of other rules for using the function, so I am not sure if we need to assert this specific one. Although might still be useful.

src/ppx/grfx/vk/vk_command.cpp
```diff
@@ -1035,6 +1035,67 @@ void CommandBuffer::CopyImageToImage(
         &region);
 }
 
+void CommandBuffer::BlitImage(
+    const grfx::ImageBlitInfo* pCopyInfo,
+    grfx::Image*               pSrcImage,
+    grfx::Image*               pDstImage)
+{
+    bool isSourceDepthStencil = grfx::GetFormatDescription(pSrcImage->GetFormat())->aspect & grfx::FORMAT_ASPECT_DEPTH_STENCIL;
```
apazylbe, line 1043: isSourceDepthOrStencil

src/ppx/grfx/vk/vk_command.cpp
```diff
@@ -1035,6 +1035,67 @@ void CommandBuffer::CopyImageToImage(
...
+    bool isDestDepthStencil   = grfx::GetFormatDescription(pDstImage->GetFormat())->aspect & grfx::FORMAT_ASPECT_DEPTH_STENCIL;
```
apazylbe, line 1044: isDestDepthOrStencil

src/ppx/grfx/vk/vk_command.cpp
```diff
@@ -1035,6 +1035,67 @@ void CommandBuffer::CopyImageToImage(
...
+    if (isSourceDepthStencil || isDestDepthStencil) {
+        PPX_ASSERT_MSG(pSrcImage->GetFormat() == pDstImage->GetFormat(), "both images in an image copy must be the same format if one is depth-stencil");
```
apazylbe, line 1046: "both images in an image copy must have the same format if either has depth or stencil"

src/ppx/grfx/vk/vk_command.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2024 Google LLC
```
apazylbe, line 1: please update back to 2022

include/ppx/grfx/vk/vk_command.h
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2024 Google LLC
```
apazylbe, line 1: please update back to 2022

include/ppx/grfx/dx12/dx12_command.h
apazylbe, line 1: update back to 2022

apazylbe (commented):
just a few nits otherwise lgtm

src/ppx/grfx/vk/vk_command.cpp
jorgeag-google: Done

apazylbe (approved):

PR #488, opened 2024-07-03
Fix the errors while generating the gradlew.bat 

wangra-google:
Fix the errors while generating the gradlew.bat by running "gradle wrapper" by following https://stackoverflow.com/questions/69163511/build-was-configured-to-prefer-settings-repositories-over-project-repositories-b


apazylbe:
Hm, this change does not build for me on Linux. I will have to investigate more on why

wangra-google:
> Hm, this change does not build for me on Linux. I will have to investigate more on why

that is strange, seems to be building fine on linux on github's Check?

apazylbe:
> > Hm, this change does not build for me on Linux. I will have to investigate more on why
> 
> that is strange, seems to be building fine on linux on github's Check?

I think it is something to do with gradle and java version, what I have probably differs from the github runner

apazylbe:
I have Java 22 locally and it seems like it requires Gradle 8.8 (https://docs.gradle.org/current/userguide/compatibility.html). Updating it in here to 8.8 works for me https://github.com/google/bigwheels/blob/main/gradle/wrapper/gradle-wrapper.properties#L3 I can't think of any issues with upgrading the gradle version, so perhaps we can do that

wangra-google:
> I have Java 22 locally and it seems like it requires Gradle 8.8 (https://docs.gradle.org/current/userguide/compatibility.html). Updating it in here to 8.8 works for me https://github.com/google/bigwheels/blob/main/gradle/wrapper/gradle-wrapper.properties#L3 I can't think of any issues with upgrading the gradle version, so perhaps we can do that

updated to 8.8 as we discussed

apazylbe (approved):

PR #492, opened 2024-07-17
deps: update ImGUI to v1.90.3

Keenuts:
Updating ImGUI to v1.90.3.
Requires 2 small changes:
 - removed not unnecessary calls to the FontsTexture functions.
 - changed the setup for dynamic rendering.


apazylbe (approved):
thanks!

PR #498, opened 2024-07-29
Add multisample to foveation app

jorgeag-google:
Makes the foveation app able to use multisample foveation

This change was missing from the previous PR #493

### DIFF
--- a/benchmarks/foveation/FoveationBenchmarkApp.cpp
+++ b/benchmarks/foveation/FoveationBenchmarkApp.cpp
@@ -59,6 +59,12 @@ void FoveationBenchmarkApp::InitKnobs()
 
     GetKnobManager().InitKnob(&mKnobs.pExtraHashRounds, "extra-hash-rounds", 0);
     mKnobs.pExtraHashRounds->SetFlagDescription("Number of extra hash rounds to execute in the fragment shader.");
+
+    GetKnobManager().InitKnob(&mKnobs.pSampleCount, "sample-count", 1);
+    mKnobs.pSampleCount->SetFlagDescription("Number of samples for multisample rendering. Must be 1, 2 or 4.");
+    mKnobs.pSampleCount->SetValidator([](int sampleCount) {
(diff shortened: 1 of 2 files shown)


apazylbe (commented):
Did you run with validation layers? I assume it creates correct multisampled images for render target?

jorgeag-google:
I run it on device with both methods and saw the foveation pattern working
I run a debug build (on Android) and then look in the logcat and saw no errors.
So I assume it has the validation layers, but not sure

apazylbe:
We don't enable validation layers by default on debug Android builds, https://github.com/google/bigwheels/blob/5ea74e5741cd53be7c88632eec9aad943740f0ac/include/ppx/application.h#L146 just cause they take additional steps to set up. 
Overall LGTM, I would just double check with validation layers just in case.

apazylbe (approved):

wangra-google (approved):

jorgeag-google:
So, there is a validation error: 
```
VUID-VkSwapchainCreateInfoKHR-compositeAlpha-01280(ERROR / SPEC): msgNum: -1341842926 - Validation Error:
[ VUID-VkSwapchainCreateInfoKHR-compositeAlpha-01280 ] | MessageID = 0xb0051a12 | vkCreateSwapchainKHR(): pCreateInfo->compositeAlpha is not supported, support values are: VK_COMPOSITE_ALPHA_INHERIT_BIT_KHR%s                                                                                                    . The Vulkan spec states: compositeAlpha must be one of the bits present in the supportedCompositeAlpha member of the VkSurfaceCapabilitiesKHR structure returned by vkGetPhysicalDeviceSurfaceCapabilitiesKHR for the surface (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkSwapchainCreateInfoKHR-compositeAlpha-01280)
```
But there is common to all application on the device rather than just the foveation app

PR #500, opened 2024-08-12
Test case to run number of ALU instructions

RenfengLiu:
added a new knob    `vs-alu-instruction-count`. The sample can be run with following knobs: 
 "enable-skybox": false,
    "enable-spheres": false,
    "enable-metrics": false,
    "fullscreen-quads-count": 1,
    "fullscreen-quads-type": "Solid_Color",
    "alu-instruction-count": 200

### DIFF
--- a/assets/benchmarks/shaders/Benchmark_VsSimpleQuads.hlsl
+++ b/assets/benchmarks/shaders/Benchmark_VsSimpleQuads.hlsl
@@ -14,15 +14,6 @@
 
 #include "Benchmark_Quad.hlsli"
 
-float randomCompute(uint32_t instCount, float4 Position) {
-  float randNum = frac(float(instCount) * 123.456f);
-  for (uint32_t i = 0; i < instCount; i++) {
-    Position.z += Position.x * (1 - randNum) + randNum * Position.y;
-  }
-
-  return frac(Position.z);;
-}
-
 VSOutputPos vsmain(float4 Position : POSITION) {
   VSOutputPos result;
   result.position = Position;
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
@@ -183,9 +183,9 @@ void GraphicsBenchmarkApp::InitKnobs()
         pResolution->SetIndent(1);
     }
 
-    GetKnobManager().InitKnob(&pKnobVsAluCount, "vs-alu-instruction-count", /* defaultValue = */ 100, /* minValue = */ 100, 400);
-    pFullscreenQuadsCount->SetDisplayName("Number of ALU instructions in the vertex shader");
-    pFullscreenQuadsCount->SetFlagDescription("Select the number of ALU instructions in the vertex shader.");
+    GetKnobManager().InitKnob(&pKnobAluCount, "alu-instruction-count", /* defaultValue = */ 100, /* minValue = */ 100, 400);
+    pKnobAluCount->SetDisplayName("Number of ALU instructions in the shader");
+    pKnobAluCount->SetFlagDescription("Select the number of ALU instructions in the shader.");
 }
 
 void GraphicsBenchmarkApp::Config(ppx::ApplicationSettings& settings)
@@ -1736,7 +1736,7 @@ void GraphicsBenchmarkApp::RecordCommandBufferFullscreenQuad(PerFrame& frame, si
 {
     // Vertex shader push constant
     {
-        mQuadPushConstant.InstCount = pKnobVsAluCount->GetValue();
+        mQuadPushConstant.InstCount = pKnobAluCount->GetValue();
         frame.cmd->PushGraphicsConstants(mQuadsPipelineInterfaces[0], GetPushConstCount(mQuadPushConstant.InstCount), &mQuadPushConstant.InstCount, offsetof(QuadPushConstant, InstCount) / sizeof(uint32_t));
     }
     switch (pFullscreenQuadsType->GetValue()) {
--- a/assets/benchmarks/shaders/Benchmark_Quad.hlsli
+++ b/assets/benchmarks/shaders/Benchmark_Quad.hlsli
@@ -30,4 +30,13 @@ struct VSOutputPos {
   float4 position : SV_POSITION;
 };
 
-#endif // BENCHMARKS_VS_OUTPUT_HLSLI
+float randomCompute(uint32_t instCount, float4 Position) {
+  float randNum = frac(float(instCount) * 123.456f);
+  for (uint32_t i = 0; i < instCount; i++) {
+    Position.z += Position.x * (1 - randNum) + randNum * Position.y;
+  }
+
+  return frac(Position.z);;
+}
+
+#endif // BENCHMARKS_QUAD_HLSLI
--- a/assets/benchmarks/shaders/Benchmark_SolidColor.hlsl
+++ b/assets/benchmarks/shaders/Benchmark_SolidColor.hlsl
@@ -16,5 +16,7 @@
 
 float4 psmain(VSOutputPos input) : SV_TARGET
 {
-    return float4(Config.ColorValue, 1.0f);
+    float4 color = float4(Config.ColorValue, 1.0f);
+    color.a = randomCompute(Config.InstCount, input.position); 
+    return color;
 }
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.h
@@ -528,7 +528,7 @@ private:
     std::shared_ptr<KnobDropdown<grfx::Format>>        pFramebufferFormat;
     std::shared_ptr<KnobDropdown<std::pair<int, int>>> pResolution;
 
-    std::shared_ptr<KnobFlag<int>> pKnobVsAluCount;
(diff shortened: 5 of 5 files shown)


assets/benchmarks/shaders/Benchmark_VsSimpleQuads.hlsl
```diff
@@ -14,9 +14,18 @@
 
 #include "VsOutput.hlsli"
 
-VSOutputPos vsmain(float4 Position : POSITION)
-{
-	VSOutputPos result;
-	result.position = Position;
-	return result;
+float randomCompute(uint32_t instCount, float4 Position) {
+  float ret = frac(sin(Position.x) + cos(Position.y));
```
wangra-google, line 18: sorry i forget to mention, try to avoid using EFU (cos, sin) instructions if possible, they are special instructions that are more costly. My worry (i could be wrong though) is that those instructions might consume more power than normal instructions

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -808,6 +814,10 @@ void GraphicsBenchmarkApp::SetupFullscreenQuadsPipelines()
         piCreateInfo.sets[0].set                       = 0;
         piCreateInfo.sets[0].pLayout                   = mFullscreenQuads.descriptorSetLayout;
 
+        // piCreateInfo.pushConstants.count               = (2 * sizeof(uint32_t) + sizeof(float3)) / 4;
```
wangra-google, line 817: can be removed?

assets/benchmarks/shaders/VsOutput.hlsli
```diff
@@ -15,9 +15,23 @@
...
 #define BENCHMARKS_VS_OUTPUT_HLSLI
 
-struct VSOutputPos
-{
-    float4 position : SV_POSITION;
+struct ColorParams {
+  float3 Value;
+};
+
+struct ConfigParams {
```
wangra-google, line 22: out of curiosity, would the shader compiler add padding in between `RandomSeed` and `Color`?

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -1721,7 +1731,7 @@ void GraphicsBenchmarkApp::RecordCommandBufferFullscreenQuad(PerFrame& frame, si
     switch (pFullscreenQuadsType->GetValue()) {
         case FullscreenQuadsType::FULLSCREEN_QUADS_TYPE_NOISE: {
             uint32_t noiseQuadRandomSeed = (uint32_t)seed;
-            frame.cmd->PushGraphicsConstants(mQuadsPipelineInterfaces[0], 1, &noiseQuadRandomSeed);
+            frame.cmd->PushGraphicsConstants(mQuadsPipelineInterfaces[0], sizeof(uint32_t) / 4, &noiseQuadRandomSeed, sizeof(uint32_t) / 4);
```
wangra-google, line 1734: maybe better with a template helper function to get count?

wangra-google:
> One issue with this sample is it's not exactly 100/200/300/4000 instructions in the Dive capture,
> 
> actually the number of instructions is the same, since it's using the loop in the shader. the cycles needed should be different though.

no worries, doesn't have to be the exact number, just want to see the ALU impact on power consumption

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -190,10 +190,11 @@ void GraphicsBenchmarkApp::Config(ppx::ApplicationSettings& settings)
     settings.grfx.api                   = kApi;
     settings.grfx.numFramesInFlight     = 1;
     settings.grfx.swapchain.depthFormat = grfx::FORMAT_D32_FLOAT;
+    settings.grfx.enableDebug           = true;
```
wangra-google, line 193: are those local changes that are not supposed to be checkedin?

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -1749,10 +1747,10 @@ void GraphicsBenchmarkApp::RecordCommandBufferFullscreenQuad(PerFrame& frame, si
             float3 colorValues = pFullscreenQuadsColor->GetValue();
             colorValues *= intensity;
             uint32_t inst = pKnobVsAluCount->GetValue();
-            PPX_LOG_INFO("ALU instruction count " << inst);
+            // PPX_LOG_INFO("ALU instruction count " << inst);
```
wangra-google, line 1750: i guess you still want to keep this?

assets/benchmarks/shaders/VsOutput.hlsli
wangra-google: ah, good to know, thanks!

assets/benchmarks/shaders/Benchmark_Texture.hlsl
```diff
@@ -14,7 +14,7 @@
 
 #include "VsOutput.hlsli"
 
-Texture2D Tex0 : register(t0);
+Texture2D Tex0 : register(t1);
```
wangra-google, line 17: why do we need to change this to slot 1?

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -1721,7 +1727,7 @@ void GraphicsBenchmarkApp::RecordCommandBufferFullscreenQuad(PerFrame& frame, si
     switch (pFullscreenQuadsType->GetValue()) {
         case FullscreenQuadsType::FULLSCREEN_QUADS_TYPE_NOISE: {
             uint32_t noiseQuadRandomSeed = (uint32_t)seed;
-            frame.cmd->PushGraphicsConstants(mQuadsPipelineInterfaces[0], 1, &noiseQuadRandomSeed);
+            frame.cmd->PushGraphicsConstants(mQuadsPipelineInterfaces[0], sizeof(uint32_t) / sizeof(uint32_t), &noiseQuadRandomSeed, sizeof(uint32_t) / 4);
```
wangra-google, line 1730: would it be cleaner to have a helper function, something like `size_t GetPushConstCount<T>()` so that we could use function everywhere we need to get the count?

assets/benchmarks/shaders/Benchmark_Texture.hlsl
wangra-google: could you add some comments here to explain why 1 is used? (in case some one want to make modifications in the future. )

assets/benchmarks/shaders/Benchmark_Texture.hlsl
```diff
@@ -12,9 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "VsOutput.hlsli"
+#include "Benchmark_Quad.hlsli"
 
-Texture2D Tex0 : register(t0);
+Texture2D Tex0 : register(t1);  // Slot 0 was used by push constant
```
wangra-google, line 17: nits: i guess you mean `is` not `was` :)

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -785,8 +798,8 @@ void GraphicsBenchmarkApp::SetupFullscreenQuadsPipelines()
     {
         grfx::PipelineInterfaceCreateInfo piCreateInfo = {};
         piCreateInfo.setCount                          = 0;
-        piCreateInfo.pushConstants.count               = sizeof(uint32_t) / 4;
-        piCreateInfo.pushConstants.binding             = 0;
+        piCreateInfo.pushConstants.count               = sizeof(QuadPushConstant) / sizeof(uint32_t);
```
wangra-google, line 801: GetPushConstCount<QuadPushConstant>

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -795,8 +808,8 @@ void GraphicsBenchmarkApp::SetupFullscreenQuadsPipelines()
     {
         grfx::PipelineInterfaceCreateInfo piCreateInfo = {};
         piCreateInfo.setCount                          = 0;
-        piCreateInfo.pushConstants.count               = sizeof(float3) / 4;
-        piCreateInfo.pushConstants.binding             = 0;
+        piCreateInfo.pushConstants.count               = sizeof(QuadPushConstant) / sizeof(uint32_t);
```
wangra-google, line 811: GetPushConstCount<QuadPushConstant>

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -807,6 +820,9 @@ void GraphicsBenchmarkApp::SetupFullscreenQuadsPipelines()
         piCreateInfo.setCount                          = 1;
         piCreateInfo.sets[0].set                       = 0;
         piCreateInfo.sets[0].pLayout                   = mFullscreenQuads.descriptorSetLayout;
+        piCreateInfo.pushConstants.count               = sizeof(QuadPushConstant) / sizeof(uint32_t);
```
wangra-google, line 823: GetPushConstCount<QuadPushConstant>

wangra-google (approved):

PR #501, opened 2024-08-13
Test case #2: PS Mem fetch with output

RenfengLiu:
The test cases add a new knob `texture-count`, and can be run with 

```
{
    "deterministic": true,
    "enable-skybox": false,
    "enable-spheres": false,
    "enable-metrics": false,
    "fullscreen-quads-count": 1,
    "fullscreen-quads-type": "Texture",
    "vs-alu-instruction-count": 100,
    "texture-count": 2
}
```

### DIFF
--- a/assets/benchmarks/shaders/Benchmark_Texture.hlsl
+++ b/assets/benchmarks/shaders/Benchmark_Texture.hlsl
@@ -14,43 +14,15 @@
 
 #include "Benchmark_Quad.hlsli"
 
-Texture2D Tex0 : register(t1);  // Slot 0 is used by push constant.
-Texture2D Tex1 : register(t2);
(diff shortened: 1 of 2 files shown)


assets/benchmarks/shaders/Benchmark_Texture.hlsl
```diff
@@ -15,8 +15,42 @@
 #include "Benchmark_Quad.hlsli"
 
 Texture2D Tex0 : register(t1);  // Slot 0 is used by push constant.
```
wangra-google, line 17: i am wondering if you could use Texture2D tex[10] so that it would be easier to extend in the future?

wangra-google (approved):

PR #502, opened 2024-08-14
Test case 3/4: Disable PS output for ALU and memfetch

RenfengLiu:
Added knob `disable-ps-output` to disable PS output.
For test case 3, use `    "fullscreen-quads-type": "Solid_Color"` and for test case 4 use `    "fullscreen-quads-type": "Texture"`
{
    "deterministic": true,
    "enable-skybox": false,
    "enable-spheres": false,
    "enable-metrics": false,
    "fullscreen-quads-count": 1,
    "fullscreen-quads-type": "Texture",
    "vs-alu-instruction-count": 100,
    "texture-count": 2,
    "disable-ps-output": true
}

### DIFF
--- a/src/ppx/grfx/grfx_pipeline.cpp
+++ b/src/ppx/grfx/grfx_pipeline.cpp
@@ -24,6 +24,15 @@ namespace grfx {
 // -------------------------------------------------------------------------------------------------
 // BlendAttachmentState
 // -------------------------------------------------------------------------------------------------
+grfx::BlendAttachmentState BlendAttachmentState::BlendModeNone()
+{
+    grfx::BlendAttachmentState state = {};
+    state.blendEnable                = false;
+    state.colorWriteMask             = grfx::ColorComponentFlags::RGBA();
+
+    return state;
+}
+
 grfx::BlendAttachmentState BlendAttachmentState::BlendModeAdditive()
 {
     grfx::BlendAttachmentState state = {};
@@ -103,12 +112,6 @@ grfx::BlendAttachmentState BlendAttachmentState::BlendModeDisableOutput()
 {
     grfx::BlendAttachmentState state = {};
     state.blendEnable                = false;
-    state.srcColorBlendFactor        = grfx::BLEND_FACTOR_ONE;
-    state.dstColorBlendFactor        = grfx::BLEND_FACTOR_ONE;
-    state.colorBlendOp               = grfx::BLEND_OP_ADD;
-    state.srcAlphaBlendFactor        = grfx::BLEND_FACTOR_ONE;
-    state.dstAlphaBlendFactor        = grfx::BLEND_FACTOR_ONE;
-    state.alphaBlendOp               = grfx::BLEND_OP_ADD;
     state.colorWriteMask             = grfx::ColorComponentFlags(0);
 
     return state;
@@ -171,7 +174,7 @@ void FillOutGraphicsPipelineCreateInfo(
         for (uint32_t i = 0; i < pDstCreateInfo->colorBlendState.blendAttachmentCount; ++i) {
             switch (pSrcCreateInfo->blendModes[i]) {
                 case grfx::BLEND_MODE_NONE: {
-                    pDstCreateInfo->colorBlendState.blendAttachments[i].colorWriteMask = grfx::ColorComponentFlags::RGBA();
+                    pDstCreateInfo->colorBlendState.blendAttachments[i] = grfx::BlendAttachmentState::BlendModeNone();
                     break;
                 }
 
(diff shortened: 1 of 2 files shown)


src/ppx/grfx/grfx_pipeline.cpp
```diff
@@ -85,14 +85,14 @@ grfx::BlendAttachmentState BlendAttachmentState::BlendModeUnder()
 grfx::BlendAttachmentState BlendAttachmentState::BlendModePremultAlpha()
 {
     grfx::BlendAttachmentState state = {};
```
wangra-google, line 87: oh, sorry, i wasn't clear. you need to add a new blending mode, please do not modify the existing PremultAlpha

src/ppx/grfx/grfx_pipeline.cpp
```diff
@@ -97,6 +98,21 @@ grfx::BlendAttachmentState BlendAttachmentState::BlendModePremultAlpha()
     return state;
 }
 
+grfx::BlendAttachmentState BlendAttachmentState::BlendModeBenchmarkDisableOutput()
+{
```
wangra-google, line 102: probably do not need `Benchmark` in the name, this blendmode can be used for other apps

src/ppx/grfx/grfx_pipeline.cpp
```diff
@@ -153,7 +169,10 @@ void FillOutGraphicsPipelineCreateInfo(
         pDstCreateInfo->colorBlendState.blendAttachmentCount = pSrcCreateInfo->outputState.renderTargetCount;
         for (uint32_t i = 0; i < pDstCreateInfo->colorBlendState.blendAttachmentCount; ++i) {
             switch (pSrcCreateInfo->blendModes[i]) {
-                default: break;
+                default: {
+                    pDstCreateInfo->colorBlendState.blendAttachments[i].colorWriteMask = grfx::ColorComponentFlags::RGBA();
```
wangra-google, line 173: this line should be out of the switch so that for all blend mode we set the color mask to RGBA?

wangra-google: in that case you can probably remove it? we should assert if the blendmode is undefined

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -243,6 +249,7 @@ void GraphicsBenchmarkApp::Setup()
         createInfo.sampler                        = 5 * GetNumFramesInFlight();  // 1 for skybox, 3 for spheres, 1 for blit
         createInfo.sampledImage                   = 15 * GetNumFramesInFlight(); // 1 for skybox, 3 for spheres, 10 for quads, 1 for blit
```
wangra-google, line 250: better with 5 + kMaxTextureCount?

src/ppx/grfx/grfx_pipeline.cpp
```diff
@@ -169,10 +169,7 @@ void FillOutGraphicsPipelineCreateInfo(
         pDstCreateInfo->colorBlendState.blendAttachmentCount = pSrcCreateInfo->outputState.renderTargetCount;
         for (uint32_t i = 0; i < pDstCreateInfo->colorBlendState.blendAttachmentCount; ++i) {
             switch (pSrcCreateInfo->blendModes[i]) {
-                default: {
```
wangra-google: can you add an assert in the default case for unknown blend mode?

wangra-google: or would it be better to add a new BlendAttachmentState::BlendModeNone() so that all the paths are unified?

src/ppx/grfx/grfx_pipeline.cpp
```diff
@@ -12,16 +12,26 @@
...
 
 namespace ppx {
 namespace grfx {
 
 // -------------------------------------------------------------------------------------------------
 // BlendAttachmentState
 // -------------------------------------------------------------------------------------------------
+grfx::BlendAttachmentState BlendAttachmentState::BlendModeNone()
+{
+    grfx::BlendAttachmentState state = {};
```
wangra-google, line 29: it would be better to be explicit about the blend mode with `state.blendEnable = false;` so that it is clear that we are not doing any blending

src/ppx/grfx/grfx_pipeline.cpp
```diff
@@ -97,6 +107,21 @@ grfx::BlendAttachmentState BlendAttachmentState::BlendModePremultAlpha()
     return state;
 }
 
+grfx::BlendAttachmentState BlendAttachmentState::BlendModeDisableOutput()
+{
+    grfx::BlendAttachmentState state = {};
+    state.blendEnable                = false;
+    state.srcColorBlendFactor        = grfx::BLEND_FACTOR_ONE;
```
wangra-google, line 114: sorry my bad in my branch, but we dont need to set all the blendFactor/ops here since blend is disabled anyways

src/ppx/grfx/grfx_pipeline.cpp
```diff
@@ -153,7 +178,10 @@ void FillOutGraphicsPipelineCreateInfo(
         pDstCreateInfo->colorBlendState.blendAttachmentCount = pSrcCreateInfo->outputState.renderTargetCount;
         for (uint32_t i = 0; i < pDstCreateInfo->colorBlendState.blendAttachmentCount; ++i) {
             switch (pSrcCreateInfo->blendModes[i]) {
-                default: break;
+                case grfx::BLEND_MODE_NONE: {
```
wangra-google, line 181: might be better to assert for the default case for unknown blend mode

wangra-google (approved):

PR #509, opened 2024-08-20
Change some of the default values

RenfengLiu:


benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -185,31 +185,40 @@ void GraphicsBenchmarkApp::InitKnobs()
         pResolution->SetIndent(1);
     }
 
-    GetKnobManager().InitKnob(&pKnobAluCount, "alu-instruction-count", /* defaultValue = */ 100, /* minValue = */ 100, 400);
+    GetKnobManager().InitKnob(&pKnobAluCount, "alu-instruction-count", /* defaultValue = */ 0, /* minValue = */ 0, 500);
```
wangra-google, line 188: we have shaders with 0 instruction?

wangra-google: the name is confusing, maybe better to change the name to something that is more accurately reflects what it does?

wangra-google (approved):

PR #511, opened 2024-09-02
vk: fix sync masks for split compute/graphics.

Keenuts:
When the graphics queue was distinct from the compute queue, access & stage bits could be incorrectly computed. This was caused by the command type not being linked to each queue, and some invalid condition in an utility function designed to derive access & stage bits from an access type.

Tested on my local machine, all sampled except one (push_descriptor) passes. This failure already exists on main.

### DIFF
--- a/src/ppx/grfx/vk/vk_device.cpp
+++ b/src/ppx/grfx/vk/vk_device.cpp
@@ -499,6 +499,7 @@ Result Device::CreateQueues(const grfx::DeviceCreateInfo* pCreateInfo)
         uint32_t queueFamilyIndex = ToApi(pCreateInfo->pGpu)->GetGraphicsQueueFamilyIndex();
         for (uint32_t queueIndex = 0; queueIndex < pCreateInfo->graphicsQueueCount; ++queueIndex) {
             grfx::internal::QueueCreateInfo queueCreateInfo = {};
+            queueCreateInfo.commandType                     = grfx::COMMAND_TYPE_GRAPHICS;
             queueCreateInfo.queueFamilyIndex                = queueFamilyIndex;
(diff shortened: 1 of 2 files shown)


src/ppx/grfx/vk/vk_device.cpp
```diff
@@ -499,6 +499,7 @@ Result Device::CreateQueues(const grfx::DeviceCreateInfo* pCreateInfo)
         uint32_t queueFamilyIndex = ToApi(pCreateInfo->pGpu)->GetGraphicsQueueFamilyIndex();
         for (uint32_t queueIndex = 0; queueIndex < pCreateInfo->graphicsQueueCount; ++queueIndex) {
             grfx::internal::QueueCreateInfo queueCreateInfo = {};
+            queueCreateInfo.commandType                     = grfx::COMMAND_TYPE_GRAPHICS;
```
apazylbe, line 502: Do you know if this all works fine if we have a single queue family supporting graphics, compute and transfer? Do they all point to the same device queue?

apazylbe (approved):
Thank you!

src/ppx/grfx/vk/vk_device.cpp
Keenuts: Tried this on my desktop GPU (which has 1 graphics|compute queue + 1 compute queue), and with llvmpipe (a single graphics|compute queue).

On llvmpipe, the selected queue is the only one (makes sense), and for my real GPU, one queue for compute, and one for graphics. Transfer is shaders with graphics.

apazylbe: makes sense, thank you!

PR #521, opened 2024-10-03
Fix incorrect uniforms passing

jorgeag-google:
For the Quad render.
Uniforms (constant) were only passed correctly in the case of type noise.

### DIFF
--- a/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
+++ b/benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
@@ -1802,6 +1802,8 @@ void GraphicsBenchmarkApp::RecordCommandBufferFullscreenQuad(PerFrame& frame, si
     {
         mQuadPushConstant.InstCount = pKnobShaderAluLoopCount->GetValue();
         frame.cmd->PushGraphicsConstants(mQuadsPipelineInterfaces[0], GetPushConstCount(mQuadPushConstant.InstCount), &mQuadPushConstant.InstCount, offsetof(QuadPushConstant, InstCount) / sizeof(uint32_t));
+        frame.cmd->PushGraphicsConstants(mQuadsPipelineInterfaces[1], GetPushConstCount(mQuadPushConstant.InstCount), &mQuadPushConstant.InstCount, offsetof(QuadPushConstant, InstCount) / sizeof(uint32_t));
+        frame.cmd->PushGraphicsConstants(mQuadsPipelineInterfaces[2], GetPushConstCount(mQuadPushConstant.InstCount), &mQuadPushConstant.InstCount, offsetof(QuadPushConstant, InstCount) / sizeof(uint32_t));
     }
     switch (pFullscreenQuadsType->GetValue()) {
         case FullscreenQuadsType::FULLSCREEN_QUADS_TYPE_NOISE: {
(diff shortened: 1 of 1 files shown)


benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -1824,12 +1826,12 @@ void GraphicsBenchmarkApp::RecordCommandBufferFullscreenQuad(PerFrame& frame, si
             float3 colorValues = pFullscreenQuadsColor->GetValue();
             colorValues *= intensity;
             mQuadPushConstant.ColorValue = colorValues;
-            frame.cmd->PushGraphicsConstants(mQuadsPipelineInterfaces[0], GetPushConstCount(mQuadPushConstant.ColorValue), &mQuadPushConstant.ColorValue, offsetof(QuadPushConstant, ColorValue) / sizeof(uint32_t));
+            frame.cmd->PushGraphicsConstants(mQuadsPipelineInterfaces[1], GetPushConstCount(mQuadPushConstant.ColorValue), &mQuadPushConstant.ColorValue, offsetof(QuadPushConstant, ColorValue) / sizeof(uint32_t));
```
wangra-google, line 1829: nice catch, maybe it would be clearer to use FullscreenQuadsType::FULLSCREEN_QUADS_TYPE_SOLID_COLOR here instead of 1 or maybe pFullscreenQuadsType->GetValue()? (same as the other places you've fixed)

benchmarks/graphics_pipeline/GraphicsBenchmarkApp.cpp
```diff
@@ -1824,13 +1826,12 @@ void GraphicsBenchmarkApp::RecordCommandBufferFullscreenQuad(PerFrame& frame, si
             float3 colorValues = pFullscreenQuadsColor->GetValue();
             colorValues *= intensity;
             mQuadPushConstant.ColorValue = colorValues;
-            frame.cmd->PushGraphicsConstants(mQuadsPipelineInterfaces[0], GetPushConstCount(mQuadPushConstant.ColorValue), &mQuadPushConstant.ColorValue, offsetof(QuadPushConstant, ColorValue) / sizeof(uint32_t));
+            frame.cmd->PushGraphicsConstants(mQuadsPipelineInterfaces[int(pFullscreenQuadsType->GetValue())], GetPushConstCount(mQuadPushConstant.ColorValue), &mQuadPushConstant.ColorValue, offsetof(QuadPushConstant, ColorValue) / sizeof(uint32_t));
```
wangra-google, line 1829: nit: you can probably declare the enum with a type of uint32_t so that you don't need the cast:

enum class FullscreenQuadsType: uint32_t
{
FULLSCREEN_QUADS_TYPE_NOISE,
FULLSCREEN_QUADS_TYPE_SOLID_COLOR,
FULLSCREEN_QUADS_TYPE_TEXTURE
};

jorgeag-google: I have just tried and it does not seem to work.
I did a quick Google serach, and some pages says that even if you specify the undelying type, you need cast to use as index.

wangra-google: ah, thanks for trying this, good to know, and sorry for the wrong information. 
also better to use `static_cast` for such casting those as suggested by the cpp styple guide: go/cstyle#Casting

wangra-google (approved):

PR #534, opened 2024-11-20
Add ALU ocuppancy Knob

jorgeag-google:
Adding the knob for the ALU ocuppancy test


assets/benchmarks/shaders/Benchmark_Texture.hlsl
```diff
@@ -20,13 +20,16 @@ RWStructuredBuffer<float> dataBuffer : register(u11);
 float4 psmain(VSOutputPos input) : SV_TARGET
```
wangra-google, line 20: what is the wave occupancy of this shader? Could you add some comments that indicate the occupancy in this shader? (same for above shaders just to be consistent, thanks :) )

wangra-google (approved):

PR #535, opened 2024-12-06
scale the spheres

RenfengLiu:


benchmarks/graphics_pipeline/SphereMesh.cpp
```diff
@@ -53,7 +53,7 @@ float4x4 OrderedGrid::GetModelMatrix(uint32_t sphereIndex, bool isXR) const
         z *= -1.0;
     }
 
-    return glm::translate(float3(x * mStep, y * mStep, z * mStep));
+    return glm::translate(float3(x * mStep, y * mStep, z * mStep)) * glm::scale(float3(5.f, 5.f, 5.f));
```
wangra-google, line 56: is 5 large enough?

Keenuts (approved):

benchmarks/graphics_pipeline/SphereMesh.cpp
wangra-google: i feel 5 is still too small as the spheres only covers less than 1/9 of the screen. were you able to see diff in the alpha blending test with scale factor of 5?

wangra-google: my worry is that with 1/9 overage, some difference might be less than noise. how about 20? (so that at least we covers about ~2/3 screen)

PR #536, opened 2024-12-06
Provide a camera if there's none in the scene

### LINKED ISSUE #456: [glTF] Provide a camera if the scene doesn't have one
Scenes are not required to have a camera (see [5.17](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-gltf)). When a scene does not provide a camera, we should provide a suitable default.

footballhead:
Allow loading scenes without a camera. If there is no camera then provide one that the user can interact with. As someone validating GLTF support, I find this preferable to a static camera so that I can look at the scene from any angle at runtime.

Fixes #456

### DIFF
--- a/projects/gltf_basic_materials/GltfBasicMaterials.cpp
+++ b/projects/gltf_basic_materials/GltfBasicMaterials.cpp
@@ -16,6 +16,8 @@
 #include "ppx/scene/scene_gltf_loader.h"
 #include "ppx/graphics_util.h"
 
+namespace {
+
 using namespace ppx;
 
 #if defined(USE_DX12)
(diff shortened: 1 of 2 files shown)


projects/gltf_basic_materials/GltfBasicMaterials.cpp
```diff
@@ -331,3 +373,39 @@ void GltfBasicMaterialsApp::InitKnobs()
     GetKnobManager().InitKnob(&mSceneAssetKnob, "gltf-scene-asset", "scene_renderer/scenes/tests/gltf_test_basic_materials.glb");
     mSceneAssetKnob->SetFlagDescription("GLTF asset to load and render");
 }
+
+void GltfBasicMaterialsApp::MouseMove(int32_t x, int32_t y, int32_t dx, int32_t dy, uint32_t buttons)
+{
+    if (!mDefaultCamera) {
```
apazylbe, line 379: Does this mean you can only move when the scene doesn't have a camera? What happens if the scene does have one?

apazylbe (approved):

projects/gltf_basic_materials/GltfBasicMaterials.cpp
footballhead: > Does this mean you can only move when the scene doesn't have a camera?

Yes. This is largely modeled after what the [gltf-Sample-Viewer](https://github.com/KhronosGroup/glTF-Sample-Viewer) does. E.g. [ToyCar](https://github.khronos.org/glTF-Sample-Viewer-Release/?model=https://raw.GithubUserContent.com/KhronosGroup/glTF-Sample-Assets/main/./Models/ToyCar/glTF-Binary/ToyCar.glb) has both a camera added by the sample viewer (which has user controls) as well as several cameras  loaded from the GLTF file (which are static).

> What happens if the scene does have one?

The loader will create all cameras from GLTF scene and gltf_basic_materials will arbitrarily prefer the first one. This will not have user controls, mostly because [GltfSceneLoader::LoadNodeInternal](https://github.com/google/bigwheels/blob/main/src/ppx/scene/scene_gltf_loader.cpp#L1983) could create an OrthoCamera or PerspCamera; both of these classes lack the ArcballCamera controls.

We could consider adding a dropdown for the user to choose which camera they want (similar to the Khronos gltf-Sample-Viewer).

PR #542, opened 2024-12-18
Introduce a script to test glTF support

footballhead:
We can use glTF-Sample-Assets to verify our level of glTF support.

test_gltf_sample_assets.py iterates through all glTF-Sample-Assets and tries to render them.

make_gltf_sample_assets_report.py will take the output of test_gltf_sample_assets.py and try to make a report website for triaging.

generate_gltf_report.sh handles the entire process end-to-end


tools/test_gltf_sample_assets.py
```diff
@@ -0,0 +1,111 @@
...
+  with args.model_index.open('r') as model_index_file:
+    model_index = json.load(model_index_file)
+
+  os.mkdir(args.output)
+
+  # Dump some state of the test environment to be included in the report.
+  with (args.output / 'meta.json').open('w') as meta_file:
+    json.dump({'host': str(socket.getfqdn()),
+               'datetime': str(datetime.datetime.now()),
+               'sha': _get_git_head_commit()}, meta_file)
```
apazylbe, line 95: This is the BigWheels commit? Is it worth logging/reporting glTF-Sample-Assets commit as well?

apazylbe (approved):

tools/test_gltf_sample_assets.py
footballhead: Good point, I've updated the report to include both. PTAL

PR #543, opened 2025-01-17
Remove VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT bit from image creation.

RenfengLiu:
No sample mismatch the image format and imageview format, meaning this feature is not used. In addition, specifying this bit prevents the driver to make some assumptions, and this can degrade performances


Keenuts (approved):
LGTM, but please add context in the description so we know why in the future, something like:
"No sample mismatch the image format and imageview format, meaning this feature is not used. In addition, specifying this bit prevents the driver to make some assumptions, and this can degrade performances".

wangra-google (approved):

PR #546, opened 2025-01-29
scale the sphere according to the number of spheres.

RenfengLiu:
The fixed 20.f scale makes the sphere out of scence if there's only 1 sphere.


jorgeag-google (approved):
Yes, I noticed that last week. Thanks for fixing it.

PR #547, opened 2025-01-30
fix a problem that quad alpha blend not working

RenfengLiu:


wangra-google (approved):

assets/benchmarks/shaders/Benchmark_SolidColor.hlsl
```diff
@@ -16,7 +16,7 @@
 
 float4 psmain(VSOutputPos input) : SV_TARGET
 {
-    float4 color = float4(Config.ColorValue, 1.0f);
-    color.a = randomCompute(Config.InstCount, input.position); 
+    float4 color = float4(Config.ColorValue, 0.5f);
+    color.b = randomCompute(Config.InstCount, input.position);
```
wangra-google, line 20: do you want to randomize color? my worry is that this might potentially change ubwc compression

PR #549, opened 2025-03-14
Enable support for AndroidXR

RenfengLiu:


src/android/AndroidManifest.XR.xml
```diff
@@ -27,6 +27,11 @@
           android:required="false"/>
     <!-- -->
 
+    <!-- Needed for AndroidXR -->
+    <uses-native-library 
+          android:name="libopenxr.google.so" 
+          android:required="false"/>
```
Keenuts, line 33: Shouldn't this be set to `true` if the library is required?

Keenuts: Ah right, this manifest is also used for Meta/Quest/Horizon thingy

Keenuts (approved):

PR #552, opened 2025-04-09
build: fix warning in upstream GLI

Keenuts:
This warning seems to only impact clang.


Keenuts:
Looks like this is on a recent-enough clang version. Sent a patch upstream to fix this

apazylbe (approved):

PR #558, opened 2025-05-23
Format with clang-format-16

footballhead:
After 976e521838f1538ec7db33deaeba82508b36f05b and talking to @angela28chen, I got thinking about formatting the code base. Turns out:

1. We don't need the clang-format-14 install in the workflow since that's handled by DoozyX/clang-format-lint-action
2. We've configured DoozyX/clang-format-lint-action to use clang-format-16 anyway. Turns out that there were parts of the code base that weren't formatted.

Command run:

```
find . ! -ipath './third_party/*' -a ! -ipath './.cxx/*' -a "(" -iname '*.h' -o -iname '*.cpp' ")" | xargs -P 0 clang-format-16 -i
```

### DIFF
--- a/.github/workflows/check-formatting.yml
+++ b/.github/workflows/check-formatting.yml
@@ -15,6 +15,9 @@ jobs:
     name: clang-format
     runs-on: "ubuntu-24.04"
     steps:
+      - name: Setup clang-format
+        run: |
+          sudo apt-get install -yqq clang-format-14
       - name: Checkout repository
         uses: actions/checkout@v3
         with:
--- a/.github/workflows/check-formatting.yml
+++ b/.github/workflows/check-formatting.yml
@@ -15,9 +15,6 @@ jobs:
     name: clang-format
     runs-on: "ubuntu-24.04"
     steps:
-      - name: Setup clang-format
-        run: |
-          sudo apt-get install -yqq clang-format-14
       - name: Checkout repository
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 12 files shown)


footballhead:
Hmm I think I misread the code formatting check. I thought it was running the Ubuntu clang-format-14 package but instead it's using DoozyX/clang-format-lint-action. We may not need this PR and might instead need to clean up check-formatting.yml

footballhead:
> Hmm I think I misread the code formatting check. I thought it was running the Ubuntu clang-format-14 package but instead it's using DoozyX/clang-format-lint-action. We may not need this PR and might instead need to clean up check-formatting.yml

Ok I've retooled this based on how I understand the workflow to be working

apazylbe (approved):

footballhead:
@angela28chen

> LGTM, one or two questions:
> 
> 1. Any particular reason that we use clang-format 16 for BigWheels? Is that easy/hard to set up for people working mainly on Windows platform?

I blamed the workflow YAML and ended up at #409 . Seems like it was chosen since it was the version installed by Visual Studio at the time, and "doing custom formatter on MSVC is a pain". I don't have any strong opinions about what version we should use so would be open to revisiting this in the future.

(FWIW I just installed VS 2022 and got clang-format 19.1.5)

> 2. Just curious but do you know if DoozyX/clang-format-lint-action does anything to validate the .clang-format file?

I don't think so. Looking at github.com/DoozyX/clang-format-lint-action, 99% of the logic is in run-clang-format.py; I glanced at the file and can't see anything to suggest that it touches .clang-format.

footballhead:
Dive recently switched to clang-format 18.1.8. Should we do the same to remain consistent?

footballhead:
I'm going to try running 18.1.8 to see what changes it suggests

footballhead:
Closing for now since what I wanted to do, #561, is merged. I will revisit this in the future when i have time.

PR #565, opened 2025-06-23
Safely reuse Present wait semaphore

### LINKED ISSUE #562: Validation layer reports VUID-vkQueueSubmit-pSignalSemaphores-00067 error
I just compiled BigWheels on Windows and tried running vk_sample_01_triangle.exe. It rendered one frame, then I was greeted with:

```
*** VULKAN VALIDATION ERROR MESSAGE ***
Severity : ERROR
Type     : [VALIDATION]
           [1]: <UNNAMED OBJECT>
Message  : vkQueueSubmit(): pSubmits[0].pSignalSemaphores[0] (VkSemaphore 0x230000000023) is being signaled by VkQueue 0x1acbe8fac60, but it may still be in use by VkSwapchainKHR 0x40000000004.
Here are the most recently acquired image indices: [0], 1.
(brackets mark the last use of VkSemaphore 0x230000000023 in a presentation operation)
Swapchain image 0 was presented but was not re-acquired, so VkSemaphore 0x230000000023 may still be in use and cannot be safely reused with image index 1.
Vulkan insight: One solution is to assign each image its own semaphore. Here are some common methods to ensure that a semaphore passed to vkQueuePresentKHR is not in use and can be safely reused:
        a) Use a separate semaphore per swapchain image. Index these semaphores using the index of the acquired image.
        b) Consider the VK_EXT_swapchain_maintenance1 extension. It allows using a VkFence with the presentation operation.
The Vulkan spec states: Each binary semaphore element of the pSignalSemaphores member of any element of pSubmits must be unsignaled when the semaphore signal operation it defines is executed on the device (https://vulkan.lunarg.com/doc/view/1.4.313.0/windows/antora/spec/latestchapters/cmdbuffers.html#VUID-vkQueueSubmit-pSignalSemaphores-00067)

Assertion failed: false, file C:\Users\hitchens\Desktop\git\bigwheels\src\ppx\grfx\vk\vk_instance.cpp, line 115
````

Since Debug builds treat this as fatal, the app immediately exits.

Log:  [ppx.log](https://github.com/user-attachments/files/20842781/ppx.log)

NOTE: This is a new check added in 1.4.313.0:

> A new validation error was added that will likely catch many applications not using semaphores in swapchains correctly. For details on how to fix it in your application see the [Vulkan Guide](https://github.com/KhronosGroup/Vulkan-Guide/blob/main/chapters/swapchain_semaphore_reuse.adoc)
>
> \- [1.4.313.0 release notes](https://vulkan.lunarg.com/doc/sdk/1.4.313.0/windows/release_notes.html)

### PR DESCRIPTION
**Pull request by footballhead**
A new check was added in SDK 1.4.313 related to the use of semaphores during presentation:

```
*** VULKAN VALIDATION ERROR MESSAGE ***
Severity : ERROR
Type     : [VALIDATION]
           [1]: <UNNAMED OBJECT>
Message  : vkQueueSubmit(): pSubmits[0].pSignalSemaphores[0] (VkSemaphore 0x230000000023) is being signaled by VkQueue 0x1acbe8fac60, but it may still be in use by VkSwapchainKHR 0x40000000004.
Here are the most recently acquired image indices: [0], 1.
(brackets mark the last use of VkSemaphore 0x230000000023 in a presentation operation)
Swapchain image 0 was presented but was not re-acquired, so VkSemaphore 0x230000000023 may still be in use and cannot be safely reused with image index 1.
Vulkan insight: One solution is to assign each image its own semaphore. Here are some common methods to ensure that a semaphore passed to vkQueuePresentKHR is not in use and can be safely reused:
        a) Use a separate semaphore per swapchain image. Index these semaphores using the index of the acquired image.
        b) Consider the VK_EXT_swapchain_maintenance1 extension. It allows using a VkFence with the presentation operation.
The Vulkan spec states: Each binary semaphore element of the pSignalSemaphores member of any element of pSubmits must be unsignaled when the semaphore signal operation it defines is executed on the device (https://vulkan.lunarg.com/doc/view/1.4.313.0/windows/antora/spec/latestchapters/cmdbuffers.html#VUID-vkQueueSubmit-pSignalSemaphores-00067)

Assertion failed: false, file C:\Users\hitchens\Desktop\git\bigwheels\src\ppx\grfx\vk\vk_instance.cpp, line 115
```

Since we wait on the `vkAcquireImageKHR` fence and the fence only guarantees that the just-acquired image index is ready, we need one semaphore for each image. As suggested in
https://docs.vulkan.org/guide/latest/swapchain_semaphore_reuse.html, I created one wait semaphore for each swapchain image.

I fixed most of this by find and replace in my IDE. If there were fewer projects, I'd store the semaphore in a local variable then reference it multiple times, but this requires manual intervention. While not optimal, I don't think this impacts performance significantly.

Fixes #562

### DIFF
--- a/benchmarks/compute_operations/main.cpp
+++ b/benchmarks/compute_operations/main.cpp
@@ -47,7 +47,6 @@ private:
         ppx::grfx::CommandBufferPtr cmd;
         ppx::grfx::SemaphorePtr     imageAcquiredSemaphore;
         ppx::grfx::FencePtr         imageAcquiredFence;
-        ppx::grfx::SemaphorePtr     renderCompleteSemaphore;
         ppx::grfx::FencePtr         renderCompleteFence;
         ppx::grfx::QueryPtr         timestampQuery;
     };
@@ -171,7 +170,6 @@ void ProjApp::Setup()
         grfx::FenceCreateInfo fenceCreateInfo = {};
         PPX_CHECKED_CALL(GetDevice()->CreateFence(&fenceCreateInfo, &frame.imageAcquiredFence));
 
-        PPX_CHECKED_CALL(GetDevice()->CreateSemaphore(&semaCreateInfo, &frame.renderCompleteSemaphore));
 
         fenceCreateInfo = {true}; // Create signaled
         PPX_CHECKED_CALL(GetDevice()->CreateFence(&fenceCreateInfo, &frame.renderCompleteFence));
@@ -449,12 +447,12 @@ void ProjApp::Render()
     submitInfo.waitSemaphoreCount   = 1;
     submitInfo.ppWaitSemaphores     = &frame.imageAcquiredSemaphore;
     submitInfo.signalSemaphoreCount = 1;
-    submitInfo.ppSignalSemaphores   = &frame.renderCompleteSemaphore;
+    submitInfo.ppSignalSemaphores   = &GetSwapchain()->GetPresentationReadySemaphore(imageIndex);
     submitInfo.pFence               = frame.renderCompleteFence;
 
     PPX_CHECKED_CALL(GetGraphicsQueue()->Submit(&submitInfo));
 
-    PPX_CHECKED_CALL(swapchain->Present(imageIndex, 1, &frame.renderCompleteSemaphore));
+    PPX_CHECKED_CALL(swapchain->Present(imageIndex, 1, &GetSwapchain()->GetPresentationReadySemaphore(imageIndex)));
     if (GetFrameCount() > 0) {
         uint64_t frequency = 0;
         GetGraphicsQueue()->GetTimestampFrequency(&frequency);
--- a/benchmarks/compute_operations/main.cpp
+++ b/benchmarks/compute_operations/main.cpp
@@ -170,7 +170,6 @@ void ProjApp::Setup()
         grfx::FenceCreateInfo fenceCreateInfo = {};
         PPX_CHECKED_CALL(GetDevice()->CreateFence(&fenceCreateInfo, &frame.imageAcquiredFence));
 
-
         fenceCreateInfo = {true}; // Create signaled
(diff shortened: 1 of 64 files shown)

### CODE REVIEW

**Comment by footballhead** [comment]
I'm slightly confused because of these passages in the Vulkan guide chapter:

> You can probably see that there is nothing special with reusing present wait semaphores safely. Call vkAcquireNextImageKHR to get the image index. Wait on the semaphore from vkAcquireNextImageKHR in one of vkQueueSubmit batches (you can also wait on the fence, but that introduces additional host sync point). After the wait it’s safe to reuse semaphores
>
> [...]
>
> Often, to synchronize with in-flight frames, the application uses a vkQueueSubmit fence. By waiting on that fence, we know that the corresponding command buffers and other frame resources are no longer in use. However, the Vulkan specification does not guarantee that waiting on a vkQueueSubmit fence also synchronizes presentation operations. The reuse of presentation resources should rely on vkAcquireNextImageKHR or additional extensions (will be mentioned at the end of this article), rather than on vkQueueSubmit fences.

We wait on both Submit (renderCompleteFence) and Acquire (imageAcquiredFence) fences but still encounter the validation error. Is there something more fundamentally wrong with the samples?

**Review by apazylbe — APPROVED** [review]

File: benchmarks/compute_operations/main.cpp [hunk 1]
```diff
@@ -449,12 +446,12 @@ void ProjApp::Render()
     submitInfo.waitSemaphoreCount   = 1;
     submitInfo.ppWaitSemaphores     = &frame.imageAcquiredSemaphore;
     submitInfo.signalSemaphoreCount = 1;
-    submitInfo.ppSignalSemaphores   = &frame.renderCompleteSemaphore;
+    submitInfo.ppSignalSemaphores   = &GetSwapchain()->GetPresentationReadySemaphore(imageIndex);
```
footballhead, line 449: FYI that I fixed most of this by find and replace in my IDE. If there were fewer projects, I'd store this once as a variable then reference it multiple times, but this required some manual fixing. I don't think this impacts performance significantly.

PR #567, opened 2025-06-27
cubexr: fix 32bit build

Keenuts:
The template inference fails on 32bit builds because of sizeof. Making the templated type explicit solves this.

### DIFF
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -35,4 +35,4 @@ jobs:
           echo "VULKAN_SDK=${HOME}/vulkan-sdk/1.3.268.0/x86_64" >> $GITHUB_ENV
       - name: Build all projects
         run: |
-          ./gradlew buildRelease
+          ./gradlew buildRelease -PENABLE_32BIT_ARM=true
--- a/build.gradle
+++ b/build.gradle
@@ -93,7 +93,7 @@ subprojects {
                 appLabel : "vk_${project.name}"
             ]
 
-            ndk.abiFilters = ['arm64-v8a']
+            ndk.abiFilters = ['armeabi-v7a', 'arm64-v8a']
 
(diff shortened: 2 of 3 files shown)


Keenuts:
> How are you testing 32 bit build? Should this become a GitHub action?

Internal android build config for the repro. I tried adding a 32-bit arm ABI in the filter list and removed the fix.  Let's see if the existing android build fails.

Keenuts:
> > How are you testing 32 bit build? Should this become a GitHub action?
> 
> Internal android build config for the repro. I tried adding a 32-bit arm ABI in the filter list and removed the fix. Let's see if the existing android CI fails.

footballhead:
Thanks! Looks like it heavily impact the build time, which I guess is to be expected. but it went from ~14 minutes to ~26 minutes. Are we ok with this jump? https://github.com/google/bigwheels/actions/workflows/android-build.yml

apazylbe:
> > Thanks! Looks like it heavily impact the build time, which I guess is to be expected. but it went from ~14 minutes to ~26 minutes. Are we ok with this jump? https://github.com/google/bigwheels/actions/workflows/android-build.yml
> 
> Github CI time is IMO not that important for now (given review turn-around is usually >30mn). But it does increase local build-time. I don't work daily on the Android BW so not the best person to ask, but if somebody is, a 10mn increase is a lot.
> 
> Now, this kind of issue doesn't happen that often, and is caught by internal CI, so we could go without the additional ABI

It looks like we could set up a property for the list of ABIs, read the property in `build.gradle`, set the ABIs based on the property if one is provided in build.gradle, and set the property from the `./gradlew` command to just include this in the CI.
I think including the additional ABI by default is probably not worth it, since it not only can increase build time but also increase the APK size since libraries are duplicated. Not sure if the property is worth the hassle/complexity, probably not.

Keenuts:
Added a new property and enabled it on the Android github ci, was not much and will only impact the CI by default.

PR #610, opened 2026-02-20
[glTF] Removed unused GltfLoader methods

footballhead:
Adding new features to GltfLoader is difficult for two main reasons:

1. Many methods are untested. This makes it hard to understand what they do and know how to modify them correctly.
2. ResourceManager is optional. This creates two code paths through GltfLoader, greatly increasing cognitive burden.

Since only LoadScene is used and LoadScene creates a ResourceManager, I suggest to:

1. Remove the unused public methods
3. Require a ResourceManager

This should simplify the class, making it easier to modify and maintain.

### DIFF
--- a/src/ppx/scene/scene_gltf_loader.cpp
+++ b/src/ppx/scene/scene_gltf_loader.cpp
@@ -637,6 +637,9 @@ void GltfLoader::CalculateMeshMaterialVertexAttributeMasks(
             // Get required vertex attributes
             auto requiredVertexAttributes = pMaterialFactory->GetRequiredVertexAttributes(materialIdent);
 
+            // Disable vertex colors for now: some work is needed to handle format conversion.
+            requiredVertexAttributes.bits.colors = false;
+
             // OR the masks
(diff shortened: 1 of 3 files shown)


src/ppx/scene/scene_gltf_loader.cpp
```diff
@@ -1349,26 +1303,31 @@ ppx::Result GltfLoader::LoadMeshData(
...
+        //
+        hasCachedGeometry = true;
     }
 
     // ---------------------------------------------------------------------------------------------
 
+    auto requiredVertexAttributesIter = loadParams.pMeshMaterialVertexAttributeMasks->find(pGltfMesh);
+    if (requiredVertexAttributesIter == loadParams.pMeshMaterialVertexAttributeMasks->end()) {
+        PPX_LOG_ERROR("pMeshMaterialVertexAttributeMasks missing entry for mesh" << pGltfMesh);
+        return ERROR_UNEXPECTED_NULL_ARGUMENT; // TODO better return code
```
footballhead, line 1321: Address TODO

PR #611, opened 2026-02-27
[glTF] Load data-URI images

### LINKED ISSUE #454: [glTF] Load embedded textures with data URI scheme
Using `scene::GltfLoader`, some models in [glTF-Sample-Assets](https://github.com/KhronosGroup/glTF-Sample-Assets) fail to load because the texture is embedded with the [data URI scheme](https://en.wikipedia.org/wiki/Data_URI_scheme).

From the spec [5.18.1](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#_image_uri):

> Instead of referencing an external file, [image.uri] MAY contain a data:-URI

This would allow us to load more models like:

- [DamagedHelmet (glTF-Embedded variant)](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/DamagedHelmet)
- [Duck (glTF-Embedded variant)](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/Duck)
- [BoxTextured (glTF-Embedded variant)](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/BoxTextured)

footballhead:
Image data can be embedded into the glTF file by using the data-URI scheme. Decode and load these images.

Fixes #454

### DIFF
--- a/src/ppx/scene/scene_gltf_loader.cpp
+++ b/src/ppx/scene/scene_gltf_loader.cpp
@@ -12,9 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <memory>
 #include <string_view>
 
 #include "ppx/scene/scene_gltf_loader.h"
+#include "ppx/config.h"
 #include "ppx/grfx/grfx_device.h"
 #include "ppx/grfx/grfx_scope.h"
 #include "ppx/graphics_util.h"
@@ -239,25 +241,6 @@ static VertexAccessors GetVertexAccessors(const cgltf_primitive* pGltfPrimitive)
     return accessors;
 }
 
-// Get a buffer view's start address
-static const void* GetStartAddress(
-    const cgltf_buffer_view* pGltfBufferView)
-{
-    //
-    // NOTE: Don't assert in this function since any of the fields can be NULL for different reasons.
-    //
-
-    if (IsNull(pGltfBufferView) || IsNull(pGltfBufferView->buffer) || IsNull(pGltfBufferView->buffer->data)) {
-        return nullptr;
-    }
-
-    const size_t bufferViewOffset = static_cast<size_t>(pGltfBufferView->offset);
-    const char*  pBufferAddress   = static_cast<const char*>(pGltfBufferView->buffer->data);
-    const char*  pDataStart       = pBufferAddress + bufferViewOffset;
-
-    return static_cast<const void*>(pDataStart);
-}
-
 const char* ToString(cgltf_component_type componentType)
 {
     switch (componentType) {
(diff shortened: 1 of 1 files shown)


src/ppx/scene/scene_gltf_loader.cpp
```diff
@@ -443,6 +426,141 @@ std::vector<glm::float4> UnpackFloat4s(const cgltf_accessor& accessor)
...
+DataUriResult ParseDataUriWithBase64Extension(char* data, size_t length)
+{
+    std::string_view dataView(data, length);
+
+    size_t dataSchemeStartPosition = dataView.find(kDataScheme);
+    if (dataSchemeStartPosition != 0) {
+        PPX_LOG_ERROR("data scheme should be at the start of the URI");
+        return {ppx::ERROR_BAD_DATA_SOURCE};
+    }
+    size_t dataSchemeEndPosition = dataSchemeStartPosition + kDataScheme.size();
```
Keenuts, line 448: This should be a constexpr no? (Given start must be 0)

src/ppx/scene/scene_gltf_loader.cpp
```diff
@@ -443,6 +426,141 @@ std::vector<glm::float4> UnpackFloat4s(const cgltf_accessor& accessor)
...
+
+    size_t base64ExtensionStartPosition = dataView.find(kBase64Extension, dataSchemeEndPosition);
+    if (base64ExtensionStartPosition == kBase64Extension.npos) {
```
Keenuts, line 451:
```suggestion
    if (base64ExtensionStartPosition == std::string_view::npos) {
```

No?

src/ppx/scene/scene_gltf_loader.cpp
```diff
@@ -443,6 +426,141 @@ std::vector<glm::float4> UnpackFloat4s(const cgltf_accessor& accessor)
...
+        case (cgltf_result_io_error): return "cgltf_result_io_error";
+        case (cgltf_result_out_of_memory): return "cgltf_result_out_of_memory";
+        case (cgltf_result_legacy_gltf): return "cgltf_result_legacy_gltf";
+        default: return "<unknown cgltf_result>";
+    }
+}
+
+int CountBase64Padding(std::string_view base64)
+{
+    PPX_ASSERT_MSG(base64.size() % 4 == 0, "bsae64 string length must be a multiple of 4");
```
Keenuts, line 481:
```suggestion
    PPX_ASSERT_MSG(base64.size() % 4 == 0, "base64 string length must be a multiple of 4");
```

src/ppx/scene/scene_gltf_loader.cpp
```diff
@@ -443,6 +426,141 @@ std::vector<glm::float4> UnpackFloat4s(const cgltf_accessor& accessor)
...
+{
+    ppx::Result            result = ppx::SUCCESS;
+    std::vector<std::byte> data;
+};
+
+DecodeBase64Result DecodeBase64(char* base64, size_t base64Size)
+{
+    std::string_view base64StringView(base64, base64Size);
+
+    // cgltf_load_buffer_base64 doesn't handlke padding characters, so hack around it.
```
Keenuts, line 503:
```suggestion
    // cgltf_load_buffer_base64 doesn't handle padding characters, so hack around it.
```

src/ppx/scene/scene_gltf_loader.cpp
```diff
@@ -443,6 +426,141 @@ std::vector<glm::float4> UnpackFloat4s(const cgltf_accessor& accessor)
...
+    std::vector<std::byte> decodedData(decodedSize, std::byte{0});
+
+    // cgltf_load_buffer_base64 allocates once and only frees on error.
+    // Instead of mallocing new memory, use the vector.
+    cgltf_options cgltfOptions     = {};
+    cgltfOptions.memory.alloc_func = +[](void* user, size_t size) -> void* {
+        return reinterpret_cast<std::vector<std::byte>*>(user)->data();
+    };
+    cgltfOptions.memory.free_func = +[](void* /*user*/, void* /*ptr*/) {};
+    cgltfOptions.memory.user_data = &decodedData;
```
Keenuts, line 525: I feel this is super wrong. It works now, but relies on cgltf now changing its internal implementation.

base64 decoding is not that complex. Should we provide out own b64 decode function handling padding instead? Seems like those hacks are as complex as a b64 implem.

src/ppx/scene/scene_gltf_loader.cpp
```diff
@@ -443,6 +426,141 @@ std::vector<glm::float4> UnpackFloat4s(const cgltf_accessor& accessor)
...
+    std::vector<std::byte> data;
+};
+
+DecodeBase64Result DecodeBase64(char* base64, size_t base64Size)
+{
+    std::string_view base64StringView(base64, base64Size);
+
+    // cgltf_load_buffer_base64 doesn't handlke padding characters, so hack around it.
+    // Padding can be treated as 0 (base64 A), Then, after decode, drop bytes to achieve the correct decoded length.
+    int paddingCount = CountBase64Padding(base64StringView);
```
Keenuts, line 505: If the base64 content length is zero, an empty stringview is passed to CountBase64Padding. If this happens, you deref past-the-end.

URI: `data:image/png;base64,`
I think in this case, the parsing succeeds, but returns 0 as b64 content length.

Keenuts (commented):
Thanks for the contribution! I think the best option given CGLTF's b64 decoding is incomplete is to roll your own. B64 is simple enough that it should be the same complexity as all the code you need to hack around to make the CGLTF one work.

footballhead:
Thanks for the review. This isn't currently a priority. Closing for now.

PR #615, opened 2026-04-22
Fix asset lookup for multi-config generators

### LINKED ISSUE #563: Can't find assets using "Ninja Multi-config" CMake generator
```
$ cmake -B build -G "Ninja Multi-config" .
$ cmake --build build --config Debug
$ ./build/bin/Debug/vk_sample_01_cube

...

*** PPX ASSERT ***
Message   : Could not determine asset path for "basic/shaders/spv/StaticVertexColors.vs.spv"
Condition : !assetPath.empty()
Function  : GetAssetPath
Location  : /usr/local/google/home/hitchens/git/bigwheels/src/ppx/base_application.cpp : 102

vk_sample_01_triangle: /usr/local/google/home/hitchens/git/bigwheels/src/ppx/base_application.cpp:102: std::filesystem::__cxx11::path ppx::BaseApplication::GetAssetPath(const std::filesystem::__cxx11::path&) const: Assertion `false' failed.
```

This works fine with the normal Ninja generator

AnouarMohamed:
## Problem

BigWheels computes `RELATIVE_PATH_TO_PROJECT_ROOT` from the runtime output
directory so applications can find `assets/` at runtime.

That logic only treated Visual Studio as a multi-config generator. With
generators like `Ninja Multi-Config`, executables are also emitted under
`bin/<Config>/`, so the computed relative path is too shallow and samples
fail to locate assets at runtime.

Fixes https://github.com/google/BigWheels/issues/563

## Solution

Use CMake's `GENERATOR_IS_MULTI_CONFIG` global property instead of matching
only `CMAKE_GENERATOR` against Visual Studio.

This keeps the existing single-config behavior unchanged and correctly
accounts for all multi-config generators when deriving the path back to
the build root.

## Testing

I could not run a full configure/build in this environment because `cmake`
was not installed.

I did verify the path math for the affected layouts:
- single-config `bin` resolves back to the build root with `..`
- multi-config `bin/Debug` resolves back to the build root with `../..`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -183,9 +183,10 @@ set(PPX_3P_ASSET_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/third_party/assets")
 
 # Determine relative path from binary location to the build root.
 cmake_path(SET BUILD_DIRECTORY NORMALIZE "${CMAKE_BINARY_DIR}")
-if (CMAKE_GENERATOR MATCHES "Visual Studio")
-    # Visual Studio doesn't follow CMAKE_RUNTIME_OUTPUT_DIRECTORY, but
-    # creates a subdirectory with the config name.
+get_property(PPX_GENERATOR_IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
+if (PPX_GENERATOR_IS_MULTI_CONFIG)
+    # Multi-config generators place executables in a per-config subdirectory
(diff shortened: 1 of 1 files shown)


Keenuts:
Hi,

Change LGTM, thanks for the contribution. Once you've signed the CLA we should be able to merge this.

AnouarMohamed:
Just signed the CLA and triggered a new CI run, should be good now!

AnouarMohamed:
any updates about this PR ?

footballhead (approved):

PR #616, opened 2026-04-28
Fix imgui viewport with xr quad layer.

hysw:


footballhead (approved):

hysw:
Note: this change did not check the xr runtime flag, although I doubt anyone is building XR version of the app and running it on plain android

PR #622, opened 2026-06-04
platform: Add AArch64 CPU feature detection

dnovillo:
This all started because I got myself an ARM laptop and tried to build BigWheels on it under Windows and WSL2.

`CpuInfo` only supported x86. On AArch64 the platform-detection code did not compile, `Platform::GetCpuInfo()` returned an empty struct, and `sample_00_ppx_info` did not build.

This PR adds AArch64 support to the CPU detection layer.

1. Add `CpuInfo::AArch64Features` to `platform.h`. It exposes 31 feature flags populated from `cpu_features::GetAarch64Info()`.
2. Add `GetAArch64CpuInfo()` to `platform.cpp`, gated on  `__aarch64__ || _M_ARM64`. Vendor and microarchitecture strings are
   resolved from the `implementer`/`part` fields using codes from the Linux kernel (`arch/arm64/include/asm/cputype.h`).
3. Update `sample_00_ppx_info` to print the AArch64 feature set on AArch64 and the x86 feature set otherwise.
4. Fix `!PPX_ANDROID` to `NOT PPX_ANDROID` in `projects/CMakeLists.txt`. The original expression always evaluated to true, so `sample_00_ppx_info` was always added regardless of the Android target.

Tested on a Snapdragon X Elite (Qualcomm Oryon X1) running Linux AArch64 and Windows 11.

NOTE: On Linux I still cannot get samples to work (they all fail with `VK_ERROR_INCOMPATIBLE_DRIVER`). I think Mesa drivers don't yet support my hardware (Qualcomm(R) Adreno(TM) X1-85 GPU). Or maybe I've not configured it properly.

All the samples work fine on Windows, though. Both in Vulkan and DX12 mode.

### DIFF
--- a/projects/sample_00_ppx_info/main.cpp
+++ b/projects/sample_00_ppx_info/main.cpp
@@ -92,6 +92,7 @@ int main(int argc, char** argv)
         PPX_ASSERT_MSG(false, "grfx::CreateInstance failed");
         return EXIT_FAILURE;
     }
+    PPX_LOG_INFO("Graphics instance and devices created successfully.");
 
     grfx::DestroyInstance(instance);
 
--- a/projects/sample_00_ppx_info/main.cpp
+++ b/projects/sample_00_ppx_info/main.cpp
@@ -21,8 +21,66 @@ const grfx::Api kApi = grfx::API_DX_12_0;
 const grfx::Api kApi = grfx::API_VK_1_1;
 #endif
 
+static void LogCpuInfo()
+{
+    const CpuInfo& cpu = Platform::GetCpuInfo();
+    // clang-format off
+    PPX_LOG_INFO("Platform         : " << Platform::GetPlatformString());
+    PPX_LOG_INFO("CPU vendor       : " << cpu.GetVendorString());
+    PPX_LOG_INFO("CPU microarch    : " << cpu.GetMicroarchitectureString());
+#if defined(__aarch64__) || defined(_M_ARM64)
+    const CpuInfo::AArch64Features& f = cpu.GetAArch64Features();
+    PPX_LOG_INFO("   fp            : " << f.fp);
+    PPX_LOG_INFO("   asimd         : " << f.asimd);
+    PPX_LOG_INFO("   aes           : " << f.aes);
+    PPX_LOG_INFO("   pmull         : " << f.pmull);
+    PPX_LOG_INFO("   sha1          : " << f.sha1);
+    PPX_LOG_INFO("   sha2          : " << f.sha2);
+    PPX_LOG_INFO("   sha512        : " << f.sha512);
+    PPX_LOG_INFO("   sha3          : " << f.sha3);
+    PPX_LOG_INFO("   crc32         : " << f.crc32);
+    PPX_LOG_INFO("   atomics       : " << f.atomics);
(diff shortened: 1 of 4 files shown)


dnovillo:
@jorgeag-google I've signed the CLA, who would be an appropriate reviewer for this change? Thanks.

src/ppx/platform.cpp
```diff
@@ -34,6 +36,207 @@ static Platform sPlatform = Platform();
...
+        case 0xc0: return "Ampere";
+        default: break;
+    }
+    return "Unknown";
+}
+
+static const char* GetAArch64MicroarchitectureName(int implementer, int part)
+{
+    // Part numbers from Linux kernel arch/arm64/include/asm/cputype.h (*_CPU_PART_*).
+    if (implementer == 0x41) { // ARM
```
hysw: nit: this function is a bit long and probably worth splitting, i..e.
```cpp
switch (implementer) {
        case 0x41: return GetAArch64ArmMicroarchitectureName(part);
        // ...
}
```

src/ppx/platform.cpp
```diff
@@ -34,6 +36,207 @@ static Platform sPlatform = Platform();
 // -------------------------------------------------------------------------------------------------
 #if defined(PPX_ANDROID)
 // TODO: Fill out ANDROID-specific info
+#elif defined(__aarch64__) || defined(_M_ARM64)
+static const char* GetAArch64VendorName(int implementer)
```
hysw: nit: internal linkage should also apply to `GetX86LongMicroarchitectureName`, so probably having a `namespace {}` block around the entire thing make more sense than individually annotate internal linkage.

hysw (approved):
Thanks.

Two nits, don't need to be fixed in this PR.

projects/sample_00_ppx_info/main.cpp
```diff
@@ -21,23 +21,78 @@ const grfx::Api kApi = grfx::API_DX_12_0;
...
     grfx::InstancePtr instance;
     Result            ppxres = grfx::CreateInstance(&createInfo, &instance);
-    if (ppxres != ppx::SUCCESS) {
+    if (Failed(ppxres)) {
         PPX_ASSERT_MSG(false, "grfx::CreateInstance failed");
         return EXIT_FAILURE;
     }
 
-    ppx::Log::Initialize(LOG_MODE_CONSOLE);
-    PPX_LOG_INFO("Graphics instance and devices created successfully.");
```
jorgeag-google, line 39: Nit: Wondering if we should continue to include this message. Since now if the instace was succesfully created we don't print anything

dnovillo: Good point. Done.

src/ppx/platform.cpp
dnovillo: Done.

src/ppx/platform.cpp
dnovillo: Sure. It was easy enough, so I rolled it in this PR. Done.

dnovillo:
@hysw @jorgeag-google thanks for the review. I think I've addressed all the feedback. PTAL.

dnovillo:
@hysw all the tests finished successfully. If the PR is ok, would you be able to submit it for me? I don't have write access to the repo. Thanks.

